PR #11, opened 2020-05-21
feat(builds): Add GitHub action to build and test

ttomsu:
Adding whole team as reviewers to see what this GitHub Actions thing is all about.

This is a simple build and test example. It has an optional step to enable caching of our (lots of) dependencies.

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -1,0 +1,29 @@
+name: Build And Test
+
+on: [push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 8
+    - uses: actions/cache@v1
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
+
+    - name: Unit Tests
+      run: |
+        ./mvnw \
+          --fail-at-end \
+          --batch-mode \
+          --activate-profiles full-checkstyle \
+          --show-version \
+          --define maven.javadoc.skip=true \
+          --threads 1.5C \
+          test


ttomsu:
Here's a sample of this working. I think it's supposed to run on first go, but maybe doesn't for security reasons?

.github/workflows/buildAndTest.yaml
```diff
@@ -0,0 +1,24 @@
...
+        java-version: 8
+    - uses: actions/cache@v1
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
+
+    - name: Build
+      run: ./mvnw compile
```
dzou, line 21: You can add the `-q` flag which is useful for suppressing extra log messages (like the ones showing what % download is reached).

meltsufin: For equivalence with what with have in `.travis.yaml`, this line would be:
```
./mvnw -T 1.5C install -DskipTests=true -Dmaven.javadoc.skip=true -Dspring.profiles.active=spring -P full-checkstyle -B -V
```

.github/workflows/buildAndTest.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+    - name: Unit Tests
+      run: ./mvnw test --fail-at-end
```
meltsufin, line 24: This line would be:
```
  ./mvnw test -B -Dit.pubsub-emulator
```
I'm not sure if we want `--fail-at-end`. I guess it will run all the tests despite intermediate failures.

.github/workflows/buildAndTest.yaml
ttomsu: Tried adding the -q flag, but it suppressed too much output - now it just has DEBUG logs, which aren't super useful without context.

I added a few of the same flags from the `travis.yaml` file - some of which we don't want though (like `skipTests!`)

I collapsed it into just running `test`, which should obviously catch any compile errors too.

.github/workflows/buildAndTest.yaml
ttomsu: I discovered `--fail-at-end` which it halted on first failure. I definitely think it's useful to run the full suite for a PR (that way we know if multiple are broken at once).

As for the pubsub emulator, that feels like more of an integration test to me. My plan is to expand these steps out farther, but hopefully in a way we can encapsulate them (say, with an `integration-test` target or something to separate them from unit tests).

.github/workflows/buildAndTest.yaml
```diff
@@ -0,0 +1,29 @@
...
+      with:
+        java-version: 8
+    - uses: actions/cache@v1
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
+
+    - name: Unit Tests
```
meltsufin, line 20: Did you remove the build step now?

.github/workflows/buildAndTest.yaml
```diff
@@ -0,0 +1,29 @@
...
+on: [push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 8
```
elefeint, line 12: Semi-rhetorical question: should we at this point move to 11?

.github/workflows/buildAndTest.yaml
elefeint: Travis folded it into `mvn test`; we likely don't need a standalone build step.

.github/workflows/buildAndTest.yaml
ttomsu: Yes, for sure. It seems support for Java 8 is waning: https://www.oracle.com/java/technologies/java-se-support-roadmap.html

I'll set up a matrix build to run against java 8 and java 11 after this.

.github/workflows/buildAndTest.yaml
ttomsu: Yes, `test` depends on `compile`

PR #14, opened 2020-05-27
builds: Adds Java versions to strategy.matrix

ttomsu:
Adds versions 11 (latest LTS) and 14 (latest non-LTS)


ttomsu:
@meltsufin - we can use [`job.needs`](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) to specify to wait for unit tests to pass across all 3 java versions before proceeding to integration tests.

PR #15, opened 2020-05-27
Add clarifying note for SQL sample

dzou:
Add a clarifying note for the SQL sample about the format that we expect the SQL instance connection name to be in.


meltsufin:
Squash and merge should take care of the noise.

PR #16, opened 2020-06-04
actions: Adds job to run integration tests

ttomsu:
A few things here:

1. This only runs integration tests guarded by `it.storage` (will start adding more next)
1. Moved non-essential _maven_ details like bucket names into the `integration-test` profile, which feels like a nice place to put it without getting in the way of the sample itself.
1. Makes maven multi-threaded by default, not just in CI (`-T 1.5C` means use _1.5*numCPU_ of threads during a parallel build) 

Side note: I think I'm seeing some Integration tests starting up in-memory emulators during the unit test phase, so we may want to add `-Dtest=!**/*IntegrationTest*"` to the UnitTest step, though I would prefer to keep a local `mvn test` as similar as possible to the CI `mvn test`

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -3,7 +3,7 @@ name: Build And Test
 on: [push, pull_request]
 
 jobs:
-  tests:
+  unitTests:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/buildAndTest.yaml
```diff
@@ -28,5 +28,39 @@ jobs:
...
+          project_id: spring-cloud-gcp-ci
+          service_account_key: ${{ secrets.SPRING_CLOUD_GCP_CI_SA_KEY }}
+          export_default_credentials: true
+      - name: Integration Tests
+        run: |
+          ./mvnw \
+            --batch-mode \
+            --activate-profiles integration-tests \
+            --show-version \
+            --define test="**/*IntegrationTest*" \
```
dzou, line 62: Just a heads up, I think there might be some integration tests that are not named correctly; might have to rename those tests as you expand to other modules.

spring-cloud-gcp-samples/pom.xml
```diff
@@ -113,6 +113,25 @@
...
+		<profile>
+			<id>integration-tests</id>
+			<build>
+				<plugins>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-surefire-plugin</artifactId>
+						<configuration>
+							<systemPropertyVariables>
+								<gcs-resource-test-bucket>gcp-storage-resource-bucket-sample</gcs-resource-test-bucket>
```
meltsufin, line 125: This seems OK, but it's important to realize that it's only possible to use this profile with the `spring-cloud-gcp-ci` project. Since bucket names are globally unique, you can't just create one with the same name in another project. So, maybe it would be more clear to name it `spring-cloud-gcp-ci-it` or something like that.

.github/workflows/buildAndTest.yaml
ttomsu: Ack - addressing in next PR.

spring-cloud-gcp-samples/pom.xml
ttomsu: Done.

PR #18, opened 2020-06-05
tests: Enable pubsub integration tests

ttomsu:
It appears that the PubSub integration tests are the reason we can't run everything in parallel. We may be able to sink some effort into making them more thread safe (removing external dependencies, maybe removing some of the standard out redirects?, marking everything as `@DirtiesContext`). 

A few more runs with this enabled may show how much time we could save by fixing pubsub and re-enabling parallel builds (which in turn test in parallel). Without Pubsub, unit + integration tests can run as fast as [4min 24sec](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/126264301)

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -24,11 +24,11 @@ jobs:
           ${{ runner.os }}-maven-
 
     - name: Unit Tests
-      continue-on-error: true
       run: |
         ./mvnw \
           --fail-at-end \
           --batch-mode \
+          --threads 1.5C \
           --activate-profiles full-checkstyle \
           --show-version \
           --define maven.javadoc.skip=true \
@@ -53,7 +53,7 @@ jobs:
           - firestore
           - logging
           - kotlin
-          # - pubsub # currently not working
+          - pubsub
           # - pubsub-emulator # not sure if this actually does anything?
           - secretmanager
           - spanner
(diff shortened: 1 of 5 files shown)


.github/workflows/buildAndTest.yaml
```diff
@@ -79,15 +79,25 @@ jobs:
           service_account_key: ${{ secrets.SPRING_CLOUD_GCP_CI_SA_KEY }}
           export_default_credentials: true
       - name: install pubsub-emulator
-        if: ${{ matrix.it }} == 'pubsub-emulator'
+        if: ${{ matrix.it == 'pubsub-emulator' }}
         run: |
           gcloud components install pubsub-emulator && \
             gcloud components update
+      # Some integration tests (pubsub) fork their own mvn instance, which needs these
+      # libraries to be installed. to the local repo.
```
dzou, line 87:
```suggestion
      # libraries to be installed to the local repo.
```

ttomsu:
Total time increased to [9m 1s](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/126338403), which isn't terribly by any stretch, especially given that current Travis CI builds are 25+min

.github/workflows/buildAndTest.yaml
ttomsu: Fixed.

ttomsu:
From chat: `pubsub-emulator` wasn't working on GHA. Will work on that next.

PR #19, opened 2020-06-08
tests: Reenable pubsub-emulator integration tests

ttomsu:
Lots of testing out why things weren't working and turned out to be something as simple as the `beta` components weren't installed.

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -54,7 +54,7 @@ jobs:
           - logging
           - kotlin
           - pubsub
-          # - pubsub-emulator # not sure if this actually does anything?
+          - pubsub-emulator
           - secretmanager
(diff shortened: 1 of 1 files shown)


.github/workflows/buildAndTest.yaml
```diff
@@ -78,14 +78,15 @@ jobs:
           project_id: spring-cloud-gcp-ci
           service_account_key: ${{ secrets.SPRING_CLOUD_GCP_CI_SA_KEY }}
           export_default_credentials: true
-      - name: install pubsub-emulator
+      - name: Install pubsub-emulator
```
elefeint, line 81: Capital "I" intended in "Install"?

ttomsu: Yes - the rest are capitalized.

PR #20, opened 2020-06-08
actions: Redirect stdout to file and always save it

ttomsu:
Two changes here:

1. Found the `always()` function that ensures the logs are actually saved, even if the job (or step) fails

2. The copious amount of Maven occasionally causes the GitHub Javascript to slow to a crawl - and most of it zooms by too fast to be useful. Found the Maven option to redirect all test stdout to files that are then archived afterward for viewing.

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -34,6 +34,7 @@ jobs:
           --define maven.javadoc.skip=true \
           test
     - name: Archive logs
+      if: always()
       uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


ttomsu:
This was also a test to confirm that integration tests did _not_ run on PRs from forks (spoiler alert: they did. Oops)

ttomsu:
One all GHA actions are complete they should become available under "Artifacts" of the top-level run: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/128948644

PR #21, opened 2020-06-08
tests: Remove Travis CI config file

ttomsu:
I think most if not all of the useful functionality has been ported over to GitHub Actions.


ttomsu:
As it stands, we're currently not even blocking on the Travis CI build.

I'd like to get to the point sooner rather than later where we're seeing green check marks by each commit rather than red Xs (see https://github.com/GoogleCloudPlatform/spring-cloud-gcp/commits/master). A failing-and-wont-be-fixed Travis CI build doesn't get us closer to that, not to mention the integration tests running on both can cause instability and flakiness.

What signal are you looking for from the Travis CI builds that makes you hesitant to stop using it?

PR #22, opened 2020-06-09
tests: One at a time integration tests

ttomsu:
This action queries GHActions endpoint to ensure only one integration test is running at a time (or at least I hope it does). This should help prevent integration tests stomping on each other's shared resources.

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -94,6 +94,12 @@ jobs:
             --define skipTests=true \
             --define maven.javadoc.skip=true \
             install
+      - name: Wait our turn for running integration tests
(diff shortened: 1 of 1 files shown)


meltsufin:
@ttomsu I believe that each category of test, such as `it.pubsub` is independent of the others. So, we should be able to run `it.pubsub` and say `it.spanner` in parallel. Does this prevent it? Also, what does the 30 min setting do?

ttomsu:
Chatted about this in our standup, but putting it here for posterity:

The intention behind this is to isolate different PRs (in this, the new repo) from interfering with each other, not between tests _within_ this workflow run.

That being said, I haven't yet seen this work, so I'm not 100% sure it's working as intended.

PR #33, opened 2020-07-27
Setting up Kokoro Configurations

dzou:
Sets up the project configuration for Spring Cloud GCP 2.0 to be published to Sonatype snapshots under `com.google.cloud`.

cc/ @meltsufin


meltsufin:
@dzou You should add release_snapshot job as well.

dzou:
> We're already a bit behind spring-cloud/spring-cloud-gcp, but it's OK.
> I guess when we do the merge, we shouldn't squash these commits?

This one we can squash - I think the only one we shouldn't squash are the ones when we merge over changes from the 1.x master branch.

meltsufin:
> > We're already a bit behind spring-cloud/spring-cloud-gcp, but it's OK.
> > I guess when we do the merge, we shouldn't squash these commits?
> 
> This one we can squash - I think the only one we shouldn't squash are the ones when we merge over changes from the 1.x master branch.

That's right! I was thinking of that one.

PR #37, opened 2020-08-03
Fix build for the Ilford version upgrade

dzou:
Should fix the build for the repo.

The Ilford snapshot of `spring-data-commons` introduces a new method to implement for our spring-data modules.

```
QueryMethodEvaluationContextProvider.getEvaluationContext(T parameters, Object[] parameterValues, ExpressionDependencies expressionDependencies)
```

This adds the method implementation.

Also, `https://repo.spring.io/snapshot` appears to be the snapshot repo we should be relying on; the old repo location was missing some dependencies.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -257,11 +257,14 @@
 					<snapshots>
 						<enabled>true</enabled>
 					</snapshots>
+					<releases>
+						<enabled>false</enabled>
+					</releases>
 				</repository>
 				<repository>
 					<id>spring-milestones</id>
 					<name>Spring Milestones</name>
-					<url>https://repo.spring.io/libs-milestone-local</url>
+					<url>https://repo.spring.io/milestone</url>
 					<snapshots>
 						<enabled>false</enabled>
 					</snapshots>
@@ -290,7 +293,7 @@
 				<pluginRepository>
 					<id>spring-snapshots</id>
 					<name>Spring Snapshots</name>
-					<url>https://repo.spring.io/libs-snapshot-local</url>
+					<url>https://repo.spring.io/snapshot</url>
 					<snapshots>
 						<enabled>true</enabled>
 					</snapshots>
@@ -301,7 +304,7 @@
 				<pluginRepository>
 					<id>spring-milestones</id>
 					<name>Spring Milestones</name>
-					<url>https://repo.spring.io/libs-milestone-local</url>
+					<url>https://repo.spring.io/milestone</url>
 					<snapshots>
 						<enabled>false</enabled>
 					</snapshots>
--- a/pom.xml
+++ b/pom.xml
@@ -250,16 +250,24 @@
 		<profile>
 			<id>spring</id>
 			<repositories>
+<!--				<repository>-->
+<!--					<id>spring-snapshots</id>-->
+<!--					<name>Spring Snapshots</name>-->
+<!--					<url>https://repo.spring.io/libs-snapshot-local</url>-->
+<!--					<snapshots>-->
+<!--						<enabled>true</enabled>-->
+<!--					</snapshots>-->
+<!--					<releases>-->
+<!--						<enabled>false</enabled>-->
+<!--					</releases>-->
+<!--				</repository>-->
 				<repository>
 					<id>spring-snapshots</id>
 					<name>Spring Snapshots</name>
-					<url>https://repo.spring.io/libs-snapshot-local</url>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -253,13 +253,10 @@
 				<repository>
 					<id>spring-snapshots</id>
 					<name>Spring Snapshots</name>
-					<url>https://repo.spring.io/libs-snapshot-local</url>
+					<url>https://repo.spring.io/snapshot</url>
 					<snapshots>
 						<enabled>true</enabled>
 					</snapshots>
-					<releases>
-						<enabled>false</enabled>
```
meltsufin, line 261: I think this should still be set to `false` because we don't want to pull release versions from the snapshots repo.

Btw, the milestone repo below should also be updated to `https://repo.spring.io/milestone`.

Also, you should do the same for plugin repositories below.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/support/DatastoreRepositoryFactory.java
```diff
@@ -118,21 +119,34 @@ public void setApplicationContext(ApplicationContext applicationContext)
...
 				evaluationContext.addPropertyAccessor(new BeanFactoryAccessor());
 				evaluationContext.setBeanResolver(new BeanFactoryResolver(
 						DatastoreRepositoryFactory.this.applicationContext));
 				return evaluationContext;
 			}
+
+			@Override
+			public <T extends Parameters<?, ?>> EvaluationContext getEvaluationContext(
+					T parameters, Object[] parameterValues, ExpressionDependencies expressionDependencies) {
+				StandardEvaluationContext evaluationContext =
```
meltsufin, line 140: How did you figure this specific implementation?

@mp911de Any suggestions on what the default implementation should be?

pom.xml
dzou: Done, changed it for all the URLs across the repo.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/support/DatastoreRepositoryFactory.java
dzou: I copied another class which implements the interface: https://github.com/spring-projects/spring-data-commons/blob/fb15a9647d9da3f4df20748b1b8de0570e4a2579/src/main/java/org/springframework/data/repository/query/ExtensionAwareQueryMethodEvaluationContextProvider.java#L93

It looks like the only extra step is to add the `ExpressionDependencies` as an extra input when creating the context.

mp911de: `ExpressionDependencies` is an optimization to use only `EvaluationContextExtension`s that are required by the SpEL expression to be evaluated. `ExpressionDependencies` was primarily driven by Reactive SpEL execution (see `ReactiveQueryMethodEvaluationContextProvider`).

On a related note: Why do you have this delegate in place? Or asked differently, what doesn't work if you use the `QueryMethodEvaluationContextProvider` provided by `RepositoryFactoryBeanSupport`?

mp911de:
We introduced with Spring Data 2020.0.0 also support for reactive auditing (`ReactiveAuditingHandler`).

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/support/DatastoreRepositoryFactory.java
dzou: > On a related note: Why do you have this delegate in place? Or asked differently, what doesn't work if you use the QueryMethodEvaluationContextProvider provided by RepositoryFactoryBeanSupport?

Hmm good question, I'm not sure - this code was written a while ago and am not sure of the context behind it.

We can definitely experiment to see if it can be removed, would be nice to reduce some code complexity.

dzou:
> We introduced with Spring Data 2020.0.0 also support for reactive auditing (`ReactiveAuditingHandler`).

Thanks for the heads up and also reviewing our code Mark! Really appreciate your support.

Issue #53, opened 2020-09-16
PubSub binder integration test broken on Boot 2.4 + Config Server 3.0.0.M3


ttomsu:
Our https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/master/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/src/test/java/com/example/LocalSampleAppIntegrationTest.java test is currently broken because of an incompatibility between our floating Spring Boot version (currently 2.4-SNAPSHOT) and pinned Config server (currently, 2.2.3.BUILD-SNAPSHOT). 

The break manifests as the Config server not starting up. I think the source incompatibility is from https://spring.io/blog/2020/08/14/config-file-processing-in-spring-boot-2-4

Upgrading the Config Server to 3.0.0-M3 fixes the server startup, but the client app doesn't appear to pick up the config on bootstrap. I think it's related to https://github.com/spring-cloud/spring-cloud-config/issues/1695.

Issue #43, opened 2020-08-11
Pub/Sub Spring Cloud Bus Config incompatible with Ilford


dzou:
I believe there are API changes upstream in `spring-cloud-context` which breaks something for our Pub/Sub integration for Spring Cloud Bus. This is related to the Secret Manager breakage here too: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/41 (see context there, same root cause PR).

Marking as P0 as a reminder to resolve this before release.

To reproduce the error:

Run:

```
cd spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-server-local
mvn spring-boot:run
```

Then visit: http://localhost:8888/application/default

You will get the following `IllegalStateException`

```
java.lang.IllegalStateException: ConfigFileApplicationListener [org.springframework.boot.context.config.ConfigFileApplicationListener] is deprecated and can only be used as an EnvironmentPostProcessor
	at org.springframework.boot.context.config.ConfigFileApplicationListener.onApplicationEvent(ConfigFileApplicationListener.java:195) ~[spring-boot-2.4.0-SNAPSHOT.jar:2.4.0-SNAPSHOT]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:203) ~[spring-context-5.3.0-M2.jar:5.3.0-M2]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:196) ~[spring-context-5.3.0-M2.jar:5.3.0-M2]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:170) ~[spring-context-5.3.0-M2.jar:5.3.0-M2]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:148) ~[spring-context-5.3.0-M2.jar:5.3.0-M2]
	at org.springframework.boot.context.event.EventPublishingRunListener.environmentPrepared(EventPublishingRunListener.java:80) ~[spring-boot-2.4.0-SNAPSHOT.jar:2.4.0-SNAPSHOT]
[...]

PR #55, opened 2020-09-17
fix(tests): Bring Config server in Pubsub integration test up to date

ttomsu:
with Spring Boot 2.4

Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/43
Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/53

1.) Upgrades pinned Spring Cloud Config server to the 3.0.0-SNAPSHOT (we've had recent discussions about whether we should be doing this, but since we're being pulled along with the Spring Boot 2.4-SNAPSHOT this needed to be upgraded too)

2.) Adds the `actuator` project to the sample so that I could ping `/actuator/refresh` manually and see the client refresh the value. Not strictly needed but was valuable for debugging.

3.) Added `spring.config.use-legacy-processing=true` to follow what @spencergibb did in his tests in the cloud config repo. I think in theory this shouldn't be necessary but it doesn't work without it right now.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-client/src/main/resources/application.properties
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-client/src/main/resources/application.properties
@@ -1,3 +1,6 @@
+spring.config.use-legacy-processing=true
+management.endpoints.web.exposure.include=*
+
 # Customize the Pub/Sub Topic to be used as the message bus.
 #spring.cloud.bus.destination=custom-spring-cloud-bus-topic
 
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-client/src/main/resources/application.properties
```diff
@@ -1,3 +1,6 @@
+spring.config.use-legacy-processing=true
```
dzou, line 1: I think you should be able to remove this with the addition of the starter.

`spring.config.use-legacy-processing=true` doesn't have any effect if it is added in `application.properties` (at least this is what I understood from Spencer's explanation).

dzou (approved):

PR #56, opened 2020-09-18
Linkage Checker configuration as GitHub Actions

suztomo:
Followup of spring-cloud/spring-cloud-gcp#2489.

- spring-cloud-gcp-dependencies to have the Linkage Checker enforcer plugin in "linkage-check" profile. It's off by default.
- It uses the latest Linkage Checker version 1.5.3, which fixes GoogleCloudPlatform/cloud-opensource-java#1609 and https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1656.
- GitHub Actions on pull requests to run the enforcer plugin

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -139,3 +139,46 @@ jobs:
         with:
           name: Test logs - ${{ matrix.it}}
           path: "**/target/failsafe-reports/*"
+
+  linkageCheck:
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 8
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: install
+        # install before running Linkage Checker
+        run: |
+          ./mvnw \
+             --batch-mode \
+             --show-version \
+             --threads 1.5C \
+             --define maven.test.skip=true \
+             --define skip.surefire.tests=true \
+             --define skipITs=true \
+             --define maven.javadoc.skip=true \
+             --errors \
+             install
+      - name: run linkage checker
+        run: |
+          ./mvnw \
+            --activate-profiles linkage-check \
+            --batch-mode \
+            --show-version \
+            --threads 1.5C \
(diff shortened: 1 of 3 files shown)


meltsufin (commented):
Should there be an explicit result of the linkage check? I see that the action overall passed, but I don't see any output from the linkage checker. Is that expected?

suztomo:
It prints the result. As the build uses multi-thread it's a bit difficult to find:

```
[INFO] --- maven-enforcer-plugin:3.0.0-M3:enforce (enforce-linkage-checker) @ spring-cloud-gcp-dependencies ---
...
[INFO] No reachable error found
```

meltsufin (approved):
LGTM.
Adding @ttomsu for double-checking the GitHub Action config.

.github/workflows/buildAndTest.yaml
```diff
@@ -126,3 +126,46 @@ jobs:
...
+      - name: run linkage checker
+        run: |
+          ./mvnw \
+            --activate-profiles linkage-check \
+            --batch-mode \
+            --show-version \
+            --threads 1.5C \
+            --define maven.test.skip=true \
+            --define skip.surefire.tests=true \
+            --define skipITs=true \
```
ttomsu, line 168: The other build and test job is also doing an mvn verify, but actually runs the integration tests. 

a) does the link checker only run in the verify stage, and if so
b) we currently run ITs as a presubmit for internal folks and on merge to master for external contributions. Is that sufficient enough to just roll with it in the existing 'build and test' job?

suztomo: a) Yes, it only runs verify phase.

b) This can be added to the existing build and test. But that would add additional 13 minutes ([the previous run](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/56/checks?check_run_id=1133866122) took 13 mins) to mark a PR as green. I think you want to run this Linkage Checker in parallel to existing GitHub Actions.

meltsufin: b) Yeah, let's definitely run this one in parallel.

ttomsu:
I promise I'm going to get to this soon - doing a bunch of POM bit-twiddling right now.

ttomsu:
Planning to get this merged in next week.

suztomo:
Thank you!

ttomsu:
Thanks for your patience! It does look like it takes a long time, so I opted to have it only run on our daily schedule and any manual runs. Going to test the manual part out in a bit once I get a few other PRs in

ttomsu:
Linkage checker completed in just over 13m - https://github.com/GoogleCloudPlatform/spring-cloud-gcp/runs/1248805849?check_suite_focus=true

suztomo:
Given other unit tests finish in 5-7 mins and dependencies seldom change thus, decoupling it from PR checks makes sense.

PR #58, opened 2020-09-28
fix(test): Run integration tests as part of nightly build and test

ttomsu:
I *think* this should work - the [documentation says](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#schedule) scheduled runes do not have an `event` payload.


elefeint (approved):
Runes :D

PR #59, opened 2020-09-29
fix(tests): Run at 5:30am Eastern (not UTC), and actually trigger on …

ttomsu:
…'scheduled' event. Confirmed this works from a [small test run I did](https://github.com/ttomsu/spring-cloud-gcp/runs/1181862957?check_suite_focus=true#step:2:75).

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -4,7 +4,7 @@ on:
   push:
   pull_request:
   schedule:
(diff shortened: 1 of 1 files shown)


ttomsu:
Thanks for the tips @ethomson https://www.edwardthomson.com/blog/github_actions_12_information_about_your_workflow.html

meltsufin (approved):

PR #61, opened 2020-09-29
Pull in recent changes from old repo

ttomsu:
Did this one a little differently: 

```
$ git merge old/master -X ours
```

where `old` is a remote for the old `spring-cloud/spring-cloud-gcp` repo, and `-X ours` makes all merge conflicts prefer the `GoogleCloudPlatform/` version.

AFAICT this didn't cause any issue, but please take a quick look over the changes and see if there was something overwritten from your commit that shouldn't have been. 

Like in the previous times we've done this, I'll relax the restriction on merge commits before submitting this change.

### DIFF
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -26,7 +26,7 @@ dependencies {
 }
 ----
 
-We provide a link:../spring-cloud-gcp-starters/spring-cloud-gcp-starter-data-datastore[Spring Boot Starter for Spring Data Datastore], with which you can use our recommended auto-configuration setup.
+We provide a link:https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-data-datastore[Spring Boot Starter for Spring Data Datastore], with which you can use our recommended auto-configuration setup.
 To use the starter, see the coordinates below.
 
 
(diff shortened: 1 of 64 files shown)


docs/src/main/asciidoc/datastore.adoc
```diff
@@ -26,7 +26,7 @@ dependencies {
 }
 ----
 
-We provide a link:../spring-cloud-gcp-starters/spring-cloud-gcp-starter-data-datastore[Spring Boot Starter for Spring Data Datastore], with which you can use our recommended auto-configuration setup.
+We provide a link:https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-data-datastore[Spring Boot Starter for Spring Data Datastore], with which you can use our recommended auto-configuration setup.
```
dzou, line 29: Change to `googlecloudplatform` org link instead of `spring-cloud`?

Though it might be easier to just make a note in your spreadsheet to fix all the links in one go and leave this here if you prefer.

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -59,7 +59,7 @@ To setup Spring Data Cloud Datastore, you have to configure the following:
 
 ==== Cloud Datastore settings
 
-You can the use link:../spring-cloud-gcp-starters/spring-cloud-gcp-starter-data-datastore[Spring Boot Starter for Spring Data Datastore] to autoconfigure Google Cloud Datastore in your Spring application.
+You can use the link:https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-data-datastore[Spring Boot Starter for Spring Data Datastore] to autoconfigure Google Cloud Datastore in your Spring application.
```
dzou, line 62: same here with the link; should change to point to `googlecloudplatform` repo

ttomsu:
Looks like this isn't going to work as expected - "it works on my machine" but apparently doesn't on GHA

PR #64, opened 2020-09-30
feat(pom): Remove dependency on spring-cloud-build POM

ttomsu:
There's a lot of maven cruft in this PR - I removed the root POM's parent and transferred over several of the plugins + their properties (and then sorted them). I'm not a maven expert (yet... coming close it seems), so I'm not 100% sure this does what I think it does, but it does feel more "composition over inheritance"-y. 

Question: There were a few PRs out recently around checkstyle and whatnot, but I also recall seeing something about using Google's checkstyle, which IIRC is more relaxed than the Spring/Pivotal one. Since this is now a Google repo I'd probably opt to follow `GoogleCloudPlatform` standards rather than Pivotals, though I don't feel strongly about it. So... **which checkstyle should we be using?**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,13 +3,6 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 
-	<parent>
-		<groupId>org.springframework.cloud</groupId>
-		<artifactId>spring-cloud-build</artifactId>
-		<version>3.0.0-SNAPSHOT</version>
-		<relativePath/><!-- lookup parent from repository -->
-	</parent>
-
 	<groupId>com.google.cloud</groupId>
 	<artifactId>spring-cloud-gcp</artifactId>
 	<version>2.0.0-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -210,6 +229,87 @@
...
+								<failOnViolation>
+									${maven-checkstyle-plugin.failOnViolation}
+								</failOnViolation>
+							</configuration>
+							<goals>
+								<goal>check</goal>
+							</goals>
+						</execution>
+						<execution>
+							<id>no-http-checkstyle-validation</id>
```
dzou, line 297: Do you think we should remove this check? I don't think we used this one before despite it being inherited.

dzou (commented):
Looks good; we might be able to cut out the `no-http-checkstyle-validation` since it looks like it was always skipped by default and we never manually enabled it in the past.

Regarding checkstyle - Yeah I would be down to use either checkstyle. But I could see switching over from pivotal to google checkstyle might be quite difficult. So if you feel the effort is not worth the benefit I wouldn't mind staying with Pivotal checkstyle either.

pom.xml
```diff
@@ -60,23 +53,41 @@
...
 	<properties>
-		<maven.compiler.source>1.8</maven.compiler.source>
-		<maven.compiler.target>1.8</maven.compiler.target>
-		<spring-cloud.version>2020.0.0-SNAPSHOT</spring-cloud.version>
-		<spring-cloud-sleuth.version>2.2.4.BUILD-SNAPSHOT</spring-cloud-sleuth.version>
-		<zipkin-gcp.version>0.17.0</zipkin-gcp.version>
 		<app-engine-maven-plugin.version>2.3.0</app-engine-maven-plugin.version>
-		<kotlin.version>1.3.31</kotlin.version>
+		<asciidoctor-maven-plugin.version>2.0.0</asciidoctor-maven-plugin.version>
+		<checkstyle.additional.suppressions.file>${maven.multiModuleProjectDirectory}/src/checkstyle/checkstyle-suppressions.xml</checkstyle.additional.suppressions.file>
```
meltsufin, line 58: Can we try to group these checkstyle properties together with the other ones?
This is just so many properties. I think the idea was that they can be overwritten more easily in the child POMs. However, since we're not using inheritance anymore, the settings can probably just go straight into the plugin configuration.

meltsufin:
@ttomsu Can you also clarify in the PR description as to why we're removing the dependency on the spring-cloud-build POM. I'm sure we've discussed this before, but honestly don't remember the exact reasons anymore? Was it for more flexibility?

pom.xml
ttomsu: I alphabetized them for easier search+finding, but you're right they can probably go straight into the plugin config.

ttomsu:
re: why we're removing it: 
IIRC, someone mentioned (you? @elefeint? I don't remember)  in our standup to remove the parent POM. It made sense to me since we're supposed to be completely independent in the new repo, so we should have full control over our build and POM file(s).

elefeint:
It was me -- since we are not part of the release train anymore, with a different build process and (in the future) a different set of checkstyle rules, I think the spring-cloud parent might be more of a hindrance. But I don't have a strong opinion on it.

pom.xml
ttomsu: Removed all `nohttp` stuff.

pom.xml
ttomsu: Moved a few of these properties inline.

ttomsu:
I think removing this inheritance may allow us to have a proper root POM for everything? Not sure if that's possible given the samples all inherit from a Boot pom, but I'll investigate a bit.

ttomsu:
Confirmed the `docs/generate-docs.sh` script still succeeds.

dzou (approved):

PR #67, opened 2020-10-06
Rename package to 'com.google.cloud.spring'

ttomsu:
This is a monster PR and I expect every one of you to review every line! :-P

This took several attempts to get just right, so here are the commands I used:

```
# Make all new directories
$ find . -path "*/src/*/org/springframework/cloud/gcp" -exec mkdir -p {}/../../../../com/google/cloud/spring/ \;

# `git mv` the files into those directories. This ends up being not that useful since we also have to modify the file
$ find . -maxdepth 9 -type d -path "*/src/*/org/springframework/cloud/gcp/*" -exec git mv -v {} {}/../../../../../com/google/cloud/spring \;

# Big ol' find and replace using some tried and true tools.
$ grep -rl -F org.springframework.cloud.gcp ./** | xargs sed -i 's/org\.springframework\.cloud\.gcp/com\.google\.cloud\.spring/g'
$ grep -rl -F org/springframework/cloud/gcp ./** | xargs sed -i 's/org\/springframework\/cloud\/gcp/com\/google\/cloud\/spring/g'
$ grep -rl -F org.springframework.cloud:spring-cloud-gcp ./** | xargs sed -i 's/org.springframework.cloud:spring-cloud-gcp/com.google.cloud:spring-cloud-gcp/g'

# Delete the old (now empty) directories
$ find . -path "*/src/*/org/springframework/cloud/gcp" -exec rmdir -p {} \;
```

You'll find a few more commits fixing up tests, and then another big one where I finagled IntelliJ to "Optimize Imports" in such a way to make the checkstyle tool happy.

I intend to squash these commits when merging in.


dzou:
Nice one!

I think you need one more find and replace for github URLs - replace `https://github.com/spring-cloud/spring-cloud-gcp` with `https://github.com/GoogleCloudPlatform/spring-cloud-gcp`.

ttomsu:
You're right! Just added:

```
$ grep -rl -F spring-cloud/spring-cloud-gcp ./** | xargs sed -i 's/spring-cloud\/spring-cloud-gcp/GoogleCloudPlatform\/spring-cloud-gcp/g'
```

dzou:
Looks good on my end.

After this change goes in, we should start main development in this repo?

ttomsu:
Short of any big and/or important customer issues, I'd be on board with most future contributions happening here. Maybe once we get a milestone release out? It's @meltsufin 's call.

meltsufin (approved):
Of course the code that needs to be reviewed is the script, and it looks great!

I'm fine with major new development starting to happen in this repo, if it's not something we plan to backport. Otherwise, I think I would prefer continuing to do work in the old repo for things that would need to be backported anyway.

PR #75, opened 2020-10-12
Fix Release Process

ttomsu:
Fixes some build warnings I was seeing locally, then fixes the stage.sh script (I hope)


elefeint (approved):
I like the expanded maven properties! Easier to grasp the meaning without trying to remember the abbreviations.

dzou (approved):

ttomsu:
> Easier to grasp the meaning without trying to remember the abbreviations.

My thoughts exactly

PR #77, opened 2020-10-12
Enable release profile during drop.sh script

ttomsu:
Also noticed some sample modules didn't have names or went against the convention.


pom.xml
```diff
@@ -407,6 +408,17 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>org.sonatype.plugins</groupId>
+				<artifactId>nexus-staging-maven-plugin</artifactId>
```
dzou, line 413: I think this was already added to the pom under the `release` profile lower in this file; was that not having the correct effect?

ttomsu: You're right! the profile was not being activated in `drop.sh` - updated PR to active the profile instead. PTAL

dzou (approved):

PR #78, opened 2020-10-12
Add nexus-staging plugin to samples, but skip it

ttomsu:
Step 1 when debugging Korkoro jobs - it's nice to know what exact command it's failed on, but without revealing any secrets. `set -v` shows what command is running without evaluating those variables, like `set -x` would, so things like `${GPG_PASSPHRASE}` should not be exposed.

Step 2: the latest job failed because it could not `nexus-staging:drop` on the code samples (yet another example of not having a strict hierarchy is a PITA):

```
...
[INFO] Spring Cloud GCP Pub/Sub Starter ................... SUCCESS [  0.001 s]
[INFO] Spring Cloud GCP Code Samples ...................... FAILURE [  5.839 s]
[INFO] Spring Cloud GCP Runtime Configuration Code Sample . SKIPPED
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 35.620 s
[INFO] Finished at: 2020-10-12T12:04:15-07:00
[INFO] ------------------------------------------------------------------------
[ERROR] No plugin found for prefix 'nexus-staging' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/kbuilder/.m2/repository), spring-snapshots (https://repo.spring.io/snapshot), spring-milestones (https://repo.spring.io/milestone), central (https://repo.maven.apache.org/maven2)] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
```

### DIFF
--- a/spring-cloud-gcp-samples/pom.xml
+++ b/spring-cloud-gcp-samples/pom.xml
@@ -30,6 +30,7 @@
 		<skip.failsafe.tests>false</skip.failsafe.tests>
 		<integration-test.pattern>**/*IntegrationTest*</integration-test.pattern>
 		<maven-deploy-plugin.version>2.8.2</maven-deploy-plugin.version>
+		<nexus-staging-plugin.version>1.6.8</nexus-staging-plugin.version>
 	</properties>
 
 	<modules>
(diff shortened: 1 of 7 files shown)


elefeint (approved):

spring-cloud-gcp-samples/pom.xml
```diff
@@ -89,6 +90,22 @@
 				</plugins>
 			</build>
 		</profile>
+		<profile>
+			<id>release</id>
```
meltsufin, line 94: Does it have to be in the `release` profile? We have a related thing for skipping deploy below, which is not in any profile. See:
```
			<plugin>
				<configuration>
					<skip>true</skip>
				</configuration>
			</plugin>
```

ttomsu: It appears the Spring release used the standard Maven deploy plugin, where as Maven central uses the Nexus-staging plugin for some enhanced release capabilities. AFAICT the Nexus-staging one supercedes maven-deploy, but I'm not 100% on that.

As far as this being in the profile, I'm including it here because the root POM and the other non-inherited module (`spring-cloud-gcp-dependencies`)  also only includes the plugin in the `-P release` profile.

PR #81, opened 2020-10-13
Add lots of debugging output for LocalSampleAppIntegrationTest

ttomsu:
This test seems to be the most flaky, and there's no logging insight into what part is failing.


ttomsu:
You can run it for yourself to see the difference in output:

```
$ ./mvnw test -Dtest=LocalSampleAppIntegrationTest -pl spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test -Dit.pubsub=true -P spring-cloud-gcp-ci-it
```

elefeint (approved):
That's a good idea. Although the two `source.toString()` statements will add a lot of output from each server starting up, so we may want to turn the log level down (up?) again after resolving the flake.

ttomsu:
Agreed - notice some of those are `.info()` calls and the raw output is sent to `.debug()`. We'll flip the logback-test back to `info` when we find what it's getting hung up on.

PR #84, opened 2020-10-13
chore(pom): Move release-related plugins to release profile, add to GHA CI

ttomsu:
Added a separate GHActions job to test the `-P release` flow, which generates javadoc and sources Jars, flattens the POM, etc. when preparing for a release.

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -15,7 +15,7 @@ jobs:
       github.repository == 'GoogleCloudPlatform/spring-cloud-gcp' &&
       (
         github.event_name == 'push' ||
-        github.event.pull_request.head.repo.full_name != github.repository
+        github.event.pull_request.head.repo.full_name != 'GoogleCloudPlatform/spring-cloud-gcp'
       )
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -659,16 +580,116 @@
...
+						</executions>
+					</plugin>
 				</plugins>
 			</build>
+			<reporting>
+				<plugins>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-javadoc-plugin</artifactId>
+						<version>${maven-javadoc-plugin.version}</version>
```
dzou, line 675: Maybe you can omit the version here if you got it covered in plugin management.

dzou (approved):

.github/workflows/buildAndTest.yaml
```diff
@@ -140,6 +140,39 @@ jobs:
           name: Test logs - ${{ matrix.it}}
           path: "**/target/failsafe-reports/*"
 
+  releaseCheck:
+    if: |
+      github.repository == 'GoogleCloudPlatform/spring-cloud-gcp' &&
+      (
+        github.event_name == 'push' ||
+        github.event.pull_request.head.repo.full_name != github.repository
```
elefeint, line 148: Am I reading it wrong or does it say "only do this when the pull request originated in a fork"?

elefeint (changes requested):
I might be having reading comprehension issues.

pom.xml
```diff
@@ -659,16 +580,116 @@
...
+							</execution>
+						</executions>
+					</plugin>
 				</plugins>
 			</build>
+			<reporting>
+				<plugins>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-javadoc-plugin</artifactId>
```
meltsufin, line 674: Do we need to repeat it under `reporting`? We don't seem to be doing it for Spanner Hibernate?

.github/workflows/buildAndTest.yaml
ttomsu: It says it's ok to run this from a user's fork.

pom.xml
ttomsu: You're right, it looks like it's not needed. Removing.

pom.xml
ttomsu: Removed entirely.

elefeint (approved):

PR #87, opened 2020-10-14
Make UserAgentHeaderProviderIntegrationTests run

meltsufin:
Fixes the regex and moves the class to another module to make sure
that the `getImplementationVersion()` method works.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/UserAgentHeaderProviderIntegrationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/UserAgentHeaderProviderIntegrationTests.java
@@ -14,10 +14,11 @@
  * limitations under the License.
  */
 
-package com.google.cloud.spring.core;
+package com.google.cloud.spring.autoconfigure.core;
 
 import java.util.regex.Pattern;
 
+import com.google.cloud.spring.core.UserAgentHeaderProvider;
 import org.junit.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -30,7 +31,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * @author Mike Eltsufin
  * @author Chengyuan Zhao
  */
-public class UserAgentHeaderProviderIT {
+public class UserAgentHeaderProviderIntegrationTests {
 
 	/**
 	 * This test is check if the generated User-Agent header is in the right format.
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/UserAgentHeaderProviderIntegrationTests.java
```diff
@@ -39,7 +40,7 @@
 	public void testGetHeaders() {
 		UserAgentHeaderProvider subject = new UserAgentHeaderProvider(this.getClass());
 
-		String versionRegex = "\\d+\\.\\d+\\.\\d+\\.[BUILD-SNAPSHOT|M\\d+|RC\\d+|RELEASE]$";
+		String versionRegex = "\\d+\\.\\d+\\.\\d+(\\-SNAPSHOT)?";
```
ttomsu, line 43: I imagine you'll want milestone releases too with the `-M1` suffix ?

ttomsu (approved):

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/UserAgentHeaderProviderIntegrationTests.java
```diff
@@ -14,10 +14,11 @@
  * limitations under the License.
  */
 
-package com.google.cloud.spring.core;
+package com.google.cloud.spring.autoconfigure.core;
```
ttomsu, line 17: I imagine you'd want to keep this in the `core` package, no? I'm still trying to explore why it works in autoconfigure and not in core

elefeint: Because it gets the package name from MANIFEST, and that's not available while the same package is being built.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/UserAgentHeaderProviderIntegrationTests.java
meltsufin: I thought we decided not to do milestone releases.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/UserAgentHeaderProviderIntegrationTests.java
meltsufin: I just put it there because that's where tests for core go in the autoconfigure module.

PR #91, opened 2020-10-16
fix(test): Adds @Param annotation to query

ttomsu:
Because they don't make reflection like they used to...


dzou (approved):

meltsufin (changes requested):
It should work without it.

PR #93, opened 2020-10-19
chore(deprecations): Remove anything that's been @Deprecated

ttomsu:
Please take an extra look at the PubSub section.


dzou (approved):

.github/workflows/buildAndTest.yaml
```diff
@@ -88,7 +88,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-java@v1
         with:
-          java-version: 8
```
meltsufin, line 91: Should we also valid with Java 11 or is Java 8 support dropped with Ilford?

ttomsu: No support was dropped AFAIK, I just figured most integrations would start to use Java 11 so lets start testing against it. Not sure we need integration tests to go against both, though.

meltsufin: Fair enough. I forgot we are already (unit) testing against 8, 11, and 15.

meltsufin (approved):

Issue #4, opened 2020-04-28
Upgrade to Spring Cloud Sleuth version that is Ilford


meltsufin:
Currently, we're overwriting the version of Sleuth in the parent POM.
This overwrite should be removed and the `-Dit.trace=true` integration tests need to be made to pass with the latest version.

Follow-up to #3.

Also look at https://github.com/spring-cloud/spring-cloud-gcp/issues/2268.

meltsufin:
Also change `B3Propagation.Format.MULTI` to `B3Propagation.Format.SINGLE_NO_PARENT` as mentioned in https://github.com/spring-cloud/spring-cloud-gcp/pull/2290#issuecomment-630510628`

elefeint:
There are a few more changes requred:
* the sampler config is defaulting to NEVER again, so we'll need to reimport it manually in trace autoconfig. 
* SleuthProperties is now package private, so we can't override `supportsJoin` and `setTraceId128` in properties anymore. We'll likely need to provide the whole `SpanHandler` bean`.
* There is an old version of brave coming from somewhere again...

PR #95, opened 2020-10-21
feat(sleuth): Upgrade to latest and greatest Sleuth (now with less Zipkin!)

ttomsu:
This one was swimming in the deep end for me. Many thanks to @elefeint for the guidance.

Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/4


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -220,8 +226,23 @@ public Sender stackdriverSender(GcpTraceProperties traceProperties,
...
 	public BaggagePropagation.FactoryBuilder baggagePropagationFactoryBuilder() {
-		return BaggagePropagation.newFactoryBuilder(StackdriverTracePropagation.newFactory(
-				B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.MULTI).build()));
+		Propagation.Factory primary = B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.MULTI).build();
+		return BaggagePropagation.newFactoryBuilder(StackdriverTracePropagation.newFactory(primary));
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public CorrelationScopeCustomizer correlationScopeCustomizer() {
```
ttomsu, line 235: @adriancole - I couldn't get this to work any other way. Before this change, we had the following values in MDC:

Screenshot from 2020-10-21 11-43-14

But afterwards we only had `traceId` and `spanId`, and NOT their `X-B3-TraceId` and `X-B3-SpanId` equivalents. Is this the intended way to do this? I would imagine the `clear()` at the beginning means multiple systems may not be able to coexists, but I'll admit I'm new to tracing/spanning.

The code as is now only puts these two `X-B3-` values in the MDC

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/logging/StackdriverLoggingAutoConfiguration.java
```diff
@@ -44,7 +44,7 @@
  */
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnClass({ HandlerInterceptorAdapter.class, LoggingAppender.class, TraceIdExtractor.class })
-@ConditionalOnMissingBean(type = "org.springframework.cloud.sleuth.autoconfig.SleuthProperties")
+@ConditionalOnMissingBean(name = "stackdriverTracingCustomizer")
```
meltsufin, line 47: What is the purpose of this condition?
I'm also not sure what the purpose of the original condition you replaced was.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -150,24 +158,22 @@ public ManagedChannel stackdriverSenderChannel() {
...
-			GcpTraceProperties trace, @Qualifier(SENDER_BEAN_NAME) Sender sender) {
-
+	@Bean(SPAN_HANDLER_BEAN_NAME)
+	@ConditionalOnMissingBean(name = SPAN_HANDLER_BEAN_NAME)
+	public SpanHandler stackdriverSpanHandler(GcpTraceProperties trace, @Qualifier(SENDER_BEAN_NAME) Sender sender) {
 		AsyncReporter<Span> asyncReporter = AsyncReporter.builder(sender)
-				// historical constraint. Note: AsyncReporter supports memory bounds
 				.queuedMaxSpans(1000)
 				.messageTimeout(trace.getMessageTimeout(), TimeUnit.SECONDS)
-				.metrics(reporterMetrics).build(StackdriverEncoder.V2);
```
meltsufin, line 162: What's replacing this setting?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -150,24 +158,22 @@ public ManagedChannel stackdriverSenderChannel() {
 				.build();
 	}
 
-	@Bean(REPORTER_BEAN_NAME)
-	@ConditionalOnMissingBean(name = REPORTER_BEAN_NAME)
-	public Reporter<Span> stackdriverReporter(ReporterMetrics reporterMetrics,
```
meltsufin, line 155: Why not keep `stackdriverReporter` as a bean, and create an additional bean for `stackdriverSpanHandler` that depends on `stackdriverReporter`?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -220,8 +226,23 @@ public Sender stackdriverSender(GcpTraceProperties traceProperties,
 	@Bean
 	@ConditionalOnMissingBean
 	public BaggagePropagation.FactoryBuilder baggagePropagationFactoryBuilder() {
-		return BaggagePropagation.newFactoryBuilder(StackdriverTracePropagation.newFactory(
-				B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.MULTI).build()));
+		Propagation.Factory primary = B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.MULTI).build();
```
meltsufin, line 229: Why did you separate this into two statements?

spring-cloud-gcp-autoconfigure/src/main/resources/com/google/cloud/spring/autoconfigure/logging/logback-appender.xml
```diff
@@ -1,18 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-
-<!-- DEPRECATED!!! PLEASE USE "com/google/cloud/spring/logging/logback-appender.xml" -->
```
meltsufin, line 3: I thought you already removed this file in another PR.

meltsufin (commented):
Nice work!
Just some clarifying questions.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/logging/StackdriverLoggingAutoConfiguration.java
ttomsu: I believe it's intended to turn off this config when Sleuth is present, but because the class we were previously using was changed to package private, we had to do something else.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
ttomsu: The reporter metrics aren't being pulled in anymore unless you have spring-cloud-sleuth-zipkin. Several iterations I had included an in memory reporter but I just took it out for simplicity. Let me know if I should put it back

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
ttomsu: Because it wasn't clear without parentheses counting what was doing what.

spring-cloud-gcp-autoconfigure/src/main/resources/com/google/cloud/spring/autoconfigure/logging/logback-appender.xml
ttomsu: See first commit message - I missed these two files.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
adriancole: configuration is decoupled now. here's how to set the old properties, but mind doing that will re-introduce the performance problems this change fixed https://github.com/spring-cloud/spring-cloud-sleuth/wiki/Spring-Cloud-Sleuth-3.0-Migration-Guide#x-b3--mdc-fields-names-are-no-longer-set

ttomsu: Can you give a little more context into the performance issue you were seeing?

adriancole: each stage in a reactive flow re-synchronizes MDC, this is amplified by various properties added to that.

The X-B3- MDC property names were deprecated for a long time and are no longer set in the 3.x release. Continuing to set these by default adds the overhead back (MDC GET/SET on every operator)

adriancole: here are the longer story https://github.com/spring-cloud/spring-cloud-sleuth/pull/1525 https://github.com/spring-cloud/spring-cloud-sleuth/pull/1226 https://github.com/spring-cloud/spring-cloud-sleuth/issues/1397

elefeint: @adriancole Let me see if I understand:
1) For propagation, Brave still uses `X-B3-TraceId` and `X-B3-SpanId`
2) but for MDC logging context, Sleuth switched to instrumentation-library-agnostic `traceId` and `spanId`?

adriancole: @elefeint yep you have it right. these topics are decoupled (incidentally, propagation over JMS always used the "b3" single header)

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/logging/StackdriverLoggingAutoConfiguration.java
meltsufin: Logging should work when Sleuth is present. Since this config is triggered after `StackdriverTraceAutoConfiguration`, this condition will always be true. I think we should just remove it.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
meltsufin: I'm not even sure what `ReporterMetrics` is for. Maybe it's used in Actuator? It's probably OK to remove it.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
meltsufin: OK :)

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
adriancole: this would show how many spans or messages were sent or dropped from client POV -> stackdriver (Assuming plumbed to micrometer)

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/logging/StackdriverLoggingAutoConfiguration.java
ttomsu: It's not clear what the ramifications are when both Logging and Sleuth are on, but there were at least a few unit tests ensuring the proper beans were being configured, so I imagine this was here for a reason.

Logging _does_ work with Sleuth both turned on and off - I think the primary case is to ensure the `TraceIdExtractor` is _not_ enabled when Sleuth is turned on, since it looks like `StackdriverTracePropagation` (from `brave-propagation-stackdriver` library) handles the `x-cloud-trace-id` header. _I think_.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
ttomsu: You're right - reverting some of these changes and adding that bean back.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
ttomsu: Added the `NOOP_METRICS` to allow users to keep reporter metrics but won't incur any memory use if not.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
ttomsu: Ok - removed this override and will rely on the instrumentation-agnostic values.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/logging/StackdriverLoggingAutoConfiguration.java
meltsufin: I think you're right... we wanted to extract the trace ID and add it to log entries, even when Sleuth was not enabled. However, when Sleuth is enabled, we get the trace ID from MDC. Took me a while to remember that.

meltsufin (approved):
Great work sorting all of this out!
:shipit:

spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/pom.xml
```diff
@@ -44,6 +44,12 @@
             <artifactId>spring-cloud-gcp-starter-logging</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.zipkin.reporter2</groupId>
+            <artifactId>zipkin-reporter-brave</artifactId>
+            <version>2.15.2</version>
```
elefeint, line 50: Odd that it's needed in a sample -- I'd expect our trace starter to bring this dependency in.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/sleuth/StackdriverHttpRequestParser.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * An {@link HttpRequestParser} that fills information for Stackdriver Trace.
+ *
+ * <p>Based on {@code org.springframework.cloud.sleuth.instrument.web.SleuthHttpClientParser}.
+ *
+ * @author Ray Tsang
+ * @author Travis Tomsu
+ */
+public class StackdriverHttpRequestParser implements HttpRequestParser {
```
elefeint, line 36: The old code also had server response parsing/customization. For sanity checking, could you test the same sample app from old repo and new repo, and just compare what shows up in Stackdriver console?

elefeint (changes requested):
Looks great; one slightly time consuming request I have is to eyeball what happens in Stackdriver  on old version vs new version.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/sleuth/StackdriverHttpRequestParser.java
ttomsu: They look to capture all the same info to me. PTAL

3H8pF5CTnRcjMyB (1)

spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/pom.xml
ttomsu: Good point, adding to the starter instead of the sample.

elefeint (approved):

PR #98, opened 2020-10-27
fix(tests): Re-enable documentation integration tests

ttomsu:
Seems failsafe (maybe surefire too?) doesn't run when `packaging` is `pom`. After discussing with the team, we agreed to move these last two tests into a real module.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/firestore/it/FirestoreDocumentationIntegrationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/firestore/it/FirestoreDocumentationIntegrationTests.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.cloud.spring.sample.firestore;
+package com.google.cloud.spring.autoconfigure.firestore.it;
 
 import java.util.Arrays;
 import java.util.concurrent.ExecutionException;
--- a/docs/src/main/asciidoc/firestore.adoc
+++ b/docs/src/main/asciidoc/firestore.adoc
@@ -355,9 +355,9 @@ The starter automatically configures and registers a `Firestore` bean in the Spr
  @Autowired
  Firestore firestore;
 
-include::{project-root}/docs/src/test/java/com/google/cloud/spring/sample/firestore/FirestoreDocumentationIntegrationTests.java[tag=write]
+include::{project-root}/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/firestore/it/FirestoreDocumentationIntegrationTests.java[tag=write]
 
-include::{project-root}/docs/src/test/java/com/google/cloud/spring/sample/firestore/FirestoreDocumentationIntegrationTests.java[tag=read]
+include::{project-root}/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/firestore/it/FirestoreDocumentationIntegrationTests.java[tag=read]
 ----
 
 === Emulator Usage
--- a/docs/src/main/asciidoc/pubsub.adoc
+++ b/docs/src/main/asciidoc/pubsub.adoc
@@ -154,7 +154,7 @@ Here is an example of how to publish a message to a Google Cloud Pub/Sub topic:
 
 [source,java,indent=0]
 ----
-include::{project-root}/docs/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubTemplateDocumentationIntegrationTests.java[tag=publish]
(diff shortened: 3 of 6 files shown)


docs/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubTemplateDocumentationIntegrationTests.java
```diff
@@ -79,9 +79,12 @@ public void testCreatePublishPullNextAndDelete() {
 			//end::publish[]
 			PubsubMessage pubsubMessage = pubSubTemplate.pullNext(subscriptionName);
 
-			assertThat(pubsubMessage.getData()).isEqualTo(ByteString.copyFromUtf8("message"));
-			assertThat(pubsubMessage.getAttributesCount()).isEqualTo(1);
-			assertThat(pubsubMessage.getAttributesOrThrow("key1")).isEqualTo("val1");
+			await().atMost(Duration.TWO_SECONDS).untilAsserted(() -> {
```
meltsufin, line 82: 2 seconds is enough?

docs/pom.xml
```diff
@@ -125,5 +143,21 @@
...
+		<profile>
+			<id>spring-cloud-gcp-ci-it</id>
+			<build>
+				<plugins>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-failsafe-plugin</artifactId>
+						<configuration>
+							<systemPropertyVariables>
+								<spring.cloud.gcp.firestore.project-id>spring-cloud-gcp-ci-firestore</spring.cloud.gcp.firestore.project-id>
```
meltsufin, line 155: Can we set this env var externally in the build config?

ttomsu: Per standup convo - moving tests is probably the better choice here.

docs/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubTemplateDocumentationIntegrationTests.java
ttomsu: Seemed to be. I can increase just to ensure it's not flaky

docs/pom.xml
ttomsu: I moved this into the test itself because it wasn't picking it up when I put it into `application-test.properties`

docs/pom.xml
```diff
@@ -88,6 +93,24 @@
 		</dependency>
 	</dependencies>
 
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<version>${maven-compiler-plugin.version}</version>
```
meltsufin, line 101: Is it still needed after moving tests?

docs/pom.xml
meltsufin: So can this be removed?

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/firestore/it/FirestoreDocumentationIntegrationTests.java
```diff
@@ -43,6 +43,7 @@
  */
 public class FirestoreDocumentationIntegrationTests {
 	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+			.withPropertyValues("spring.cloud.gcp.firestore.project-id=spring-cloud-gcp-ci-firestore")
```
meltsufin, line 46: We tried before to externalize CI-sepcific config. See: https://github.com/spring-cloud/spring-cloud-gcp/blob/master/.travis.yml#L42.
Can we make that work?

docs/pom.xml
ttomsu: Yes, removed. I originally was trying to find a way to do it outside of `-P spring-cloud-gcp-ci-it` since that profile only exists on `spring-cloud-gcp-samples`. Alas, we do what we have to do...

docs/pom.xml
ttomsu: Removed.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/firestore/it/FirestoreDocumentationIntegrationTests.java
ttomsu: Yes, added a new profile in `autoconfig` module.

ttomsu:
I removed some of the stuff about getting doc gen to work again - instead filed https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/99 and will follow up with another PR.

meltsufin (approved):
LGTM!

Issue #99, opened 2020-10-27
Refdoc generation broken


ttomsu:
It looks like `docs/generate-docs.sh` is missing a small maven config to bind to a lifecycle phase, so it's not actually doing any asciidoc --> HTML conversion.

PR #100, opened 2020-10-28
docs: Have docs module inherit from Spring Cloud pom to get docgen working again

ttomsu:
So the doc setup Spring uses is pretty complex. Instead of doing the normal copy-pasting, if we intend to continue publishing to spring.io, we probably need to continue using their exact doc-generating setup. 

Given that we no longer have any _code_ in the `docs` module, we no longer need to pull in everything as a dependency.

A good portion of this PR involves ensuring GitHub actions saves off these newly-generated archives in each run, so take a look at https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/100/checks and download the `Release Artifacts.zip` file to explore what is actually generated:
* `docs/target/generated-docs/reference/html/index.html` 
* `target/site/apidocs/index.html`

Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/99

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -3,15 +3,18 @@
 		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<parent>
-		<artifactId>spring-cloud-gcp</artifactId>
-		<groupId>com.google.cloud</groupId>
-		<version>2.0.0-SNAPSHOT</version>
+		<groupId>org.springframework.cloud</groupId>
+		<artifactId>spring-cloud-build</artifactId>
+		<version>3.0.0-M4</version>
+		<relativePath/>
 	</parent>
 	<modelVersion>4.0.0</modelVersion>
 
-	<artifactId>spring-cloud-gcp-docs</artifactId>
-	<packaging>pom</packaging>
 	<name>Spring Cloud GCP Documentation</name>
+	<groupId>com.google.cloud</groupId>
+	<artifactId>spring-cloud-gcp-docs</artifactId>
+	<version>2.0.0-SNAPSHOT</version>
+	<packaging>pom</packaging>
 
 	<properties>
 		<docs.main>spring-cloud-gcp</docs.main>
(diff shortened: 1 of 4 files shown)


docs/pom.xml
```diff
@@ -110,17 +66,17 @@
 					<plugin>
 						<groupId>org.asciidoctor</groupId>
 						<artifactId>asciidoctor-maven-plugin</artifactId>
-						<configuration>
-							<attributes>
-								<project-version>${project.version}</project-version>
```
meltsufin, line 115: Why are you removing this?

docs/pom.xml
```diff
@@ -110,17 +66,17 @@
...
-						</configuration>
 					</plugin>
 					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
 						<artifactId>maven-antrun-plugin</artifactId>
 					</plugin>
 					<plugin>
 						<artifactId>maven-deploy-plugin</artifactId>
+						<goals>
+							<goal>deploy</goal>
```
meltsufin, line 77: Why are you adding these goals, even though we didn't need them before?

meltsufin (commented):
How do we make sure the `docs` module doesn't get published to Maven Central?

docs/pom.xml
ttomsu: This should already be present: https://github.com/spring-cloud/spring-cloud-build/blob/master/pom.xml#L1149

docs/pom.xml
ttomsu: I was copy/pasting from https://github.com/spring-cloud/spring-cloud-build#reusing-the-documentation, but you're right, this should be skipped instead.

meltsufin (approved):
LGTM

docs/pom.xml
```diff
@@ -98,9 +52,11 @@
 						<artifactId>git-commit-id-plugin</artifactId>
 					</plugin>
 					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
```
dzou, line 55: Just confirming, I saw that the `maven-dependency-plugin` and `maven-resources-plugin` were missing the `groupId`s; Are these plugins necessary for the docs gen to work?

dzou (approved):
LGTM - minor question about the plugins if they are necessary or if they're removable.

docs/pom.xml
ttomsu: Not sure - it was part of a copy/paste from https://github.com/spring-cloud/spring-cloud-build#reusing-the-documentation. I didn't test it without the group IDs

PR #108, opened 2020-11-06
Allow for manually specifying BigQuery table schema

dzou:
This allows for manually specifying BigQuery schema when a datafile is loaded.

Introducing an overloaded method:

```
ListenableFuture<Job> writeDataToTable(
			String tableName, InputStream inputStream, FormatOptions dataFormatOptions, Schema schema);
```

Which allows you to manually specify the `Schema` if you do not want to rely on schema auto-detect and are unable to create the table beforehand.

https://github.com/spring-cloud/spring-cloud-gcp/issues/2576

### DIFF
--- a/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
+++ b/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
@@ -28,6 +28,7 @@ import com.google.cloud.bigquery.FormatOptions;
 import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.JobInfo.WriteDisposition;
 import com.google.cloud.bigquery.JobStatus.State;
+import com.google.cloud.bigquery.Schema;
 import com.google.cloud.bigquery.TableDataWriteChannel;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.WriteChannelConfiguration;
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -124,16 +125,26 @@ public void setJobPollInterval(Duration jobPollInterval) {
...
 				.newBuilder(tableId)
 				.setFormatOptions(dataFormatOptions)
-				.setAutodetect(this.autoDetectSchema)
 				.setWriteDisposition(this.writeDisposition)
-				.build();
+				.setAutodetect(this.autoDetectSchema);
 
-		TableDataWriteChannel writer = bigQuery.writer(writeChannelConfiguration);
+		if (schema != null) {
+			writeChannelConfiguration.setSchema(schema);
```
meltsufin, line 144: Is this compatible with auto-detecting schema?
What happens if schema is not provided, and auto-detection is off?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -124,16 +125,26 @@ public void setJobPollInterval(Duration jobPollInterval) {
...
+		WriteChannelConfiguration.Builder writeChannelConfiguration = WriteChannelConfiguration
 				.newBuilder(tableId)
 				.setFormatOptions(dataFormatOptions)
-				.setAutodetect(this.autoDetectSchema)
 				.setWriteDisposition(this.writeDisposition)
-				.build();
+				.setAutodetect(this.autoDetectSchema);
 
-		TableDataWriteChannel writer = bigQuery.writer(writeChannelConfiguration);
+		if (schema != null) {
```
meltsufin, line 143: Why allow `null` if there is a version of the method that skips this parameter?

meltsufin (commented):
Some questions.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
dzou: Yes I added it because I wanted to delegate to this method from `ListenableFuture<Job> writeDataToTable(String tableName, InputStream inputStream, FormatOptions dataFormatOptions);`

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
dzou: Yes good question - the behavior is like this:

With auto-detect schema on:
* If schema is provided - it does its best to "fit" the data to the provided schema. For example, if you provided 3 columns in the schema, and the dataset contains extra (more than 3) columns, it will only extract the 3 columns you specified from the dataset.
* If schema is not provided - It will simply infer schema from the dataset

With auto-detect schema off:
* If schema is provided - Use the explicitly provided schema to create table for dataset. This is _strict_, so if the schema specifies a different number of columns than the dataset it will fail. It will not make a best effort attempt to fit the data.
* If schema is not provided - This will simply fail.

meltsufin (approved):
LGTM

PR #113, opened 2020-11-11
fix(release): Don't activate release profile until deploy

ttomsu:
The `verify` stage fails because when `release` is activated it expects the GPG key. I could have added the skip config but I'm sure there's something else in there and we really only need the `release` profile active during the `deploy` invocation.


ttomsu:
Temporarily relaxing review requirement to get this release candidate moving.

PR #125, opened 2020-11-16
feat(integrationTests): Retry integration tests on failure

ttomsu:
This PR adds a conditional retry step for integration tests. Most of our test failures are merely flakes, and restarting the _whole_ tasks adds significant time. A single retry should help smooth out many of these random flakes in shorter time (since we don't have to repeat the `mvn install` or any of the other passing tests.

### DIFF
--- a/.github/workflows/buildAndTest.yaml
+++ b/.github/workflows/buildAndTest.yaml
@@ -139,6 +139,20 @@ jobs:
         with:
           continue-after-seconds: 1200 # 30 min
       - name: Integration Tests
+        id: intTest1
+        continue-on-error: true
+        run: |
+          ./mvnw \
+            --batch-mode \
+            --activate-profiles spring-cloud-gcp-ci-it \
+            --define maven.javadoc.skip=true \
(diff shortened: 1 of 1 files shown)


dzou (approved):
Nice!

Also cool - didn't know we could ask for reviewers from `GoogleCloudPlatform/cloud-java-frameworks-team`. Gonna start using that...

elefeint (approved):
Great idea.

meltsufin (approved):
Two questions:
1) Are macros of some kinds supported in GHAs, because it would be nice not to have to repeat these commands.
2) If the retry succeeds, does the whole build succeed, or it still marked as failed?

ttomsu:
@meltsufin I don't think there's any macro/user variable support. Since the two commands are right next to each other and don't change frequently it shouldn't be too much of an issue (especially given the number of mvn commands we juggle now anyway.)

Yes, if the retry succeeds the whole build will still pass (this is the primary intention). Obviously if they both fail the whole build fails.

PR #131, opened 2020-11-17
Replaces CloudSqlAutoConfiguration with CloudSqlEnvironmentPostProcessor

meltsufin:
Fixes: https://github.com/spring-cloud/spring-cloud-gcp/issues/272


spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring.factories
```diff
@@ -1,33 +1,33 @@
...
-org.springframework.cloud.gcp.autoconfigure.security.IapAuthenticationAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.security.FirebaseAuthenticationAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.vision.CloudVisionAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.datastore.GcpDatastoreEmulatorAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.bigquery.GcpBigQueryAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.datastore.DatastoreTransactionManagerAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.firestore.FirestoreRepositoriesAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.pubsub.health.PubSubHealthIndicatorAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.metrics.GcpStackdriverMetricsAutoConfiguration
+com.google.cloud.spring.autoconfigure.pubsub.GcpPubSubEmulatorAutoConfiguration,\
```
ttomsu: This seems like a big miss in my find/replace script, yet everything seems to work as expected. What does this file do?

ttomsu (approved):

spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring.factories
meltsufin: You didn't miss it. It's actually fine. I just had issues reapplying changes that I initially started on the old repo to here. Look at the overall diff.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,203 @@
...
+		if (databaseType != null) {
+			if (LOGGER.isInfoEnabled()) {
+				LOGGER.info("post-processing Cloud SQL properties for + " + databaseType.name());
+			}
+
+			CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider = buildCloudSqlJdbcInfoProvider(environment, databaseType);
+
+			// configure default JDBC driver and username as fallback values when not specified
+			Map<String, Object> fallbackMap = new HashMap<>();
+			fallbackMap.put("spring.datasource.username", "root");
```
elefeint, line 64: Should we be giving a default at all? It's `root` for MySQL, but [not for Postgres](https://cloud.google.com/sql/docs/postgres/users#default-users).

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,203 @@
...
+			// always set the spring.datasource.url property in the environment
+			Map<String, Object> primaryMap = new HashMap<>();
+			primaryMap.put("spring.datasource.url", cloudSqlJdbcInfoProvider.getJdbcUrl());
+			environment.getPropertySources()
+					.addFirst(new MapPropertySource("CLOUD_SQL_DATA_SOURCE_URL", primaryMap));
+
+			setCredentials(environment, application);
+
+			// support usage metrics
+			CoreSocketFactory.setApplicationName("spring-cloud-gcp-sql/"
```
elefeint, line 78: Should we have a single place where these UserAgentString constants are defined?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,203 @@
...
+			}
+			else if (isOnClasspath("com.google.cloud.sql.postgres.SocketFactory")
+					&& isOnClasspath("org.postgresql.Driver")) {
+				return DatabaseType.POSTGRESQL;
+			}
+		}
+		return null;
+	}
+
+	private boolean isOnClasspath(String className) {
```
elefeint, line 100: Is there no Spring utility we can use for this?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,203 @@
...
+			return true;
+		}
+		catch (ClassNotFoundException ex) {
+			return false;
+		}
+	}
+
+	private CloudSqlJdbcInfoProvider buildCloudSqlJdbcInfoProvider(ConfigurableEnvironment environment, DatabaseType databaseType) {
+		CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider = new DefaultCloudSqlJdbcInfoProvider(
+				getSqlProperty(environment, "database-name", null),
```
elefeint, line 112: Should we be documenting these properties? Or at least enumerating them in some enum or constant set?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,203 @@
...
+	 */
+	private void setCredentialsFileProperty(ConfigurableEnvironment environment, SpringApplication application) {
+		File credentialsLocationFile;
+
+		try {
+			String sqlCredentialsLocation = getSqlProperty(environment, "credentials.location", null);
+			String globalCredentialsLocation = environment.getProperty("spring.cloud.gcp.credentials.location", (String) null);
+			// First tries the SQL configuration credential.
+			if (sqlCredentialsLocation != null) {
+				credentialsLocationFile = application.getResourceLoader().getResource(sqlCredentialsLocation).getFile();
```
elefeint, line 169: Can `getResource()` return null?

elefeint (changes requested):
Most of my comments are on code that got moved from one class to another, so not really blocking this specific PR. Still, any cleanup is welcome.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
meltsufin: I think we should set the default as to promote the notion of convention over configuration.
But, yeah good catch on Postgres having a different username.
It's sort of breaking change, but we can do it for 2.0.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
meltsufin: Typically it's automatically detected in `UserAgentHeaderProvider`. This is a special case because we can't use `UserAgentHeaderProvider`. So, I don't think it's worthwhile to define a constant that will be used just in one place.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
meltsufin: I didn't find it. They do exactly the same thing in their `@ConditionalOnClass` implementation, but it's not reusable.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
meltsufin: What's the benefit?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
meltsufin: Javadoc says it's never `null`, but `getFile()` can throw an `IOException` which I guess is what we want anyway.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
elefeint: If we have to answer the question of "did we document everything", it would be easier to look in one spot and not play hide-and-seek with string constants. But it's up to you.

elefeint (approved):

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessorTests.java
```diff
@@ -114,7 +114,7 @@ public void testCloudSqlDataSourceWithIgnoredProvidedUrl() {
 	}
 
 	@Test
-	public void testCloudSqlAppEngineDataSourceDefaultUserNameRootTest() {
+	public void testCloudSqlAppEngineDataSourceDefaultUserNameMySqlTest() {
```
elefeint, line 117: This is a Postgres test, right? `FilteredClassLoader` removes classes, so mysql is removed here.

meltsufin: Where is `FilteredClassLoader` in this test?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,203 @@
...
+					&& isOnClasspath("org.postgresql.Driver")) {
+				return DatabaseType.POSTGRESQL;
+			}
+		}
+		return null;
+	}
+
+	private boolean isOnClasspath(String className) {
+		try {
+			ClassUtils.forName(className, null);
```
snicoll, line 102: You could use `ClassUtils#isPresent` rather.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
snicoll: You can use `Resource#exists` or `Resource#isFile`.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
snicoll: I don't know what version you looked at but `OnClassCondition` uses `ClassUtils#isPresent` that does what this method does.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/GcpCloudSqlProperties.java
```diff
@@ -1,74 +0,0 @@
...
-import org.springframework.boot.context.properties.ConfigurationProperties;
-
-/**
- * Google Cloud SQL properties.
- *
- * @author João André Martins
- * @author Øystein Urdahl Hardeng
- */
-@ConfigurationProperties("spring.cloud.gcp.sql")
-public class GcpCloudSqlProperties {
```
snicoll, line 30: I am confused. If you remove this class, no metadata is going to be generated for those keys anymore. If you hare exclusively using those in the `EnvironmentPostProcessor` you could keep this object and binds it programmatically using the `Binder`. If you don't, you need to generate the metadata manually so that those keys are still advertized in the IDE.

snicoll (commented):
Thanks for the ping @meltsufin, this looks much better. I've adde a few comments.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessorTests.java
elefeint: Never mind, github collapsed this visually with the next change, and I did not notice.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
meltsufin: I might've actually seen it somewhere else. `ClassUtils#isPresent` works, Thanks!

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
meltsufin: You're right. I added them.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
meltsufin: I think we're fine with `IOException` here.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/GcpCloudSqlProperties.java
meltsufin: I didn't know about `Binder` and forgot about the metadata support. I'll look into it.

meltsufin: Can you please see if I did it right in the last commit? Thanks!

snicoll: Thanks for the follow-up. The use of `Binder` looks correct to me.

elefeint (approved):

PR #134, opened 2020-11-19
GH Actions Updates

ttomsu:
Fixes some announcement warnings in our GHA workflow, notably:
* `ubuntu-latest` going to 20.04 (since this works, it's not strictly necessary to specify 20.04, but I want the warnings to go away)
* `setup-gcloud` moved to a new repo


elefeint (approved):
Can we revert to latest in the future, or will we need to maintain upgrades manually?

ttomsu:
We can revert to 'latest' when the transition is fully rolled out (early
2021 IIRC)

PR #135, opened 2020-11-19
switch to commit to avoid write limitations

### LINKED ISSUE #126: Use Firestore.commit call instead of Firestore.write
`write` call is limited to 10 simultaneous writes. For writing large number of entities this introduces a bottleneck. 

`commit` call is the recommended way to do that.

dmitry-s:
fixes #126

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
@@ -28,6 +28,8 @@ import com.google.cloud.spring.data.firestore.mapping.FirestorePersistentPropert
 import com.google.cloud.spring.data.firestore.transaction.ReactiveFirestoreResourceHolder;
 import com.google.cloud.spring.data.firestore.util.ObservableReactiveUtil;
 import com.google.cloud.spring.data.firestore.util.Util;
+import com.google.firestore.v1.CommitRequest;
+import com.google.firestore.v1.CommitResponse;
 import com.google.firestore.v1.Document;
 import com.google.firestore.v1.DocumentMask;
 import com.google.firestore.v1.FirestoreGrpc.FirestoreStub;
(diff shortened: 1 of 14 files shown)


ttomsu (approved):

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
```diff
@@ -209,9 +211,17 @@ public boolean isUsingStreamTokens() {
 				return Flux.from(instances).doOnNext(t -> writes.add(createUpdateWrite(t)));
 			}
 
-			Flux<List<T>> inputs = Flux.from(instances).bufferTimeout(this.writeBufferSize, this.writeBufferTimeout);
-			return ObservableReactiveUtil.streamingBidirectionalCall(
-					this::openWriteStream, inputs, this::buildWriteRequest);
+			return Flux.from(instances).bufferTimeout(this.writeBufferSize, this.writeBufferTimeout)
+					.flatMap(ts -> {
```
elefeint, line 215: I keep reading it as "timestamp". Rename to something more meaningful?

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
```diff
@@ -291,10 +301,17 @@ public FirestoreMappingContext getMappingContext() {
 				//In a transaction, all write operations should be sent in the commit request, so we just collect them
 				return Flux.from(documentNames).doOnNext(t -> writes.add(createDeleteWrite(t)));
 			}
-			return ObservableReactiveUtil.streamingBidirectionalCall(
-					this::openWriteStream,
-					documentNames.bufferTimeout(this.writeBufferSize, this.writeBufferTimeout),
-					this::buildDeleteRequest);
+			return Flux.from(documentNames).bufferTimeout(this.writeBufferSize, this.writeBufferTimeout)
+					.flatMap(ts -> {
```
elefeint, line 305: Same, rename `ts` to something -- it also turns into `e` for Entity later in this block, so maybe something more consistent like "entity".

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/it/FirestoreIntegrationTests.java
```diff
@@ -221,8 +222,8 @@ public void saveAllTest() {
 	@Test
 	public void saveAllBulkTest() {
 		Flux<User> users = Flux.create(sink -> {
-			for (int i = 0; i < 1000; i++) {
-				sink.next(new User("testUser " + i, i));
+			for (int i = 0; i < 10000; i++) {
```
elefeint, line 225: We can probably keep it at 1,000. 10K was useful for weeding out flakes, but likely not necessary for a regularly run test.

elefeint (changes requested):
Small naming things, otherwise lgtm.

meltsufin (approved):

elefeint (approved):

PR #138, opened 2020-11-23
docs(releasing): Update releasing instructions

ttomsu:


RELEASING.adoc
```diff
@@ -1,35 +1,56 @@
...
+        <module>spring-cloud-gcp-data-firestore</module>
+        <module>spring-cloud-gcp-data-spanner</module>
+        <module>spring-cloud-gcp-logging</module>
+        <module>spring-cloud-gcp-pubsub</module>
+        <module>spring-cloud-gcp-pubsub-stream-binder</module>
+        <module>spring-cloud-gcp-security-iap</module>
+        <module>spring-cloud-gcp-storage</module>
+        <module>spring-cloud-gcp-secretmanager</module>
+        <module>spring-cloud-gcp-security-firebase</module>
+        <module>spring-cloud-gcp-vision</module>
```
elefeint, line 34: Maybe a `...` under Code modules, too, so we don't have to keep it perfectly synchronized.

RELEASING.adoc
```diff
@@ -51,6 +72,9 @@ All uploaded javadocs will be published under the following URL:
 https://googleapis.dev/java/spring-cloud-gcp/{BRANCH_VERSION_NAME}/index.html
 ----
 +
-Example: If you published the javadocs for version `1.2.2.BUILD-SNAPSHOT`, then the URL would be: https://googleapis.dev/java/spring-cloud-gcp/1.2.2.BUILD-SNAPSHOT/index.html
+Example: If you published the javadocs for version `2.0.0-SNAPSHOT`, then the URL would be: https://googleapis.dev/java/spring-cloud-gcp/2.0.0-SNAPSHOT/index.html
 
 5. For GA releases, ask on the https://gitter.im/spring-cloud-gcp/Lobby[Spring Cloud GCP Gitter] for the Spring team to update the https://spring.io/projects/spring-cloud-gcp#learn[API Documentation link] to point to the published Javadocs of the correct version.
+
+6. Smoke test changes by running through a handful of our link:https://github.com/spring-io/start.spring.io/blob/master/start-site/src/main/resources/application.yml[Spring Cloud GCP codelabs] with the newly release version.
```
elefeint, line 79: This link looks wrong.

elefeint (changes requested):

RELEASING.adoc
ttomsu: Done.

RELEASING.adoc
ttomsu: Good catch - fixed.

elefeint (approved):

RELEASING.adoc
```diff
@@ -1,35 +1,57 @@
...
 
-The actual release process is triggered by the release admin and happens together with the rest of the Spring Cloud release.
-In order to simplify that process, here is a list of prerequisites to be performed before asking the release admin to run the release scripts.
+The actual release process is triggered by the release admin.
+As opposed to the 1.x releases, the 2.0 release happens independent of the Spring Cloud release train.
+To simplify the process, perform this list of prerequisites before asking the release admin to run the release scripts.
 
-. Ensure that modules that shouldn't be released have the `maven-deploy-plugin` configuration set to `<skip>true</skip>`.
+. Ensure that modules that shouldn't be released are not included in the `<modules>` block in the `release` profile of `pom.xml`.
 For GA releases this should be configured under the `central` Maven profile.
```
meltsufin, line 8: Do we even have the `central` profile now?

RELEASING.adoc
```diff
@@ -1,35 +1,57 @@
...
+        <module>spring-cloud-gcp-security-firebase</module>
+        <module>spring-cloud-gcp-vision</module>
+        ...
+    </modules>
+    ...
+</profile>
 ----
 
-. Wait for the release scripts to have been run by the release admin.
+. Ensure all Maven commands listed in `.kokoro/stage.sh` script can be executed successfully locally.
```
meltsufin, line 41: What's the purpose of this step? Does our CI verify the same things?

RELEASING.adoc
```diff
@@ -1,35 +1,57 @@
...
+You can add `-Dgpg.skip` in place of the actual GPG key locally.
+
+. Invoke the Kokoro `stage` job.
```
meltsufin, line 44: Can you list the full name of the job like we did [here](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/blob/master/RELEASING.adoc)? This will help whoever is doing it to find the job.

RELEASING.adoc
```diff
@@ -1,35 +1,57 @@
...
+Log in to OSS Sonatype and verify the contents of the staged package.
```
meltsufin, line 45: Maybe add a link and more info like [here](https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc/blob/master/RELEASING.adoc)?

RELEASING.adoc
```diff
@@ -1,35 +1,57 @@
...
 
-. Make sure that the release tag was automatically added by the release process.
+. Invoke to the Kokoko `promote` job.
 
-. Make sure that the link:https://github.com/spring-io/initializr/blob/master/initializr-service/src/main/resources/application.yml[Spring Boot Initializr] is updated to support the new version of Spring Cloud GCP.
+. Make sure that the link:https://github.com/spring-io/start.spring.io/blob/master/start-site/src/main/resources/application.yml[Spring Boot Initializr] is updated to support the new version of Spring Cloud GCP.
```
meltsufin, line 49: Does this happen automatically, or do we need to update Initializr config for every new version we release?

RELEASING.adoc
```diff
@@ -51,6 +73,9 @@ All uploaded javadocs will be published under the following URL:
 https://googleapis.dev/java/spring-cloud-gcp/{BRANCH_VERSION_NAME}/index.html
 ----
 +
-Example: If you published the javadocs for version `1.2.2.BUILD-SNAPSHOT`, then the URL would be: https://googleapis.dev/java/spring-cloud-gcp/1.2.2.BUILD-SNAPSHOT/index.html
+Example: If you published the javadocs for version `2.0.0-SNAPSHOT`, then the URL would be: https://googleapis.dev/java/spring-cloud-gcp/2.0.0-SNAPSHOT/index.html
 
 5. For GA releases, ask on the https://gitter.im/spring-cloud-gcp/Lobby[Spring Cloud GCP Gitter] for the Spring team to update the https://spring.io/projects/spring-cloud-gcp#learn[API Documentation link] to point to the published Javadocs of the correct version.
```
meltsufin, line 78: What do we need to do to get the reference documentation published and also pointed to on the project web site?

meltsufin (commented):
Sorry for the post-merge comments/questions. 😄

RELEASING.adoc
ttomsu: Yes - but I don't know why this module is disabled: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/pom.xml#L38

RELEASING.adoc
ttomsu: The intent is that the `releaseCheck` in GHA CI should catch errors, but sometimes it doesn't.

RELEASING.adoc
ttomsu: I don't think it happens automatically - but the ranges specified should be updated as new Spring Boot versions appear. This way they know the tool isn't going to give users incompatible modules.

RELEASING.adoc
ttomsu: I don't know - we talked about using our own GitHub pages to do this and having their pages link to us, rather than having them host our docs.

RELEASING.adoc
meltsufin: Ah! Great question! This is what we say in our reference documentation about it.
```
The Google Cloud Runtime Configuration service is in Beta status, and is only available in snapshot and milestone versions of the project. It’s also not available in the Spring Cloud GCP BOM, unlike other modules.
```
I think it's a good time to decide to either completely remove it, or make it part of our GA releases.

RELEASING.adoc
meltsufin: What's the danger of not doing the check locally? Wouldn't the stage job just fail?

RELEASING.adoc
meltsufin: Can ranges capture our future releases? And then, does Initializr automatically discover latest release in Maven Central or something?

RELEASING.adoc
ttomsu: Good to know - it is in fact being included in our RC releases right now: https://mvnrepository.com/artifact/com.google.cloud/spring-cloud-gcp-starter-config/2.0.0-RC1

RELEASING.adoc
ttomsu: No - it's just a time saver if it does fail for a silly reason.

RELEASING.adoc
meltsufin: I guess let's track this in a separate issue then.

RELEASING.adoc
meltsufin: Then, in the interests of streamlining the release process I would either mark this step as optional or remove it.

RELEASING.adoc
ttomsu: *Our* future releases aren't the problem as much as *their* future releases are. 

Initializer does not auto-discover anything - there is a URL (https://spring.io/project_metadata/spring-boot) that returns all "current" versions of Spring Boot and I don't know where that comes from, but AFAIK it's the only dynamic piece.

We would have to update Initializr when we do a release to bump a version.

meltsufin: Can we put that into the instructions with more details like where the version would need to be updated?

RELEASING.adoc
ttomsu: Famous last words :-)

RELEASING.adoc
ttomsu: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/141

Issue #41, opened 2020-08-10
Updates needed for Secret Manager Bootstrap Configuration


dzou:
In a change in `spring-cloud-context`, we will now need to set a property in order to enable secret bootstrap initialization: `spring.config.use-legacy-processing=true`

https://github.com/spring-cloud/spring-cloud-commons/pull/703

@meltsufin - Should we ask users to set this themselves? Or I was thinking - we could add this setting in our secret-manager-starter `application.properties` file? You could say that the addition of our starter implies they want to enable this...

elefeint:
Hmm, the presence of `use-legacy-processing` implies that there is a new way of getting properties into the bootstrap phase. Maybe document this as workaround, and get the new way into 1.2.5?

meltsufin:
+1 on what @elefeint said regarding adoption of the new non-legacy bootstrap initialization method, if possible.
Does this changes in `spring-cloud-commons` apply to Hoxton as well, or just Ilford?

dzou:
For now going to simply use the property in the tests to unblock ourselves. I've added the `p0` label as a reminder to ask for the recommended solution from the Spring team to find a proper solution before we release.

Note I think this only affects the Ilford release; the spring-cloud-gcp 1.2.x is relying on the Hoxton snapshots and should be OK.

eddumelendez:
The new way is using `spring.config.import` property key. Read `Importing Additional Configuration` and `Supporting Additional Locations` https://spring.io/blog/2020/08/14/config-file-processing-in-spring-boot-2-4

You can use `gcp-secretmanager:` as a prefix

dzou:
Followed up with Spencer and his response:

> spring.config.import bypasses some of the edge cases with bootstrap that can cause problems for folks
> By far the easiest way is to use the bootstrap starter. Just document that folks will use bootstrap still.

Seems like this should be OK for now until a user runs into issue.

Notes on how to implement `spring.config.import`: 

Eddu's PR: https://github.com/spring-cloud/spring-cloud-aws/pull/721

Some Spring Gitter Q/A:
![image](https://user-images.githubusercontent.com/3209274/100002656-e37c9580-2d92-11eb-9340-badeb96d5a81.png)

dzou:
Gonna re-open this - I think we have to do this at some point eventually anyway. And it seems like bootstrap can co-exist with the new property source loading method.

[...]

PR #139, opened 2020-11-23
fix(boostrap): Use bootstrap starter and properties, removing use-leg…

ttomsu:
…acy-processing.

Changes based on 2020.0 release notes: https://github.com/spring-cloud/spring-cloud-release/wiki/Spring-Cloud-2020.0-Release-Notes#breaking-changes

Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/41


dzou (approved):
Looks good.

I think this solves the problem for us, and I can look more into the `spring.config.import` method that Eddu mentions in #41. I originally thought that bootstrap was no longer being recommended (with them calling it `legacy-processing`) but now not so sure as they also provide a starter for it.

eddumelendez:
@dzou you can take a look at https://github.com/spring-cloud/spring-cloud-aws/pull/721. let me know if you need support, glad to help.

dzou:
@eddumelendez - Thanks for the reference, this is very helpful!

PR #145, opened 2020-11-24
[WIP] ConfigData implementation for Secret Manager

### LINKED ISSUE #41: Updates needed for Secret Manager Bootstrap Configuration

dzou:
This is the ConfigData implementation for Secret Manager.

Fixes #41.


meltsufin:
@dzou Would this cause all config properties (secrets) to be loaded at once like we had before? If so, it might be a major disadvantage vs. the way things are now where we load secrets only when used.

dzou:
Closing - we can revisit in the future if needed.

PR #161, opened 2020-12-01
chore(releases): Exclude RTC starter from GA release

ttomsu:
This PR should exclude `spring-cloud-gcp-starter-config` from the GA and future RC releases (but keeps it included in `-SNAPSHOT` releases)

### DIFF
--- a/.kokoro/promote.sh
+++ b/.kokoro/promote.sh
@@ -39,6 +39,6 @@ create_settings_xml_file $MAVEN_SETTINGS_FILE
   --settings ${MAVEN_SETTINGS_FILE} \
   -DstagingRepositoryId=${STAGING_REPOSITORY_ID} \
   -Drelease=true \
-  --activate-profiles central
+  --activate-profiles skip-unreleased-modules
 
 popd
--- a/.kokoro/promote.sh
+++ b/.kokoro/promote.sh
@@ -38,6 +38,7 @@ create_settings_xml_file $MAVEN_SETTINGS_FILE
   --batch-mode \
   --settings ${MAVEN_SETTINGS_FILE} \
   -DstagingRepositoryId=${STAGING_REPOSITORY_ID} \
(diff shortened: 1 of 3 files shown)


dzou (approved):
Looks good - Maybe this is a good time to rename the `central` profile in the config pom to something more descriptive like `skip-runtime-config`?

ttomsu:
Renamed profile

.kokoro/promote.sh
```diff
@@ -38,6 +38,7 @@ create_settings_xml_file $MAVEN_SETTINGS_FILE
   --batch-mode \
   --settings ${MAVEN_SETTINGS_FILE} \
   -DstagingRepositoryId=${STAGING_REPOSITORY_ID} \
-  -Drelease=true
+  -Drelease=true \
+  --activate-profiles skip-unreleased-modules
```
meltsufin, line 42: Is there a way to test this? Cloud this accidentally mess with other profile activations?

meltsufin (commented):
I guess RTC would now only be released in snapshots. However, in the past we would include it in pre-releases like milestones and RCs. Is it possible to easily enable it be include it with those versions like `2.0.0-RC2 but not GA versions like `2.0.0`?

.kokoro/promote.sh
ttomsu: > Is there a way to test this

Not really - at least we'd be able to check during the first `stage` step to see if it's included (though it shouldn't be).

ttomsu:
I don't know of a way to include it in RCs but not in GAs.

meltsufin:
> I don't know of a way to include it in RCs but not in GAs.

Can we make it conditional on presence of a hyphens or letters in the version name?

ttomsu:
Not that I know of

meltsufin (approved):

PR #173, opened 2020-12-11
Change secret manager module to use v1 instead of v1beta

dzou:
Change secret manager module packages to depend on the `v1` library version, not `v1beta`.


meltsufin (approved):

dzou:
@meltsufin - Do you think I should backport this to the old repo? Or would it be a breaking change if someone was Autowiring the `v1beta` version client?

Issue #174, opened 2020-12-13
Spring Cloud KMS integration


gkatzioura:
**Is your feature request related to a problem? Please describe.**
Key management service has various usages. It is integrated with the GCP infrastructure or can be used in order to encrypt/decrypt text in an application (for example database field encryption)

**Describe the solution you'd like**
A spring GCP KMS integration that encrypts and decrypts text

PR #175, opened 2020-12-13
Spring Cloud KMS integration

gkatzioura:
This pull request is related to issue #174 

This component gives the ability to encrypt/decrypt text by using a specified kms encryption key.

### DIFF
--- a/docs/src/main/asciidoc/kms.adoc
+++ b/docs/src/main/asciidoc/kms.adoc
@@ -1,0 +1,74 @@
+== Google Cloud Key Management Service
+
+The https://cloud.google.com/kms/docs[Google Cloud Key Management Service (KMS)] allows you to create, import, and manage cryptographic keys and perform cryptographic operations in a single centralized cloud service.
+
+Spring Cloud GCP offers a utility template class `KmsTemplate` which allows you to conveniently specify secrets for encryption/decryption.
+
+=== Dependency Setup
+
+To begin using this library, add the `spring-cloud-gcp-starter-vision` artifact to your project.
+
+Maven coordinates, using <<getting-started.adoc#_bill_of_materials, Spring Cloud GCP BOM>>:
+
+[source,xml]
+----
+<dependency>
+  <groupId>com.google.cloud</groupId>
+  <artifactId>spring-cloud-gcp-starter-kms</artifactId>
+</dependency>
+----
+
+Gradle coordinates:
+
+[source]
+----
+dependencies {
+  implementation("com.google.cloud:spring-cloud-gcp-starter-kms")
+}
+----
+
+=== Configuration
+
+The following options may be configured with Spring Cloud GCP KMS libraries.
+
+|===========================================================================
+| Name | Description | Required | Default value
+| `spring.cloud.gcp.kms.enabled` | Enables or disables Google Cloud KMS autoconfiguration | No | `true`
+| `spring.cloud.gcp.kms.project-id` | GCP project ID of the project using Cloud KMS APIs, if different from the one in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>>. | No | Project ID is typically inferred from https://cloud.google.com/sdk/gcloud/reference/config/set[`gcloud`] configuration.
+| `spring.cloud.gcp.kms.credentials.location` | Credentials file location for authenticating with the Cloud KMS APIs, if different from the ones in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> | No | Inferred from https://cloud.google.com/docs/authentication/production[Application Default Credentials], typically set by https://cloud.google.com/sdk/gcloud/reference/auth/application-default[`gcloud`].
+|===========================================================================
+
+
+=== Basic Usage
+
+Once you have added the `spring-cloud-gcp-starter-kms` to your project, the autoconfiguration class `com.google.cloud.spring.autoconfigure.kms.GcpKmsAutoConfiguration` will be activated for your project.
+
+The `com.google.cloud.spring.kms.KmsTemplate` class provided by the autoconfiguration is the entrypoint to using Spring Cloud GCP support for Google KMS. This class allows you to specify a Cloud KMS key in your project via a URI string (format described below) and perform encryption/decryption with it.
+
+The template class automatically validates the CRC32 checksums received responses from Cloud KMS APIs to verify correctness of the response.
+
+==== Cloud KMS URI format
+
+Spring Cloud GCP offers the following URI syntax to specify Cloud KMS keys in your project:
+
+[source]
+----
+ 1. Shortest form - specify the key by location ID, key ring ID, and key ID.
+ The project is inferred from the spring.cloud.gcp.kms.project-id if set, otherwise
+ the default GCP project (such as using application-default-credentials) is used.
+
+ kms://{location-id}/{key-ring-id}/{key-id}
+
+ 2. Short form - specify project ID, location ID, key ring ID, and key ID
+
+ kms://{project-id}/{location-id}/{key-ring-id}/{key-id}
+
+ 3. Long form - specify project ID, location ID, key ring ID, and key ID.
+ This format is equivalent to the fully-qualified resource name of a Cloud KMS key.
+
+ kms://projects/{project-id}/locations/{location-id}/keyRings/{key-ring-id}/cryptoKeys/{key-id}
+----
+
+=== Sample
+
+A https://github.com/GoogleCloudPlatform/spring-cloud-gcp/tree/master/spring-cloud-gcp-samples/spring-cloud-gcp-kms-sample[Cloud KMS Sample Application] is provided which demonstrates basic encryption and decryption operations.
--- a/docs/src/main/asciidoc/kms.adoc
+++ b/docs/src/main/asciidoc/kms.adoc
@@ -53,17 +53,24 @@ Spring Cloud GCP offers the following URI syntax to specify Cloud KMS keys in yo
 
 [source]
 ----
- 1. Shortest form - specify the key by location ID, key ring ID, and key ID.
+ 1. Shortest form - specify the key by key ring ID, and key ID.
+ The project is inferred from the spring.cloud.gcp.kms.project-id if set, otherwise
+ the default GCP project (such as using application-default-credentials) is used.
+ The location is assumed to be `global`.
+
+ kms://{key-ring-id}/{key-id}
+
+ 2. Short form - specify the key by location ID, key ring ID, and key ID.
  The project is inferred from the spring.cloud.gcp.kms.project-id if set, otherwise
  the default GCP project (such as using application-default-credentials) is used.
 
  kms://{location-id}/{key-ring-id}/{key-id}
 
- 2. Short form - specify project ID, location ID, key ring ID, and key ID
+ 3. Full form - specify project ID, location ID, key ring ID, and key ID
 
  kms://{project-id}/{location-id}/{key-ring-id}/{key-id}
 
- 3. Long form - specify project ID, location ID, key ring ID, and key ID.
+ 4. Long form - specify project ID, location ID, key ring ID, and key ID.
  This format is equivalent to the fully-qualified resource name of a Cloud KMS key.
 
  kms://projects/{project-id}/locations/{location-id}/keyRings/{key-ring-id}/cryptoKeys/{key-id}
--- a/docs/src/main/asciidoc/kms.adoc
+++ b/docs/src/main/asciidoc/kms.adoc
@@ -6,7 +6,7 @@ Spring Cloud GCP offers a utility template class `KmsTemplate` which allows you 
 
 === Dependency Setup
 
-To begin using this library, add the `spring-cloud-gcp-starter-vision` artifact to your project.
+To begin using this library, add the `spring-cloud-gcp-starter-kms` artifact to your project.
 
 Maven coordinates, using <<getting-started.adoc#_bill_of_materials, Spring Cloud GCP BOM>>:
 
--- a/docs/src/main/asciidoc/kms.adoc
+++ b/docs/src/main/asciidoc/kms.adoc
@@ -47,9 +47,9 @@ The `com.google.cloud.spring.kms.KmsTemplate` bean provided by the autoconfigura
 
 The template class automatically validates the CRC32 checksums received responses from Cloud KMS APIs to verify correctness of the response.
 
-==== Cloud KMS URI format
+==== Cloud KMS Key ID format
 
-Spring Cloud GCP offers the following URI syntax to specify Cloud KMS keys in your project:
+Spring Cloud GCP offers the following key syntax to specify Cloud KMS keys in your project:
 
 [source]
 ----
@@ -58,22 +58,22 @@ Spring Cloud GCP offers the following URI syntax to specify Cloud KMS keys in yo
  the default GCP project (such as using application-default-credentials) is used.
  The location is assumed to be `global`.
 
- kms://{key-ring-id}/{key-id}
+ {key-ring-id}/{key-id}
 
  2. Short form - specify the key by location ID, key ring ID, and key ID.
  The project is inferred from the spring.cloud.gcp.kms.project-id if set, otherwise
  the default GCP project (such as using application-default-credentials) is used.
 
- kms://{location-id}/{key-ring-id}/{key-id}
+ {location-id}/{key-ring-id}/{key-id}
 
  3. Full form - specify project ID, location ID, key ring ID, and key ID
 
- kms://{project-id}/{location-id}/{key-ring-id}/{key-id}
+ {project-id}/{location-id}/{key-ring-id}/{key-id}
 
  4. Long form - specify project ID, location ID, key ring ID, and key ID.
  This format is equivalent to the fully-qualified resource name of a Cloud KMS key.
 
- kms://projects/{project-id}/locations/{location-id}/keyRings/{key-ring-id}/cryptoKeys/{key-id}
+ projects/{project-id}/locations/{location-id}/keyRings/{key-ring-id}/cryptoKeys/{key-id}
 ----
 
 === Sample
--- a/docs/src/main/asciidoc/kms.adoc
+++ b/docs/src/main/asciidoc/kms.adoc
@@ -2,7 +2,7 @@
 
 The https://cloud.google.com/kms/docs[Google Cloud Key Management Service (KMS)] allows you to create, import, and manage cryptographic keys and perform cryptographic operations in a single centralized cloud service.
 
-Spring Cloud GCP offers a utility template class `KmsTemplate` which allows you to conveniently specify secrets for encryption/decryption.
+Spring Cloud GCP offers a utility template class `KmsTemplate` which allows you to conveniently encrypt and decrypt binary or text data.
 
 === Dependency Setup
 
--- a/docs/src/main/asciidoc/kms.adoc
+++ b/docs/src/main/asciidoc/kms.adoc
@@ -43,7 +43,7 @@ The following options may be configured with Spring Cloud GCP KMS libraries.
 
 Once you have added the `spring-cloud-gcp-starter-kms` to your project, the autoconfiguration class `com.google.cloud.spring.autoconfigure.kms.GcpKmsAutoConfiguration` will be activated for your project.
 
-The `com.google.cloud.spring.kms.KmsTemplate` class provided by the autoconfiguration is the entrypoint to using Spring Cloud GCP support for Google KMS. This class allows you to specify a Cloud KMS key in your project via a URI string (format described below) and perform encryption/decryption with it.
+The `com.google.cloud.spring.kms.KmsTemplate` bean provided by the autoconfiguration is the entrypoint to using Spring Cloud GCP support for Google KMS. This class allows you to specify a Cloud KMS key in your project via a URI string (format described below) and perform encryption/decryption with it.
 
 The template class automatically validates the CRC32 checksums received responses from Cloud KMS APIs to verify correctness of the response.
 
--- a/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSException.java
+++ b/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSException.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spring.kms;
+
+import org.springframework.core.NestedRuntimeException;
+
+/**
+ * The Spring Google Cloud KMS specific {@link NestedRuntimeException}.
+ *
+ * @author Emmanouil Gkatziouras
+ */
+public class KMSException extends NestedRuntimeException {
+
+	public KMSException(String msg) {
+		super(msg);
+	}
+
+	public KMSException(String msg, Throwable cause) {
+		super(msg, cause);
+	}
+}
--- a/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
+++ b/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2017-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spring.kms;
+
+/**
+ * Describes supported operations that one can perform on the KMS API.
+ * @author Emmanouil Gkatziouras
+ */
+public interface KMSOperations {
+
+	String encrypt(String cryptoKey, String text);
+
+	String decrypt(String cryptoKey, String encryptedText);
+
+}
--- a/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
+++ b/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
@@ -18,12 +18,45 @@ package com.google.cloud.spring.kms;
 
 /**
  * Describes supported operations that one can perform on the KMS API.
+ *
+ * <p>For the methods you need specify the secret from GCP KMS by URI string.
+ * The following secret URI syntax is supported:
+ *
+ * 1. Shortest form - specify location ID, key ring ID, and key ID. Project is derived through {@link GcpProjectIdProvider}
+ * kms://{location-id}/{key-ring-id}/{key-id}
+ *
+ * 2.  Short form - specify project ID, location ID, key ring ID, and key ID
+ * kms://{project-id}/{location-id}/{key-ring-id}/{key-id}
+ *
+ * 3. Long form - specify project ID, location ID, key ring ID, and key ID
+ * kms://projects/{project-id}/locations/{location-id}/keyRings/{key-ring-id}/cryptoKeys/{key-id}
+ *
  * @author Emmanouil Gkatziouras
  */
 public interface KMSOperations {
 
+	/**
+	 * Encrypt the text using the specified KMS URI string {@code cryptoKey}.
+	 *
+	 * <p>
+	 * An encryption request will be issued using GCP KMS. The encrypted bytes received the response are Bas64 encoded.
+	 *
+	 * @param cryptoKey The KMS URI string
+	 * @param text The text to encrypt
+	 * @return The encrypted bytes in Base64 encoding
+	 */
 	String encrypt(String cryptoKey, String text);
 
+	/**
+	 * Decrypt the text using the specified KMS URI string {@code cryptoKey}.
+	 *
+	 * <p>
+	 * An decryption request will be issued using GCP KMS. The encrypted shall be Bas64 decode to extract the encrypted bytes for the request.
+	 *
+	 * @param cryptoKey The KMS URI string
+	 * @param encryptedText The encrypted bytes in Base64 encoding
+	 * @return The decrypted text
+	 */
 	String decrypt(String cryptoKey, String encryptedText);
 
 }
--- a/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
+++ b/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
@@ -41,7 +41,7 @@ public interface KMSOperations {
 	 * Encrypt the text using the specified KMS URI string {@code cryptoKey}.
 	 *
 	 * <p>
-	 * An encryption request will be issued using GCP KMS. The encrypted bytes received the response are Bas64 encoded.
+	 * An encryption request will be issued using GCP KMS. The encrypted bytes received the response are Base64 encoded.
 	 *
 	 * @param cryptoKey The KMS URI string
 	 * @param text The text to encrypt
--- a/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
+++ b/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.spring.kms;
 
+import com.google.cloud.spring.core.GcpProjectIdProvider;
+
 /**
  * Describes supported operations that one can perform on the KMS API.
  *
--- a/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
+++ b/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
@@ -53,7 +53,7 @@ public interface KMSOperations {
 	 * Decrypt the text using the specified KMS URI string {@code cryptoKey}.
 	 *
 	 * <p>
-	 * An decryption request will be issued using GCP KMS. The encrypted shall be Bas64 decode to extract the encrypted bytes for the request.
+	 * A decryption request will be issued using GCP KMS. The encrypted text shall be Base64-encoded to extract the encrypted bytes for the request.
 	 *
 	 * @param cryptoKey The KMS URI string
 	 * @param encryptedText The encrypted bytes in Base64 encoding
--- a/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSPropertyUtils.java
+++ b/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSPropertyUtils.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2017-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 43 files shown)


gkatzioura:
The integration tests require a GCP Project and a KMS keyring with two keys. 
Once a setup is done, the code can be adapted to facilitate it.

meltsufin:
@gkatzioura Many thanks for the contribution! Please allow us some time to review this.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSException.java
```diff
@@ -0,0 +1,35 @@
...
+package com.google.cloud.spring.kms;
+
+import org.springframework.core.NestedRuntimeException;
+
+/**
+ * The Spring Google Cloud KMS specific {@link NestedRuntimeException}.
+ *
+ * @author Emmanouil Gkatziouras
+ */
+public class KMSException extends NestedRuntimeException {
```
meltsufin, line 26: Please use Pascal case (or camel case for acronyms) across the board, which means `KmsException`.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @author Emmanouil Gkatziouras
+ */
+public interface KMSOperations {
+
+	/**
+	 * Encrypt the text using the specified KMS URI string {@code cryptoKey}.
+	 *
+	 * <p>
+	 * An encryption request will be issued using GCP KMS. The encrypted bytes received the response are Bas64 encoded.
```
meltsufin, line 44:
```suggestion
	 * An encryption request will be issued using GCP KMS. The encrypted bytes received the response are Base64 encoded.
```

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSOperations.java
```diff
@@ -0,0 +1,64 @@
...
+	 * @param text The text to encrypt
+	 * @return The encrypted bytes in Base64 encoding
+	 */
+	String encrypt(String cryptoKey, String text);
+
+	/**
+	 * Decrypt the text using the specified KMS URI string {@code cryptoKey}.
+	 *
+	 * <p>
+	 * An decryption request will be issued using GCP KMS. The encrypted shall be Bas64 decode to extract the encrypted bytes for the request.
```
meltsufin, line 56:
```suggestion
	 * A decryption request will be issued using GCP KMS. The encrypted text shall be Base64-encoded to extract the encrypted bytes for the request.
```

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSTemplate.java
```diff
@@ -0,0 +1,91 @@
...
+		CryptoKeyName cryptoKeyName = KMSPropertyUtils.getCryptoKeyName(cryptoKey, projectIdProvider);
+
+		ByteString plaintextByteString = ByteString.copyFromUtf8(plaintext);
+
+		EncryptRequest request = EncryptRequest.newBuilder()
+				.setName(cryptoKeyName.toString())
+				.setPlaintext(plaintextByteString)
+				.build();
+
+		EncryptResponse response = client.encrypt(request);
```
meltsufin, line 59: Should we add an option to perform integrity verification on response?
See: https://cloud.google.com/kms/docs/encrypt-decrypt

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSTemplate.java
```diff
@@ -0,0 +1,91 @@
...
+
+		return encodeBase64(response);
```
meltsufin, line 61: Why do you prefer to do base64 encoding vs. the plain String with `response.getCiphertext().toStringUtf8()`?

spring-cloud-gcp-kms/src/test/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,2 @@
+org.springframework.cloud.bootstrap.BootstrapConfiguration=\
```
meltsufin, line 1: What's the purpose of registering it as a bootstrap configuration?

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSPropertyUtils.java
```diff
@@ -0,0 +1,101 @@
...
+				keyRingId, "The GCP KMS keyRing id must not be empty: " + input);
+
+		Assert.hasText(
+				keyId, "The GCP KMS key id must not be empty: " + input);
+
+		return CryptoKeyName.newBuilder()
+				.setProject(projectId)
+				.setKeyRing(keyRingId)
+				.setCryptoKey(keyId)
+				.setLocation(locationId)
```
meltsufin, line 91: nit: I would keep the consistent order of project, location, ring, key.

meltsufin (changes requested):
First pass...
Ultimately, we should also create auto-configuration, add documentation, and a sample app, like we have for other modules.

spring-cloud-gcp-kms/src/test/resources/META-INF/spring.factories
gkatzioura: Removed - saw it on `spring-cloud-gcp-secretmanager` and thought it was convention.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSException.java
gkatzioura: Applied on all files

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSTemplate.java
gkatzioura: Yes indeed. When I started I used the guava Crc32C hashing function `Hashing.crc32c` on the old repo. However due to being annotated as `@Beta` it was failing.

I will bring back the Crc32C functionality and let me know how to proceed.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSTemplate.java
gkatzioura: If I try to do the following there is a failure on decrypting

```
		KeyManagementServiceClient client = KeyManagementServiceClient.create();

		CryptoKeyName cryptoKeyName = CryptoKeyName.newBuilder()
				.setProject("project-id")
				.setLocation("europe-west2")
				.setKeyRing("spring-cloud-gcp")
				.setCryptoKey("key-id")
				.build();

		String encrypted = client.encrypt(cryptoKeyName.toString(), ByteString.copyFromUtf8( "1234")).getCiphertext().toStringUtf8();
		String decrypted = client.decrypt(cryptoKeyName.toString(), ByteString.copyFromUtf8(encrypted)).getPlaintext().toStringUtf8();

		Assert.assertEquals("1234",decrypted);
```

`Caused by: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Decryption failed: the ciphertext is invalid.`

Also the text is gibberish, something like

```
$!YЋ��
������^%����������R��fL*7�9-�r��p�];#��v���c&<�w��	��I�R_r���AQ�k�$z�
```

Base64 helped me a lot due to portability (storing to db, passing around) and representation.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSTemplate.java
gkatzioura: So I did put the changes with the crc32c verification and indeed there is a failure due to the `@Beta` annotation from Guava.
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/175/checks?check_run_id=1552995546#step:5:760 

If there is a workaround for this will proceed as requested.

meltsufin: This is just a checkstyle violation. We may just have to disable the rule on this file or something. When you build locally you can just add `-Dcheckstyle.skip` to your maven command.

gkatzioura: Perfect! I thought it was a strict rule. Let me know what is the preferable workaround.

For example can be disabled through maven
```
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-checkstyle-plugin</artifactId>
				<executions>
					<execution>
						<id>checkstyle-validation</id>
						<phase>validate</phase>
						<goals>
							<goal>check</goal>
						</goals>
						<configuration>
							<!-- Checkstyle rules inherited from spring-cloud-build -->
							<suppressionsLocation>src/checkstyle/checkstyle-suppressions.xml</suppressionsLocation>
....
							<excludes>**/KmsTemplate.java</excludes>
						</configuration>
					</execution>
				</executions>
			</plugin>
```

meltsufin: This works for now, but it will disable all rules on the file. I think you can also disable the specific rule on the specific file in `checkstyle-suppressions.xml`.

gkatzioura: Thanks. Moved it there!

```
<suppressions>
    <suppress files="[\\/]spring-cloud-gcp-samples[\\/]" checks="HideUtilityClassConstructorCheck" />
    <suppress files="package\-info\.java" checks="RegexpHeader" />
    <suppress files=".*" checks="LineLength" />
    <suppress checks="Illegal" files=".*RSAKeyGeneratorUtils.java"/>
    <suppress checks="IllegalImport" files=".*KmsTemplate.java"/>
</suppressions>
```

gkatzioura: Due to being a granular configuration I set this as resolved. Let me know otherwise.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsOperations.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.cloud.spring.core.GcpProjectIdProvider;
+
+/**
+ * Describes supported operations that one can perform on the KMS API.
+ *
+ * <p>For the methods you need specify the secret from GCP KMS by URI string.
+ * The following secret URI syntax is supported:
+ *
+ * 1. Shortest form - specify location ID, key ring ID, and key ID. Project is derived through {@link GcpProjectIdProvider}
+ * kms://{location-id}/{key-ring-id}/{key-id}
```
dzou, line 28: Regarding the new `kms://` syntax -- In your personal use-case, did you find the custom URI syntax easier or would you prefer a different kind of interface, maybe something like:

```
KmsTemplate.getKey(projectId, locationId, keyRingId, keyId).encrypt(...)
KmsTemplate.getKey(locationId, keyRingId, keyId).decrypt(...)
```

One of the reasons we chose the URI approach for Secret Manager was because we wanted to let users specify URIs in their `application.properties` files to access secrets directly.

However this use-case doesn't seem to exist in KMS, so I am curious if you would have found dedicated interface methods easier to use (i.e. we don't have to lock into a URI syntax just because we did it in secret manager).

dzou (commented):
Thank you for the contribution; we really appreciate this!

I wanted to circle back on the design choice regarding the `kms://` URI; just to confirm that this was indeed the easiest way that you felt you can access the API. Want to make sure that you didn't feel constrained by what we did in Secret Manager.

I was also curious; if you had free reign over the design of the interface, what do you think would be the easiest way to access KMS? I posted a comment suggestion for one idea I had.

I think we should take a step back and think about what we think might be the optimal way to access the API. And settle on the proposed API first. 

If we decide we want to change things; I am happy to do the refactoring for the PR -- We do very appreciate your time and wouldn't want you do have to do redo everything on your own.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsOperations.java
gkatzioura: Hi @dzou. Indeed the choice of `kms://` was driven by the Secret Manager implementation. Also at the time it seemed nice to have it as uri resource. I am not strong on it and I don't see extra value, my assumption was that it was expected to be like this.

gkatzioura: The `KmsTemplate` has the functions that do the encryption and decryption and one of the arguments is the encryption Key. Based on the previous comments (`kms` prefix not to be necessary probably redundant) this could be instead the full encryption key as expected from GCP `cryptoKeyName.toString()` (which happens to be called inside the code)

The reason why I choose the template to have a kms key as an argument on its methods vs having a template that has a key preconfigured, had to do with the need of multiple key usage. 
For example applications thats need to encrypt and persist data using different encryption keys, and cases where the key is dynamically specified (KEK/DEK scenarios).

This approach also seems great.

```
KmsTemplate.getKey(projectId, locationId, keyRingId, keyId).encrypt(...)
KmsTemplate.getKey(locationId, keyRingId, keyId).decrypt(...)
```

I am more than happy to proceed on what is best for the project and most suitable for the existing codebase.

dzou: I see makes sense. I can see the value the URI brings since it is more concise than the `cryptoKeyName.toString()`.

Regarding the URIs, I did some research and found some past projects/work describing people trying to decrypt values in `application.properties` with KMS [[1]](https://github.com/hfgbarrigas/gcp-kms-spring-boot) [[2]](https://stackoverflow.com/questions/57984457/cant-decrypt-property-from-spring-cloud-config-or-environment-variable), so this kind of invalidates what I said earlier; looks like it will be useful to have this URI so that users can specify keys by String in `application.properties` if necessary.

So I think we can move forward with the URIs (sorry for the confusion). I think if we want alternate ways to specify keys we can add it on top in the future too.

Right now our team is trying to reach out to the Google KMS team to get some more input on best practices too. Looking forward to hearing updates from them too.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Base64;
+
+import com.google.cloud.kms.v1.CryptoKeyName;
+import com.google.cloud.kms.v1.DecryptRequest;
+import com.google.cloud.kms.v1.DecryptResponse;
+import com.google.cloud.kms.v1.EncryptRequest;
+import com.google.cloud.kms.v1.EncryptResponse;
+import com.google.cloud.kms.v1.KeyManagementServiceClient;
+import com.google.cloud.spring.core.GcpProjectIdProvider;
+import com.google.common.hash.Hashing;
```
dzou, line 28: @meltsufin - Is it ok to use Guava here? Looks like @gkatzioura needed it to verify crc32c; looked around and this looks like the most convenient library to use in java 8.

dzou:
Added changes to our CI to enable integration tests. I removed the two tests checking error cases; it's okay to omit these since they're implementation details to KMS.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
meltsufin: I think it's fine, especially since it's the recommended approach in our documentation.

meltsufin:
@dzou Integration tests won't run because the code is in a fork, which is fine.
Regarding the `kms://...` shorthand, I'm also not sure about it. One alternative is to use the fully qualified resource name, which has the syntax `projects/{project}/locations/{location}/keyRings/{key-ring}/cryptoKeys/{key}`, but it's more verbose.

dzou:
> @dzou Integration tests won't run because the code is in a fork, which is fine.
> Regarding the `kms://...` shorthand, I'm also not sure about it. One alternative is to use the fully qualified resource name, which has the syntax `projects/{project}/locations/{location}/keyRings/{key-ring}/cryptoKeys/{key}`, but it's more verbose.

Right makes sense.

Sure, I'm down for removing the `kms://` prefix. We could go with the fully-qualified resource name for now `projects/{project}/locations/{location}/keyRings/{key-ring}/cryptoKeys/{key}` and just expand /iterate on it later. How do you feel about that @meltsufin ?

meltsufin:
> > @dzou Integration tests won't run because the code is in a fork, which is fine.
> > Regarding the `kms://...` shorthand, I'm also not sure about it. One alternative is to use the fully qualified resource name, which has the syntax `projects/{project}/locations/{location}/keyRings/{key-ring}/cryptoKeys/{key}`, but it's more verbose.
> 
> Right makes sense.
> 
> Sure, I'm down for removing the `kms://` prefix. We could go with the fully-qualified resource name for now `projects/{project}/locations/{location}/keyRings/{key-ring}/cryptoKeys/{key}` and just expand/iterate on it later. How do you feel about that @meltsufin ?

I got a response from the KMS team, and they're fine with the various shorthands as we have with Secret Manager. On a second thought, I think it's actually OK to keep the `kms://` protocol prefix, as it makes it more of a proper URI, which could be useful in the future. Also, we can always make it optional in the future.

dzou:
> thought, I think it's actually OK to keep the `kms://` protocol prefix, as it makes it more of a proper URI, which could be useful in the future. Also, we can always make it optional in the future.

Ok great, in that case let me do another pass of polish changes and then we can get this merged in.

@gkatzioura - Thank you again for the great contribution, we really appreciate the thorough work you've done adding in tests and the sample app.

dzou (approved):
PR looks good on my end. I ran the integration tests for the module and sample and both worked for me locally.

@meltsufin - If you feel good with the changes I think the PR should be good to go.

gkatzioura:
Hi an update on my side.

Due to this integration not being released on the sample I had to specify the actual version of the kms
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/17daa34c71b1c707de3d63eadf93099fbdcbd5ec/spring-cloud-gcp-samples/spring-cloud-gcp-kms-sample/pom.xml#L40-L49

Obviously this can be tackled after the release (or the process that takes place)

Had the same workaround on the autoconfigure but haven't figured out why I had to specify the version there (as if kms dependency management is not picked up by parent project)

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/17daa34c71b1c707de3d63eadf93099fbdcbd5ec/spring-cloud-gcp-autoconfigure/pom.xml#L255-L261

dzou:
> Hi an update on my side.
> 
> Due to this integration not being released, on the sample I had to specify the actual version of the kms
> https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/17daa34c71b1c707de3d63eadf93099fbdcbd5ec/spring-cloud-gcp-samples/spring-cloud-gcp-kms-sample/pom.xml#L40-L49
> 
> Obviously this can be tackled after the release (or the process that takes place)
> 
> Had the same workaround on the autoconfigure but haven't figured out why I had to specify the version there (as if kms dependency management is not picked up by parent project)
> 
> https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/17daa34c71b1c707de3d63eadf93099fbdcbd5ec/spring-cloud-gcp-autoconfigure/pom.xml#L255-L261

Thanks for catching that -- I think what was missing was adding the KMS modules to our spring-cloud-gcp-dependencies module (our BOM) and also adding the module to the starter parent pom.

Should be good now!

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KMSTemplate.java
gkatzioura: Hi @dzou & @meltsufin 
I do see Base64 to be a bit opinionated.

In any case I have a branch [here](https://github.com/gkatzioura/spring-cloud-gcp-1/tree/kms-base64-proposal/spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms)
with a proposal on returning back an object, that can give the encryption results either as byte[] or a Base64 string.

```
public class EncryptionResult {

	private final ByteString byteString;

	EncryptionResult(ByteString byteString) {
		this.byteString = byteString;
	}

	public byte[] toByteArray() {
		return byteString.toByteArray();
	}

	public String toBase64() {
		byte[] bytes = byteString.toByteArray();
		byte[] encoded = Base64.getEncoder().encode(bytes);
		return new String(encoded);
	}

}
```

```
public DecryptionResult decrypt(String cryptoKey, byte[] bytes) {
		CryptoKeyName cryptoKeyName = KmsPropertyUtils.getCryptoKeyName(cryptoKey, projectIdProvider);

		ByteString encryptedByteString = ByteString.copyFrom(bytes);
...
		return new DecryptionResult(response.getPlaintext());
	}
```

Let me know if this could be better.

dzou:
@meltsufin - Added the refdoc for the KMS integration.

Remaining items:
- Review if the refdoc is okay
- Figure out what should be the correct return type of the encrypt/decrypt messages -- https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/175#discussion_r546871909

docs/src/main/asciidoc/kms.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+Spring Cloud GCP offers the following URI syntax to specify Cloud KMS keys in your project:
+
+[source]
+----
+ 1. Shortest form - specify the key by location ID, key ring ID, and key ID.
+ The project is inferred from the spring.cloud.gcp.kms.project-id if set, otherwise
+ the default GCP project (such as using application-default-credentials) is used.
+
+ kms://{location-id}/{key-ring-id}/{key-id}
```
meltsufin, line 60: What do you think of an even shorter form for global keys?
kms://{key-ring-id}/{key-id}

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
```diff
@@ -0,0 +1,124 @@
...
+				.setName(cryptoKeyName.toString())
+				.setPlaintext(plaintextByteString)
+				.setPlaintextCrc32C(
+						Int64Value.newBuilder().setValue(crc32c).build())
+				.build();
+
+		EncryptResponse response = client.encrypt(request);
+		assertCrcMatch(response);
+
+		return encodeBase64(response);
```
meltsufin, line 68: Is the Base64 encoding and decoding really necessary?
It adds overhead that some users may not like.
I think there are two main use cases: (1) encrypting/decoding utf8-encoded text, and (2) encrypting/decrypting binary data (byte[]).

So, I would propose just having 4 methods:
```java
byte[] encryptBytes(String cryptoKey, byte[] plaintext);
byte[] encryptText(String cryptoKey, String plaintext); // UTF-8 encoded text
byte[] decryptBytes(String cryptoKey, byte[] ciphertext);
String decryptText(String cryptoKey, byte[] ciphertext); // returns UTF-8 encoded text
```

What do you think?

meltsufin (commented):
Thank you @gkatzioura and @dzou for putting this together!
My main concern is about the Base64 encoding. See my comments below.
I'm curious what you think.

docs/src/main/asciidoc/kms.adoc
gkatzioura: Great Idea, will do it!

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
gkatzioura: Understood. Will implement those methods however when it comes to the problem described on [previous](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/175#discussion_r542753418) comment, encoding to UTF-8 encode I fail to decrypt the UTF-8 encoded text.

```
		KeyManagementServiceClient client = KeyManagementServiceClient.create();

		CryptoKeyName cryptoKeyName = CryptoKeyName.newBuilder()
				.setProject("project")
				.setLocation("europe-west2")
				.setKeyRing("spring-cloud-gcp")
				.setCryptoKey("key-id")
				.build();

		String encrypted = client.encrypt(cryptoKeyName.toString(), ByteString.copyFromUtf8( "1234")).getCiphertext().toStringUtf8();
		String decrypted = client.decrypt(cryptoKeyName.toString(), ByteString.copyFromUtf8(encrypted)).getPlaintext().toStringUtf8();
```

The decryption of the String that was UTF-8 after it's encryption will fail. I am not sure what the sdk does behind the scenes but that's why I resided on base64.

Would it make sense to just have the byte methods and then up to the user on how to proceed storing and using? 
```
byte[] encryptBytes(String cryptoKey, byte[] plaintext);
byte[] decryptBytes(String cryptoKey, byte[] ciphertext);
```

Additionally those method can be put for an extra for convenience 

```
byte[] encryptText(String cryptoKey, String plaintext); // Base64 encoded text
String decryptText(String cryptoKey, byte[] ciphertext); // returns Base64
```

docs/src/main/asciidoc/kms.adoc
meltsufin: Cool! I also think it could be more user-friendly to change the sample to instruct the users to create a specific key-id, and use this short version of it to refer to it.

Similarly, we could modify the integration tests to do the same, so that they are portable across GCP projects. However, we would have to get rid of the tests that rely on a specific ciphertext literal.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
meltsufin: Let me play around with this.

meltsufin: I think there is an issue with converting between `byte[]` and `String` back and forth.
The ciphertext should probably be kept as `byte[]` because it's binary.
This works:
```java
String plaintext = "1234";
byte[] encrypted = client.encrypt(cryptoKeyName, ByteString.copyFromUtf8(plaintext)).getCiphertext().toByteArray();
String decrypted = client.decrypt(cryptoKeyName, ByteString.copyFrom(encrypted)).getPlaintext().toStringUtf8();
assertThat(plaintext).isEqualTo(decrypted);
```

gkatzioura: Yeap it seems the `.toStringUtf8()` corrupts the result (that's why base64 works, just carries the bytes as is).

Going for this one as proposed, since it will work and sticks to bytes only.

```
byte[] encryptBytes(String cryptoKey, byte[] plaintext);
byte[] decryptBytes(String cryptoKey, byte[] ciphertext);
byte[] encryptText(String cryptoKey, String plaintext); // UTF-8 encoded text
String decryptText(String cryptoKey, byte[] ciphertext); // returns UTF-8 encoded text
```

meltsufin: Exactly, UTF-8 encoding is not always reversible, although it is for Base64.
For example, this test fails:
```java
byte[] bytes = new byte[] {10, -10};
String stringFromBytes = new String(bytes);
assertThat(bytes).isEqualTo(stringFromBytes.getBytes());
```

gkatzioura: The methods have been updated as discussed. The implementation is bas64 clear.
The sample in order to make a copy paste feasible uses bas64 encoding on the controller.

docs/src/main/asciidoc/kms.adoc
gkatzioura: Changed it! Will also change also the sample.  

```
However, we would have to get rid of the tests that rely on a specific ciphertext literal.
```

For the ones mocked it is covered due to mocking the response. On the Integration tests I avoided hardcoding the ciphertexts and instead used encrypt/decrypt together  so anyone could run the tests.

gkatzioura: Updated also the sample.

docs/src/main/asciidoc/kms.adoc
```diff
@@ -0,0 +1,81 @@
+== Google Cloud Key Management Service
+
+The https://cloud.google.com/kms/docs[Google Cloud Key Management Service (KMS)] allows you to create, import, and manage cryptographic keys and perform cryptographic operations in a single centralized cloud service.
+
+Spring Cloud GCP offers a utility template class `KmsTemplate` which allows you to conveniently specify secrets for encryption/decryption.
```
meltsufin, line 5:
```suggestion
Spring Cloud GCP offers a utility template class `KmsTemplate` which allows you to conveniently encrypt and decrypt binary or text data.
```

docs/src/main/asciidoc/kms.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+=== Dependency Setup
+
+To begin using this library, add the `spring-cloud-gcp-starter-vision` artifact to your project.
```
meltsufin, line 9:
```suggestion
To begin using this library, add the `spring-cloud-gcp-starter-kms` artifact to your project.
```

docs/src/main/asciidoc/kms.adoc
```diff
@@ -0,0 +1,81 @@
...
+| `spring.cloud.gcp.kms.credentials.location` | Credentials file location for authenticating with the Cloud KMS APIs, if different from the ones in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> | No | Inferred from https://cloud.google.com/docs/authentication/production[Application Default Credentials], typically set by https://cloud.google.com/sdk/gcloud/reference/auth/application-default[`gcloud`].
+|===========================================================================
+
+
+=== Basic Usage
+
+Once you have added the `spring-cloud-gcp-starter-kms` to your project, the autoconfiguration class `com.google.cloud.spring.autoconfigure.kms.GcpKmsAutoConfiguration` will be activated for your project.
+
+The `com.google.cloud.spring.kms.KmsTemplate` class provided by the autoconfiguration is the entrypoint to using Spring Cloud GCP support for Google KMS. This class allows you to specify a Cloud KMS key in your project via a URI string (format described below) and perform encryption/decryption with it.
```
meltsufin, line 46:
```suggestion
The `com.google.cloud.spring.kms.KmsTemplate` bean provided by the autoconfiguration is the entrypoint to using Spring Cloud GCP support for Google KMS. This class allows you to specify a Cloud KMS key in your project via a URI string (format described below) and perform encryption/decryption with it.
```

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsOperations.java
```diff
@@ -0,0 +1,87 @@
...
+
+import com.google.cloud.spring.core.GcpProjectIdProvider;
+
+/**
+ * Describes supported operations that one can perform on the KMS API.
+ *
+ * <p>For the methods you need specify the secret from GCP KMS by URI string.
+ * The following secret URI syntax is supported:
+ *
+ * 1. Shortest form - specify location ID, key ring ID, and key ID. Project is derived through {@link GcpProjectIdProvider}.
```
meltsufin, line 27: Javadoc needs to be updated to describe the `key-ring/key` form.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+	public KmsTemplate(
+			KeyManagementServiceClient keyManagementServiceClient,
+			GcpProjectIdProvider projectIdProvider) {
+		this.client = keyManagementServiceClient;
+		this.projectIdProvider = projectIdProvider;
+	}
+
+	@Override
+	public byte[] encryptText(String cryptoKey, String text) {
+		byte[] bytes = text.getBytes(Charset.defaultCharset());
```
meltsufin, line 54: We should decode using `UTF-8`, since that's the convention. Or, we can just document that the platform's default charset would be used.
```suggestion
		byte[] bytes = text.getBytes(StandardCharsets.UTF_8);
```

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+
+		EncryptResponse response = client.encrypt(request);
+		assertCrcMatch(response);
+		return response.getCiphertext().toByteArray();
+	}
+
+	@Override
+	public String decryptText(String cryptoKey, byte[] cipherText) {
+		byte[] decryptedBytes = decryptBytes(cryptoKey, cipherText);
+		return new String(decryptedBytes, Charset.defaultCharset());
```
meltsufin, line 79: Again, `UTF-8` or we need to switch to platform's default charset, which might not be UTF-8.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+						.setCiphertextCrc32C(
+								Int64Value.newBuilder().setValue(crc32c).build())
+						.build();
+
+		DecryptResponse response = client.decrypt(request);
+		assertCrcMatch(response);
+		return response.getPlaintext().toByteArray();
+	}
+
+	private String encodeBase64(byte[] bytes) {
```
meltsufin, line 102: No longer used, can be removed.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+		byte[] encoded = Base64.getEncoder().encode(bytes);
+		return new String(encoded);
+	}
+
+	private byte[] decodeBase64(String encryptedText) {
```
meltsufin, line 107: No longer used, can be removed.

spring-cloud-gcp-kms/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
meltsufin, line 1: Are you actually relying on this?

spring-cloud-gcp-samples/spring-cloud-gcp-kms-sample/README.adoc
```diff
@@ -0,0 +1,41 @@
...
+```
+
+3. Authenticate in one of two ways:
+
+a. Use the Google Cloud SDK to https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login[authenticate with application default credentials].
+This method is convenient but should only be used in local development.
+b. https://cloud.google.com/iam/docs/creating-managing-service-accounts[Create a new service account], download its private key and point the `spring.cloud.gcp.kms.credentials.location` property to it.
+Such as: `spring.cloud.gcp.credentials.location=file:/path/to/creds.json`
+
+4. Using the https://console.cloud.google.com/security/kms[Key Management Service in Cloud Console], create a new global key ring named `spring-cloud-gcp` to the location of your choice, and create a key on that key-ring.
```
meltsufin, line 23: We can just ask them here to create a global key and tell them what to name the key, so that in Step 6a, we can tell them exactly what to paste into the box for the key.

meltsufin (changes requested):
Mostly polish.

Also, we had a discussion with Dan, and we're leaning now again towards removing the `kms://` prefix, as it doesn't seem to be giving us any benefit. Sorry, for flipping back and forth on this.

spring-cloud-gcp-kms/src/main/java/com/google/cloud/spring/kms/KmsTemplate.java
gkatzioura: Switched to StandardCharsets.UTF_8

spring-cloud-gcp-kms/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
gkatzioura: Yes due to mocking the final methods of the SDK

spring-cloud-gcp-samples/spring-cloud-gcp-kms-sample/README.adoc
gkatzioura: Changed it. They should create the global key `kms://spring-cloud-gcp/key-id`

gkatzioura:
Hi @meltsufin @dzou ! Are we just gonna remove the `kms` prefix? The short version do come in handy.

dzou:
> Hi @meltsufin @dzou ! Are we just gonna remove the `kms` prefix or there are going to be more changes? The short versions do come in handy.

I think we are pretty good, don't think there will be any more major changes. I just added a commit to remove the `kms://` prefix in the resource format so we can see if we're happy with that.

Most of our team is going to be on vacation for the next couple days for the holidays, so we might be a little bit slower to respond -- Sorry about this! We'll make sure this makes it into the 2.0 release :)

meltsufin (approved):
Looks great! @gkatzioura Many thanks for your patience and an end-to-end module contribution!

gkatzioura:
This is great! Thank you both for your time and assistance :)

PR #184, opened 2020-12-24
Add Logback config for KMS tests

meltsufin:
Avoid too verbose `DEBUG` level output.

### DIFF
--- a/spring-cloud-gcp-kms/src/test/resources/logback-test.xml
+++ b/spring-cloud-gcp-kms/src/test/resources/logback-test.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+		<!-- encoders are assigned the type
+		ch.qos.logback.classic.encoder.PatternLayoutEncoder by default -->
+		<encoder>
+			<pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+		</encoder>
+	</appender>
+
+	<root level="INFO">
+		<appender-ref ref="STDOUT" />
+	</root>
+</configuration>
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-kms/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+		<!-- encoders are assigned the type
```
elefeint, line 4: Remove comment? This is not relying on default.

spring-cloud-gcp-samples/spring-cloud-gcp-kms-sample/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+		<!-- encoders are assigned the type
```
elefeint, line 4: Same.

elefeint (approved):
I wonder if we had a test-support project if we could just have one place for the test logback config. Test-support came up before, I think, in the context of emulators.

spring-cloud-gcp-kms/src/test/resources/logback-test.xml
meltsufin: Why do you say it's not relying on default? When you use `<encoder>` without specifying the class, it's supposed to be the `PatternLayoutEncoder`.

elefeint: I meant that there is an explicit `<pattern>` tag, so it's obvious that `PatternLayoutEncoder` is used even without the comment.

PR #193, opened 2021-01-05
chore(builds): Use JCenter instead of the less reliable Maven Central

ttomsu:
Most of our flakey integration tests from the past month have been download errors from Maven Central. 

I recall reading somewhere (probably some JFrog marketing material) that Bintray is more reliable than Maven Central, so I'm proposing we prioritize JCenter. A recent clean build performed like so:

```
$ rm -rf ~/.m2

$ time ./mvnw install -DskipTests --quiet

real	3m16.721s
user	2m2.449s
sys	0m19.088s
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -345,6 +345,11 @@
 			<id>spring</id>
 			<repositories>
 				<repository>
+					<id>jcenter</id>
+					<name>jcenter</name>
+					<url>https://jcenter.bintray.com</url>
+				</repository>
+				<repository>
 					<id>spring-milestones</id>
 					<name>Spring Milestones</name>
 					<url>https://repo.spring.io/milestone</url>
@@ -373,6 +378,11 @@
 				</repository>
 			</repositories>
 			<pluginRepositories>
+				<pluginRepository>
+					<id>jcenter</id>
+					<name>jcenter</name>
+					<url>https://jcenter.bintray.com</url>
+				</pluginRepository>
 				<pluginRepository>
 					<id>spring-milestones</id>
 					<name>Spring Milestones</name>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -25,6 +25,11 @@
 
 	<repositories>
 		<repository>
+			<id>jcenter</id>
+			<name>jcenter</name>
+			<url>https://jcenter.bintray.com</url>
+		</repository>
+		<repository>
 			<id>spring-milestones</id>
 			<name>Spring Milestones</name>
 			<url>https://repo.spring.io/milestone</url>
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -266,6 +266,11 @@
 			<id>spring</id>
 			<repositories>
 				<repository>
+					<id>jcenter</id>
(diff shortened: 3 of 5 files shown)


meltsufin (commented):
We also have the GCS Maven Mirror. How good is that one?
Also, since we cache `.m2`, are we re-downloading dependencies on every build still?

ttomsu:
So I went looking for examples of a flaky download and can't find one because I kept re-running it.

From what I recall, the GCS mirror didn't hit as often as you'd think for being a *mirror*. For example, here's a recent run: https://pipelines.actions.githubusercontent.com/JQ7HoQSrL03ai79YfMhrTUF983wgnig0EXQ6GzwfTI8yeUdLvM/_apis/pipelines/1/runs/1053/signedlogcontent/21?urlExpires=2021-01-05T18%3A50%3A30.5847311Z&urlSigningMethod=HMACV1&urlSignature=6litRzcidHEcAeJ0IIOdyYkA2oQOvhne%2FSbJp1zydG0%3D

Search for "central" and "gcs-mirror" and you'll see that we're pulling from central far more often than mirror, and I have no idea why.

As for the .m2 cache - I believe it's working, as _most_ of our builds run without needing to redownload the world. Something expires out of the cache or something (sometimes Maven plugins?) and it halts the whole build.

ttomsu:
Here's a performance benchmark as well using Central:

```
$ time ./mvnw install -DskipTests --quiet

real	9m59.987s
user	3m27.219s
sys	0m23.754s

```

meltsufin:
Another example?
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/197/checks?check_run_id=1685496006

pom.xml
```diff
@@ -344,6 +344,11 @@
 		<profile>
 			<id>spring</id>
 			<repositories>
+				<repository>
```
meltsufin, line 347: Would you have to run this with `-P spring` spring profile enabled?

dzou:
+1 to this; I'm getting some `connection resets` every now and then too.

pom.xml
ttomsu: This is already done when run with the maven wrapper: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/master/.mvn/maven.config

ttomsu:
Had a big failure the other night with timeouts: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/500854863

ttomsu:
We had a timeout in the old repo this morning. I ported over the JVM configs that seem to be helping in the new repo: https://github.com/spring-cloud/spring-cloud-gcp/pull/2635

meltsufin:
> We had a timeout in the old repo this morning. I ported over the JVM configs that seem to be helping in the new repo: [spring-cloud/spring-cloud-gcp#2635](https://github.com/spring-cloud/spring-cloud-gcp/pull/2635)

Should we still keep this PR around?

ttomsu:
I don't think the issue is completely mitigated. I saw several failures yesterday afternoon, for example.

ttomsu:
Closing this PR because... sigh... https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

PR #217, opened 2021-01-20
Implement withParent overload for FirestoreTemplate

### LINKED ISSUE #206: FirestoreTemplate.withParent enhancement
Using`org.springframework.cloud: spring-cloud-gcp-dependencies: 1.2.6.RELEASE` and `org.springframework.cloud: spring-cloud-gcp-starter-data-firestore`.

To work with sub collections, we currently use `<T> FirestoreReactiveOperations withParent(T parent)`. To use this method we have to either:

- Have an instance of the parent object
- Create a dummy instance of the parent object. 

There are cases where neither of these are feasible. For example, there are situations where we don't have an instance of the parent object but the id. Thus we have to create a dummy/fake instance of the parent object as the sample project demonstrate. But this can result in unreadable code because the parent object can be big with many properties. For example:

`firestormTemplate.withParent(new User("the_actual_id", "", "", "", new NestedObject("", new NestedObject2()), ""))` etc

Reading the source code, `firestormTemplate.withParent` will only extract the id value from the supplied parent and discard the rest. So, my wish is an overloaded method:

`public FirestoreReactiveOperations withParent(Object id, Class<?> clazz)`

With the supplied class object, we should be able find the collection name etc. We don't need an actual instance of the parent object and we don't need to create fake objects.

dzou:
This adds an overloaded `withParent(Object id, Class<?> entityClass)` method so that users are not required to have a reference to an entity to scope the template to some subcollection.

Fixes #206.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreReactiveOperations.java
@@ -140,10 +140,11 @@ public interface FirestoreReactiveOperations {
 	<T> Flux<T> execute(StructuredQuery.Builder builder, Class<T> entityClass);
 
 	/**
-	 * Creates FirestoreReactiveOperations object with a provided parent.
+	 * Creates FirestoreReactiveOperations object with a specified parent document.
 	 * The parent doesn't have to exist in Firestore, but should have a non-empty id field.
 	 * All operations and queries will be scoped to that parent's subcollections.
-	 * By default, FirestoreReactiveOperations uses the root as a "parent".
+	 * By default, FirestoreReactiveOperations uses the root document as the parent.
+	 *
 	 * @param <T> the type param of the parent.
 	 * @param parent the query builder.
 	 * @return template with a given parent.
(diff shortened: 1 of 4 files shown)


ttomsu (approved):

ttomsu:
Check out the sonar cloud bug as well

dzou:
> Check out the sonar cloud bug as well

Done -- Added an extra nullpointer check for that case.

dzou:
@meltsufin -- Hey Mike, this is the `withParent(..)` PR. 

From it looks like in the past, we haven't used the "Key" concept you mentioned in the meeting. The `withParent(..)` method (both the old one and the one I added) only "descend" one level a time in a document tree.

I.e. If you want to have the FirestoreTemplate operate on a nested subcollection, you would write:

```
FirestoreTemplate template = firestoreTemplate
    .withParent(new Parent(...))
    .withParent(new Child(..))
    .withParent(new GrandChild(..));
```

If things get heavily nested, we might need a better way of accessing those subcollections though.

meltsufin:
@dzou Thanks for the clarification. I think I was just confused with the `Key` from Datastore.
Anyway, I'm curious why you chose `Object` for ID, rather than `String`, since only `String` is supported for `@DocumentId` in Firestore.

dzou:
> @dzou Thanks for the clarification. I think I was just confused with the `Key` from Datastore.
> Anyway, I'm curious why you chose `Object` for ID, rather than `String`, since only `String` is supported for `@DocumentId` in Firestore.

Ah I see -- I did not know this. So there's no use-case where someone might use like a `Long` for a document ID? (Or I guess it's not possible?)

meltsufin:
@dzou According to [this](https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/DocumentId) only `String` or `DocumentReference` are allowed, but I don't think we support `DocumentReference` anywhere.

dzou:
Gotcha, I'll make it String then.

PR #220, opened 2021-01-25
fix firestore-emulator test

ttomsu:
Tests have been failing over the weekend


elefeint (approved):
Will a blank project ID work, or is a dummy placeholder better?

ttomsu:
I think the purpose of the test was to test the fallback to "unused". I think the original author missed that our integration tests set this value in `spring-cloud-gcp-autoconfigure`'s `pom.xml`, so this PR drops that value just for this test.

dzou (approved):

Issue #232, opened 2021-01-28
Spring Cloud Stream Unable to setup subscription to a topic in a different GCP project


maganv:
Our application is consuming messages from a topic that is setup in another team's GCP project.
Example topic name: "projects/team-a/topics/topic-1". Our application runs in "team-b" project. Since the projects are different we need to use the full topic name in the "destination" otherwise the `PubSubChannelProvisioner`
does not find the topic (`makeSureTopicExists`-method). But then the subscription name will be "illegal" since the expected subscription name is `<topicName>.<group>`.

`spring-cloud-gcp-pubsub-stream-binder` version: 2.0.0
This is the configuration:
```
spring.cloud.stream.bindings.input.destination=projects/team-a/topics/topic-1
spring.cloud.stream.bindings.input.group=my-event-handler
```

Is there a way to set it up that I am missing?

If is not possible today, a solution could be that if the topic name contains "/" the right-most value could be used when getting the subscription. 
In the above case the expected subscription name would be "topic-1.my-event-handler".

meltsufin:
@maganv This is a really good question. Looking at it, our `PubSubChannelProvisioner#provisionConsumerDestination` is not aware of fully-qualified topic names. So, it composes something like `projects/team-a/topics/topic-1.my-event-handler` which is invalid both as a short name and as a fully qualified name of a subscription. We can probably fix it, but I'm wondering what you would expect the subscription name to be? Would you want it to be created in project `team-a` or where application is running under `team-b`?

maganv:
In our case we would create the subscription manually in project `team-b` i.e. the same project as the application is running. So if the topic's fully qualified name is `projects/team-a/topics/topic-1` we want to use the subscription with id/short-name `topic-1.my-event-handler` in project `team-b`, fully qualified name would be `projects/team-b/subscriptions/topic-1.my-event-handler`

If the topic was created in the same project as the application is using, i.e. `projects/team-b/topics/topic-1` the configuration would be like this:
```
spring.cloud.stream.bindings.input.destination=topic-1
spring.cloud.stream.bindings.input.group=my-event-handler
```
And it would use the subscription `projects/team-b/subscriptions/topic-1.my-event-handler`. We would like to achieve the same naming convention of the subscription even if the topic is in a different project. The naming convention of the subscription: `<topicName>.<group>` is fine if the `<topicName>` is the short name and not the fully qualified name even when the topic is in another project.

PR #234, opened 2021-01-29
Allow pubsub binder to use fully qualified topic name when creating subscription (consumer destination)

ttomsu:
Two commits in here:

1.) Update from deprecated `ProjectTopicName` to `TopicName`

2.) Allow specifying a fully-qualified consumer destination, which allows that destination to be in a different project. To do this, I had to add a public `getProjectId()` to the `PubSubAdmin` object, which simply returns its immutable project ID.

Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/232

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
@@ -23,7 +23,6 @@ import java.util.UUID;
 
 import com.google.api.gax.rpc.AlreadyExistsException;
 import com.google.cloud.spring.pubsub.PubSubAdmin;
-import com.google.cloud.spring.pubsub.support.PubSubTopicUtils;
 import com.google.cloud.spring.stream.binder.pubsub.properties.PubSubConsumerProperties;
 import com.google.cloud.spring.stream.binder.pubsub.properties.PubSubProducerProperties;
 import com.google.pubsub.v1.Subscription;
@@ -75,11 +74,8 @@ public class PubSubChannelProvisioner
 			throws ProvisioningException {
 
 		// topicName may be either the short or fully-qualified version.
-		TopicName topicNameObj = PubSubTopicUtils.toTopicName(topicName, this.pubSubAdmin.getProjectId());
-		String topicShortName = topicNameObj.getTopic();
-		String topicFullyQualifiedName = topicNameObj.toString();
-
-		ensureTopicExists(topicName, properties.getExtension().isAutoCreateResources());
+		String topicShortName = TopicName.isParsableFrom(topicName) ? TopicName.parse(topicName).getTopic() : topicName;
+		Optional<Topic> topic = ensureTopicExists(topicName, properties.getExtension().isAutoCreateResources());
 
 		String subscriptionName;
 		Subscription subscription;
@@ -90,19 +86,19 @@ public class PubSubChannelProvisioner
 		else {
 			// Generate anonymous random group since one wasn't provided
 			subscriptionName = "anonymous." + topicShortName + "." + UUID.randomUUID().toString();
-			subscription = this.pubSubAdmin.createSubscription(subscriptionName, topicFullyQualifiedName);
+			subscription = this.pubSubAdmin.createSubscription(subscriptionName, topicName);
 			this.anonymousGroupSubscriptionNames.add(subscriptionName);
 		}
 
 		if (subscription == null) {
 			if (properties.getExtension().isAutoCreateResources()) {
-				this.pubSubAdmin.createSubscription(subscriptionName, topicFullyQualifiedName);
+				this.pubSubAdmin.createSubscription(subscriptionName, topicName);
 			}
 			else {
 				throw new ProvisioningException("Non-existing '" + subscriptionName + "' subscription.");
 			}
 		}
-		else if (!subscription.getTopic().equals(topicFullyQualifiedName)) {
+		else if (topic.isPresent() && !subscription.getTopic().equals(topic.get().getName())) {
 			throw new ProvisioningException(
 					"Existing '" + subscriptionName + "' subscription is for a different topic '"
 							+ subscription.getTopic() + "'.");
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
@@ -17,15 +17,18 @@
 package com.google.cloud.spring.stream.binder.pubsub.provisioning;
 
 import java.util.HashSet;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 
 import com.google.api.gax.rpc.AlreadyExistsException;
 import com.google.cloud.spring.pubsub.PubSubAdmin;
+import com.google.cloud.spring.pubsub.support.PubSubTopicUtils;
 import com.google.cloud.spring.stream.binder.pubsub.properties.PubSubConsumerProperties;
 import com.google.cloud.spring.stream.binder.pubsub.properties.PubSubProducerProperties;
 import com.google.pubsub.v1.Subscription;
 import com.google.pubsub.v1.Topic;
+import com.google.pubsub.v1.TopicName;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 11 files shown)


spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/PubSubTopicUtils.java
```diff
@@ -57,4 +60,21 @@ public static ProjectTopicName toProjectTopicName(String topic, @Nullable String
 
 		return projectTopicName;
 	}
+
+	public static TopicName toTopicName(String topic, @Nullable String projectId) {
```
dzou, line 64: Add javadoc?

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -71,32 +74,35 @@ public ConsumerDestination provisionConsumerDestination(String topicName, String
 			ExtendedConsumerProperties<PubSubConsumerProperties> properties)
 			throws ProvisioningException {
 
-		Topic topic = makeSureTopicExists(topicName, properties.getExtension().isAutoCreateResources());
+		// topicName may be either the short or fully-qualified version.
+		TopicName topicNameObj = PubSubTopicUtils.toTopicName(topicName, this.pubSubAdmin.getProjectId());
```
dzou, line 78: If you wanted to avoid introducing the extra `getProjectId` on pubsubadmin, you could just compute the `topicShortName` and then continue passing `topicName` to `this.pubSubAdmin.createSubscription(..)` since it appears to accept both short and fully qualified names (so no need for `topicFullyQualifiedName` and getting the project Id).

```
String topicShortName = TopicName.isParsableFrom(topicName) ? TopicName.parse(topic).getTopic() : topicName;
```

dzou (approved):

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
meltsufin: +1, we just need to extract the short name.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -132,6 +138,6 @@ private Topic makeSureTopicExists(String topicName, boolean autoCreate) {
 			}
 		}
 
-		return topic;
+		return Optional.ofNullable(topic);
```
meltsufin, line 141: I don't think the topic could ever be null at this point. If it's null, and we don't auto-create, an exception will be thrown.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/PubSubTopicUtils.java
```diff
@@ -40,7 +41,9 @@ private PubSubTopicUtils() {
 	 * @param topic the topic name in the project or the fully-qualified project name
 	 * @param projectId the project ID to use if the topic is not a fully-qualified name
 	 * @return the Pub/Sub object representing the topic name
+	 * @deprecated Use toTopicName instead.
 	 */
+	@Deprecated
 	public static ProjectTopicName toProjectTopicName(String topic, @Nullable String projectId) {
```
meltsufin, line 47: Is `ProjectTopicName` really deprecated? It looks like it was at some point, but [not anymore](https://github.com/googleapis/java-pubsub/commit/bf316caa67b0e475d8d2148ea5d09d3e9cac3693).

meltsufin (changes requested):
Looks good overall! Some minor nits/questions.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
ttomsu: I thought so too, but take a look at the comment above the `LOGGER` line - it can, in fact, be null, and one the unit tests  failed for me when I assumed it always to be non-null.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/PubSubTopicUtils.java
ttomsu: `ProjectTopicName` isn't - the `createTopic` method is, though: https://github.com/googleapis/java-pubsub/blob/master/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/TopicAdminClient.java#L258

meltsufin: I see. I guess you can just do `TopicName name = TopicName.ofProjectTopicName("[PROJECT]", "[TOPIC]");` as they show in the javadoc.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
meltsufin: That has something to do with concurrency. Which test?

ttomsu: Right, but it could still return `null`, however unlikely. Hence, the `Optional`

meltsufin: It doesn't look like we even use the returned object anywhere. We might as well return nothing, and make the method just that tiny bit more efficient.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/PubSubTopicUtils.java
ttomsu: I un-deprecated this method since PTN isn't deprecated.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/PubSubTopicUtils.java
ttomsu: Done.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
ttomsu: Nice - I had to rework a few other things to get this to work, but it's a bit nicer now.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
ttomsu: I ended up having to use it in the validation check that ensures the subscription & topic match.

meltsufin: I think this whole situation with the topic potentially being created concurrently and being null is error-prone. Maybe the way to handle that `AlreadyExistsException` is to just try to fetch the topic instead of ignoring it. WDYT?

dzou (approved):

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
ttomsu: It's a very small chance of it happening, and if it does happen, it just means that a validation check is skipped. Wrapping it in an optional is a clean way to ensure the code doesn't throw any unexpected NPEs.

meltsufin: I think it's just more confusing to the caller of the method, but we can take care of it in a separate PR.

meltsufin (commented):
Do those code smells Sonar is report make sense?

PR #243, opened 2021-02-02
fix 'Info' level smells

ttomsu:
These smells were at the `INFO` level - all of which were `// TODO`s in the code. 

1. Firestore one: fixed as the TODO instructed

2. `list` method: just moved the comment to the Javadoc and renamd the param to be more clear what the function supports. A FR (not TODO) should be filed if we want to expand the functionality.

3. This isn't really a TODO - it's missing functionality because of a limit in GCS. Removed as not-really-actionable as a TODO


meltsufin (approved):
What are you starting with `info` level smells? Don't we have more major ones?

ttomsu:
They looked easy to knock out, so that's where I started.

PR #244, opened 2021-02-02
dedicated enhanced logging exception

ttomsu:
To be clear, I agree 100% with your [comment](https://sonarcloud.io/project/issues?id=GoogleCloudPlatform_spring-cloud-gcp&issues=AXdA21rr1IuZwn2BaFBQ&open=AXdA21rr1IuZwn2BaFBQ) that it's unnecessary, but since it captures a wide range of possible exceptions (my IDE counted at least 5), and it's only 6 additional lines to make Sonar happy, I figured it may be worth it. 

WDYT?


spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/StackdriverJsonLayout.java
```diff
@@ -281,4 +281,10 @@ private void addTraceId(ILoggingEvent event, Map<String, Object> map) {
 
 		add(StackdriverTraceConstants.TRACE_ID_ATTRIBUTE, this.includeTraceId, traceId, map);
 	}
+
+	static class EnhancedLoggingEventException extends RuntimeException {
```
meltsufin, line 285: Should it be `public`?

ttomsu: I don't think so - it's not instantiated anywhere else and I wouldn't expect any user to create one.

meltsufin: What about try-catch in user code?

ttomsu: I don't think we want this exception exposed as part of our API. If we do, then it should go into its own class.

meltsufin: Right, otherwise I don't see the benefit of introducing this exception class.

ttomsu: That's fine with me - I put this small PR together to spur the discussion, since it could be as small as 6LOC but as @dzou noted it's probably not worth a whole class.

spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/StackdriverJsonLayout.java
```diff
@@ -173,7 +173,7 @@ public void addLoggingEventEnhancer(String enhancerClassName) {
 			loggingEventEnhancers.add(clz.getDeclaredConstructor().newInstance());
 		}
 		catch (Exception ex) {
-			throw new RuntimeException(ex);
+			throw new EnhancedLoggingEventException(ex);
```
dzou, line 176: I see. I wasn't sure what Sonar's reason for the exception subtype was -- I imagined it is just so the user has more context right?

What do you think of just adding a message for the user instead to add context:

```
throw new RuntimeException("Failed to a StackdriverJsonLayout logging enhancer class: " + enhancerClassName, ex);
```

This might achieve the intended effect without needing the extra subclass.

ttomsu: SonarCloud's reasoning below:

```
Using such generic exceptions as Error, RuntimeException, Throwable, and Exception prevents calling methods from handling true, system-generated exceptions differently than application-generated errors.
```

So I think SC would still raise the same objection because you're still throwing a generic `RuntimeException` directly.

meltsufin: I kind of agree with Sonar here.

dzou: Ahh, I see. Ok let's just introduce the new `public` exception class then.

elefeint: I vote in favor of introducing new custom `RuntimeException` subtypes as normal practice. They make it easier to debug at a glance, and yes, allow special handling by downstream code. Making users rely on exception message parsing turns literal strings into a set-in-stone part of the API, which sucks.

spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/StackdriverJsonLayout.java
elefeint: I would make it its own public class. Yes, it will be a tiny one, but that's not dissimilar to marker interfaces. If the goal is to allow downstream classes to handle this exception* separately from other library-produced exceptions, then it has to officially be part of the public API.

* granted, I just "made it" a goal in the previous comment. Perhaps that's not a good goal? I think it is, but that's a "reasonable people can disagree" kind of thing.

elefeint (changes requested):
So, I personally feel uneasy whenever I throw a `RuntimeException` because that makes the end user rely on the error message if special handling is required.

spring-cloud-gcp-logging/src/test/resources/logback-test.xml
```diff
@@ -29,10 +29,29 @@
 		</encoder>
 	</appender>
 
+	<appender name="CONSOLE_JSON_SERVICE_CTX_MISSING" class="ch.qos.logback.core.ConsoleAppender">
```
elefeint, line 32: Is the name right? the context configuration seems to be present here.

elefeint (approved):

spring-cloud-gcp-logging/src/test/resources/logback-test.xml
ttomsu: It references an appender that has a missing `loggingEventEnhancer` - open to making it more clear, but it looks like Logback swallows a bunch of configuration exceptions (see test).

spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/StackdriverJsonLayout.java
meltsufin: Sorry for getting hung up on this, but I just looked at this method more carefully, and the only reason we would get an exception here is when an invalid class name is passed in. Either the class is not found, or it doesn't have an appropriate constructor. So, I think it might make a lot of sense to just throw an IllegalArgumentException with an additional message that the class that was passed in is invalid.

meltsufin (changes requested):

ttomsu:
Having written the test, I agree with @meltsufin 's reasoning.

meltsufin (approved):

PR #245, opened 2021-02-02
Fix a few 'major' smells that were not NPE related

ttomsu:
the CloudSqlOne is unreachable code, and the InterruptedException one points here for more information https://www.ibm.com/developerworks/java/library/j-jtp05236/index.html?ca=drs-#2.1

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
@@ -170,14 +170,6 @@ public class CloudSqlEnvironmentPostProcessor implements EnvironmentPostProcesso
 			// A resource might not be in the filesystem, but the Cloud SQL credential must.
 			File credentialsLocationFile = credentialsLocation.getFile();
 
-			// This should happen if the Spring resource isn't in the filesystem, but a URL,
-			// classpath file, etc.
-			if (credentialsLocationFile == null) {
-				LOGGER.info("The private key of the Google Cloud SQL credential must "
-						+ "be in a file on the filesystem.");
-				return;
-			}
-
 			System.setProperty(SqlCredentialFactory.CREDENTIAL_LOCATION_PROPERTY_NAME,
 					credentialsLocationFile.getAbsolutePath());
 
--- a/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/integration/outbound/BigQueryFileMessageHandler.java
+++ b/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/integration/outbound/BigQueryFileMessageHandler.java
@@ -203,10 +203,14 @@ public class BigQueryFileMessageHandler extends AbstractReplyProducingMessageHan
 			throw new MessageHandlingException(
 					message, "Failed to write data to BigQuery tables in message handler: " + this, e);
 		}
-		catch (InterruptedException | ExecutionException | TimeoutException e) {
+		catch (ExecutionException | TimeoutException e) {
 			throw new MessageHandlingException(
 					message, "Failed to wait for BigQuery Job to complete in message handler: " + this, e);
 		}
+		catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
+		}
+		return null;
 	}
 
 	private static InputStream convertToInputStream(Object payload) throws IOException {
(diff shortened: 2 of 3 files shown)


spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/integration/outbound/BigQueryFileMessageHandler.java
```diff
@@ -203,10 +203,14 @@ protected Object handleRequestMessage(Message<?> message) {
 			throw new MessageHandlingException(
 					message, "Failed to write data to BigQuery tables in message handler: " + this, e);
 		}
-		catch (InterruptedException | ExecutionException | TimeoutException e) {
+		catch (ExecutionException | TimeoutException e) {
 			throw new MessageHandlingException(
 					message, "Failed to wait for BigQuery Job to complete in message handler: " + this, e);
 		}
+		catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
```
meltsufin, line 211: Is this a potentially silent failure? How do we let the caller know that the write failed?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -170,14 +170,6 @@ private void setCredentialsFileProperty(Resource credentialsLocation) {
 			// A resource might not be in the filesystem, but the Cloud SQL credential must.
 			File credentialsLocationFile = credentialsLocation.getFile();
 
-			// This should happen if the Spring resource isn't in the filesystem, but a URL,
-			// classpath file, etc.
-			if (credentialsLocationFile == null) {
```
meltsufin, line 175: Are we saying that it actually cannot be null?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/integration/outbound/BigQueryFileMessageHandler.java
ttomsu: I think the calling code is supposed to handle it - the article listed in the description said if you can't throw the checked exception at least do this.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
ttomsu: It in fact cannot be null because there is a null check at the (one and only) call site.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/integration/outbound/BigQueryFileMessageHandler.java
meltsufin: The we shouldn't catch exception, right?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
meltsufin: The null check there is for getLocation() not getFile()

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/integration/outbound/BigQueryFileMessageHandler.java
ttomsu: It's a checked exception - you have to handle it.

meltsufin: or declare it thrown

ttomsu: It's an overridden method - so you can't change the signature.

meltsufin: Would it still make sense to throw the `MessageHandlingException` after the `interrupt()` in order not to introduce a breaking change?

meltsufin (approved):

PR #253, opened 2021-02-04
Add gRPC spoofed Spanner test

elefeint:
`ReplacingGrpcCallInterceptor` can be moved into a test support module later.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/spanner/ReplacingGrpcCallInterceptor.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/spanner/ReplacingGrpcCallInterceptor.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2021-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.google.cloud.spring.autoconfigure.spanner;
 
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/spanner/TransactionalTests.java
```diff
@@ -0,0 +1,111 @@
...
+import org.springframework.data.annotation.Id;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+import org.springframework.transaction.annotation.Transactional;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.Assert.assertNotNull;
+
+
+public class TransactionalTests {
```
dzou, line 28: Is it possible to put the tests inside the Spanner module instead of autoconfig?

dzou (commented):
Nice, looks really helpful. Didn't know you could intercept GRPC calls like this!

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/spanner/TransactionalTests.java
elefeint: No because we need the autoconfigured dependencies to be the same as in a real application, and referencing autoconfiguration module from spanner module would result in a circular dependency. Source: tried that first :D

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/spanner/ReplacingGrpcCallInterceptor.java
```diff
@@ -0,0 +1,137 @@
...
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+
+/**
+ * gRPC interceptor that stops the chain and returns a pre-set response object.
```
ttomsu, line 44: So it's basically a Mock? If so, have you looked at https://github.com/Fadelis/grpcmock ?

elefeint: It is, an I have not seen the `grpcmock` project before. It looks good! For spanner, we could use it by making the test think it's running against an emulator on localhost while actually using the mocked endpoint.

On the other hand, I have two concerns: 1) `grpcmock` starts a grpc server, which is more heavyweight than intercepting all requests in the test, and 2) this is bringing another dependency, and a single-person project for <50 lines of code. The irritating part of mocking is creating the response objects, which we still have to do (life is too short to construct a `PartialResultSet`, but here we are...)

So I could go either way.

ttomsu: I definitely agree on the risks of introducing a dependency than may become abandoned, especially given how fast Spring Boot moves and there may be an eventual incompatibility. 

That being said I don't know how many other tests you plan to write against this intercepter, and if lets say its 2+, I'd suggest using the library. Otherwise lets just keep it simple and use your mock/fake.

elefeint:
Closing this PR, since programmatically validating the scenario under test (that the session was returned to the pool) is nigh impossible without tons of reflection, at which point the test will be too fragile to survive.

PR #259, opened 2021-02-05
Update index-docinfo.xml Pivotal to Google

meltsufin:


docs/src/main/asciidoc/index-docinfo.xml
```diff
@@ -1,8 +1,8 @@
 <productname>Spring Cloud GCP</productname>
 <releaseinfo>{project-version}</releaseinfo>
 <copyright>
-	<year>2017-2019</year>
-	<holder>Pivotal Software, Inc.</holder>
+	<year>2017-2021</year>
+	<holder>Google LLC</holder>
```
ttomsu, line 5:
```suggestion
	<holder>Google, LLC</holder>
```

ttomsu (approved):

docs/src/main/asciidoc/index-docinfo.xml
meltsufin: I thought so too, but I see it used without the comma everywhere.

Issue #181, opened 2020-12-23
Can't create ConsumerDestination with group value only (Spring Cloud GCP PubSub Stream Binder)


buianhtai:
Currently i can't create consumer bean with supscription value from the `group` value only
I also check the current code ,it does'nt support subscription with value just `group` value 
as it's using <**topicName**>**.**<**group**> as subscription name
![image](https://user-images.githubusercontent.com/11797322/102968919-50688400-4527-11eb-853c-c2cd59a51163.png)

Below is my sample
spring.cloud.stream.bindings.consumeMessage-in-0.destination=order
spring.cloud.stream.bindings.consumeMessage-in-0.group=order-supscription

Can i create new comsumer which comsuming message from **order-supscription** supscription

elefeint:
You are correct; the Pub/Sub binder does not support specifying a pre-configured subscription name. This was originally done for consistency with the [common binding properties](https://docs.spring.io/spring-cloud-stream/docs/current/reference/html/spring-cloud-stream.html#_common_binding_properties) in Spring Cloud Stream.

It would be possible to support custom subscription names; we would consider it as a feature request.

Would you be interested in contributing a pull request? A new property would need to be introduced to `PubSubConsumerProperties` and used to override the default Spring Cloud name when creating the `PubSubInboundChannelAdapter` [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/8df7b52fcc5359dc7d45aa21db25f7c03f5d85c4/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java#L84).

PR #262, opened 2021-02-05
Allow specifying a custom subscription as a consumer endpoint

ttomsu:
Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/181

### DIFF
--- a/docs/src/main/asciidoc/spring-stream.adoc
+++ b/docs/src/main/asciidoc/spring-stream.adoc
@@ -72,15 +72,13 @@ These options are described in detail in the <<inbound-channel-adapter-using-pub
 spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.ack-mode=AUTO_ACK
 ----
 
-If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
-The topic name will be the same as the destination name, and the subscription name will be the destination name followed by the consumer group name.
+With automatic resource creation turned ON for a consumer, the library creates a topic and/or a subscription if they do not exist.
+The topic name becomes the same as the destination name, and the subscription name follows these rules (in order of precedence):
 
-Regardless of the `auto-create-resources` setting, you can control the subscriber with the following options (in order of precedence):
-
-* A pre-existing subscription (use `spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.subscriptionName`)
-* A consumer group that uses the topic name (use `spring.cloud.stream.bindings.events.group` to create a subscription named `<topicName>.<group>`)
-* If neither of the above are specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
-Then when the binder shuts down, all Pub/Sub subscriptions created for anonymous consumer groups will be automatically cleaned up.
+* A user-defined, pre-existing subscription (use `spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.subscriptionName`)
+* A consumer group using the topic name (use `spring.cloud.stream.bindings.events.group` to create a subscription named `<topicName>.<group>`)
+* If neither of the above are specified, the library creates an anonymous subscription with the name `anonymous.<destinationName>.<randomUUID>`.
+Then when the binder shuts down, the library automatically cleans up all Pub/Sub subscriptions created for anonymous consumer groups.
 
 For example, with this configuration:
 
@@ -90,7 +88,7 @@ For example, with this configuration:
 spring.cloud.stream.bindings.events.destination=myEvents
 spring.cloud.stream.gcp.pubsub.bindings.events.consumer.auto-create-resources=false
 ----
-Only an anonymous subscription named `anonymous.myEvents.a6d83782-c5a3-4861-ac38-e6e2af15a7be` will be created and later cleaned up.
+Only an anonymous subscription named `anonymous.myEvents.a6d83782-c5a3-4861-ac38-e6e2af15a7be` is created and later cleaned up.
 
 In another example, with the following configuration:
 
--- a/docs/src/main/asciidoc/spring-stream.adoc
+++ b/docs/src/main/asciidoc/spring-stream.adoc
@@ -75,13 +75,24 @@ spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.ack-mode=AUTO_A
 If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
 The topic name will be the same as the destination name, and the subscription name will be the destination name followed by the consumer group name.
 
-Regardless of the `auto-create-resources` setting, if the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
+Regardless of the `auto-create-resources` setting, you can control the subscriber with the following options (in order of precedence):
+
+* A pre-existing subscription (use `spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.subscriptionName`)
+* A consumer group that uses the topic name (use `spring.cloud.stream.bindings.events.group` to create a subscription named `<topicName>.<group>`)
+* If neither of the above are specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
 Then when the binder shuts down, all Pub/Sub subscriptions created for anonymous consumer groups will be automatically cleaned up.
 
-For example, for the following configuration, a topic named `myEvents` and a subscription called `myEvents.consumerGroup1` would be created.
-If the consumer group is not specified, a subscription called `anonymous.myEvents.a6d83782-c5a3-4861-ac38-e6e2af15a7be` would be created and later cleaned up.
+For example, with this configuration:
 
-IMPORTANT: If you are manually creating Pub/Sub subscriptions for consumers, make sure that they follow the naming convention of `<destinationName>.<consumerGroup>`.
+.application.properties
+[source]
+----
+spring.cloud.stream.bindings.events.destination=myEvents
+spring.cloud.stream.gcp.pubsub.bindings.events.consumer.auto-create-resources=false
+----
+Only an anonymous subscription named `anonymous.myEvents.a6d83782-c5a3-4861-ac38-e6e2af15a7be` will be created and later cleaned up.
+
+In another example, with the following configuration:
 
 .application.properties
 [source]
@@ -92,6 +103,10 @@ spring.cloud.stream.gcp.pubsub.bindings.events.consumer.auto-create-resources=tr
 # specify consumer group, and avoid anonymous consumer group generation
 spring.cloud.stream.bindings.events.group=consumerGroup1
 ----
+These resources will be created:
+
+* A topic named `myEvents`
+* A subscription named `myEvents.consumerGroup1`
 
 === Binding with Functions
 
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/properties/PubSubConsumerProperties.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/properties/PubSubConsumerProperties.java
@@ -32,6 +32,8 @@ public class PubSubConsumerProperties extends PubSubCommonProperties {
 
 	private Integer maxFetchSize = 1;
 
+	private String subscriptionName = null;
+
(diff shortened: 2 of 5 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/properties/PubSubConsumerProperties.java
```diff
@@ -32,6 +32,8 @@
 
 	private Integer maxFetchSize = 1;
 
+	private String customSubscription = null;
```
meltsufin, line 35: I think I would name it `customSubscriptionName` or just `subscriptionName`.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -79,7 +80,14 @@ public ConsumerDestination provisionConsumerDestination(String topicName, String
 
 		String subscriptionName;
 		Subscription subscription;
-		if (StringUtils.hasText(group)) {
+		if (StringUtils.hasText(properties.getExtension().getCustomSubscription())) {
+			String customName = properties.getExtension().getCustomSubscription();
+			subscriptionName = ProjectSubscriptionName.isParsableFrom(customName) ?
```
meltsufin, line 85: I think allowing fully qualified subscription names here could be problematic when combined with auto-creation of resources because the auto-created subscription would be in the wrong project.
However, I think you will avoid the problem if you don't parse out the subscription short name, but keep it as fully qualified.

meltsufin (changes requested):
Refdoc also needs to be updated accordingly.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -79,7 +80,14 @@ public ConsumerDestination provisionConsumerDestination(String topicName, String
...
+					ProjectSubscriptionName.parse(customName).getSubscription() :
+					customName;
+			subscription = this.pubSubAdmin.getSubscription(customName);
+		}
+		else if (StringUtils.hasText(group)) {
```
elefeint, line 90: We should also log a warning if _both_ the custom subscription name and group name are specified. And document in refdoc which takes precedence.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -79,7 +80,14 @@ public ConsumerDestination provisionConsumerDestination(String topicName, String
 
 		String subscriptionName;
 		Subscription subscription;
-		if (StringUtils.hasText(group)) {
+		if (StringUtils.hasText(properties.getExtension().getCustomSubscription())) {
```
elefeint, line 83: Can `properties.getExtension()` be null?

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -79,7 +80,14 @@ public ConsumerDestination provisionConsumerDestination(String topicName, String
...
+			String customName = properties.getExtension().getCustomSubscription();
```
elefeint, line 84: Might as well move this line to before `if` block, since you need to check its contents anyway.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
```diff
@@ -114,6 +114,29 @@ public void testProvisionConsumerDestination_specifiedGroupTopicInDifferentProje
...
+	@Test
+	public void testProvisionConsumerDestination_customSubscriptionDifferentProject() {
+		when(this.properties.getExtension()).thenReturn(this.pubSubConsumerProperties);
+		when(this.pubSubConsumerProperties.getCustomSubscription())
+				.thenReturn("projects/differentProject/subscriptions/my-custom-subscription");
+
+		PubSubConsumerDestination result = (PubSubConsumerDestination) this.pubSubChannelProvisioner
+				.provisionConsumerDestination("topic_A", "group_A", this.properties);
+
+		assertThat(result.getName()).isEqualTo("my-custom-subscription");
```
elefeint, line 137: So, this scenario is actually interesting. If we discard the different project prefix, will the binder will try to connect to `my-custom-subscription` in the wrong (current) project?

elefeint (dismissed):
Looking good! I am just not sure about what fully qualified subscription behavior will be in a different project. Did you try it?

meltsufin:
@elefeint I think we were reviewing in parallel :)

elefeint:
Yup. I went through my mailbox in reverse chronological order, and you moved in chronological order, and @ttomsu PR is where the two workflows intersected.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
ttomsu: I don't think so - [here](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java#L106-L113) appears to be where the extended properties are created.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
ttomsu: Good catch - done.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
ttomsu: Hmm... I think you're right - I assumed that the subscription created inside this method would used to create the subscriber, but I didn't look close enough.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
ttomsu: > when combined with auto-creation of resources

This is the the code that does the auto-provisioning. The FR was specifically to not do the auto-provisioning and use a pre-existing subscription (I expanded the requirement to be potentially outside the default GCP project)

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/properties/PubSubConsumerProperties.java
ttomsu: Done.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
ttomsu: Removed.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
ttomsu: Done.

docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -75,13 +75,24 @@ spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.ack-mode=AUTO_A
 If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
 The topic name will be the same as the destination name, and the subscription name will be the destination name followed by the consumer group name.
```
meltsufin, line 76: Since you talk about the subscription name below, maybe take out the second part of the sentence?

docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -75,13 +75,24 @@ spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.ack-mode=AUTO_A
...
 
-Regardless of the `auto-create-resources` setting, if the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
+Regardless of the `auto-create-resources` setting, you can control the subscriber with the following options (in order of precedence):
```
meltsufin, line 78:
```suggestion
Regardless of the `auto-create-resources` setting, the subscription name will be determined as follows (in order of precedence):
```

docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -75,13 +75,24 @@ spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.ack-mode=AUTO_A
...
+
+* A pre-existing subscription (use `spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.subscriptionName`)
```
meltsufin, line 80:
```suggestion
* If specified, `spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.subscriptionName` property
```

docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -75,13 +75,24 @@ spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.ack-mode=AUTO_A
...
+* A consumer group that uses the topic name (use `spring.cloud.stream.bindings.events.group` to create a subscription named `<topicName>.<group>`)
```
meltsufin, line 81:
```suggestion
* If specified, `spring.cloud.stream.bindings.events.group` property that results in a subscription named `<topicName>.<group>`
```

docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -75,13 +75,24 @@ spring.cloud.stream.gcp.pubsub.bindings.{CONSUMER_NAME}.consumer.ack-mode=AUTO_A
 If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
```
meltsufin, line 75: I think we should add a note here that, only anonymous subscriptions will be cleaned up.

meltsufin (changes requested):
Some clarifications in the doc

docs/src/main/asciidoc/spring-stream.adoc
ttomsu: It's mentioned twice below this.

docs/src/main/asciidoc/spring-stream.adoc
ttomsu: Ended up removing this text block

docs/src/main/asciidoc/spring-stream.adoc
ttomsu: removed

docs/src/main/asciidoc/spring-stream.adoc
ttomsu: With the "in order of precedence" above, I don't think it's necessary to have to say "if specified"

meltsufin (approved):

PR #266, opened 2021-02-08
Remove returning generic wildcard types

ttomsu:
From SC:

> It is highly recommended not to use wildcard types as return types. Because the type inference rules are fairly complex it is unlikely the user of that API will know how to use it correctly.

There are dozens and dozens of 'raw use of parameterized class' warnings all over Datastore, so I didn't feel bad about adding one more (it isn't clear how to fix this since there are a handful of `Builder` objects, the code smashed them all together, and the parameterized classes don't have a common base class).


ttomsu:
Looks like SC still doesn't like this change. Looks like we may have a giant refactor of Datastore in our future.

PR #269, opened 2021-02-08
Remove expressions that always evaluate to true/false

ttomsu:
There is [one more](https://sonarcloud.io/project/issues?assignees=ttomsu%40github&id=GoogleCloudPlatform_spring-cloud-gcp&open=AXZNafd0hMoO4rKzKcLg&resolved=false) of these that I'm not convinced will always return non-null, but would like some additional eyes to help out.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/DatastoreTemplate.java
@@ -411,7 +411,7 @@ public class DatastoreTemplate implements DatastoreOperations, ApplicationEventP
 		if (queryOptions.getCursor() != null) {
 			builder.setStartCursor(queryOptions.getCursor());
 		}
-		if (queryOptions.getSort() != null && persistentEntity != null) {
+		if (queryOptions.getSort() != null) {
 			queryOptions.getSort().stream()
 					.map((order) -> createOrderBy(persistentEntity, order))
 					.forEachOrdered((orderBy) -> builder.addOrderBy(orderBy));
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
@@ -223,9 +223,7 @@ public class PartTreeDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 		ProjectionInformation projectionInformation =
 				this.projectionFactory.getProjectionInformation(this.queryMethod.getReturnedObjectType());
 
-		if (projectionInformation != null &&
-				projectionInformation.getType() != this.entityType
-				&& projectionInformation.isClosed()) {
+		if (projectionInformation.getType() != this.entityType && projectionInformation.isClosed()) {
 			ProjectionEntityQuery.Builder projectionEntityQueryBuilder = Query.newProjectionEntityQueryBuilder();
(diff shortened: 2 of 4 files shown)


spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/DatastoreTemplate.java
```diff
@@ -411,7 +411,7 @@ public static void applyQueryOptions(StructuredQuery.Builder builder, DatastoreQ
 		if (queryOptions.getCursor() != null) {
 			builder.setStartCursor(queryOptions.getCursor());
 		}
-		if (queryOptions.getSort() != null && persistentEntity != null) {
+		if (queryOptions.getSort() != null) {
```
elefeint, line 414: Do we know that it's not null? I mean, if we get here, it's clearly not null because it would have NPEd on line 392 otherwise. But persistent entity comes from `this.datastoreMappingContext.getPersistentEntity(entityClass)`, and I don't see a null check anywhere.

elefeint (changes requested):
One question about nullability.

spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -223,9 +223,7 @@ else if (options.isCountingQuery() && this.tree.isDelete()) {
 		ProjectionInformation projectionInformation =
 				this.projectionFactory.getProjectionInformation(this.queryMethod.getReturnedObjectType());
 
-		if (projectionInformation != null &&
```
meltsufin, line 226: Hmm, how do you know it cannot be null here? Did Sonar make that claim?

meltsufin (approved):
Looks good, except for one place where I'm unsure.

spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
elefeint: This one I looked at -- projection factory always gets or creates a projection. Now, it beats me how _Sonar_ knew that, because it's buried a few layers deep in code.

https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/projection/ProxyProjectionFactory.java#L145

spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/DatastoreTemplate.java
ttomsu: IntelliJ also identified this one and has a helpful "Find Cause" action - and you've nailed it as to why: it's being dereferenced previously in this scope of code.

ttomsu:
Figured out the one in question above ^ It seems the whole package is annotated with `@NonNullApi` ([source](https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/convert/package-info.java#L6) and thus when it is assigned it is assumed to be not null.

We should probably revisit some things with this in mind.

spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
ttomsu: Yep ^

elefeint (approved):
I expect Sonar to pick up that potential NPE on top of the method, so we can fix it separately.

PR #271, opened 2021-02-08
Add a null guard around Datastore getPersistentEntity

dzou:
Adds a null check around `getPersistentEntity(...)` usage in Datastore.

If a user passes in an invalid class to be converted to a Datastore entity (such as `Integer`, `Boolean`, etc.) this will throw `IllegalArgumentException` now instead of yielding a null pointer.


meltsufin (approved):
If it's easy enough to add a test for this, that would be cool. If not, it's fine.

ttomsu:
I think once you fix the unit tests failures you'll find SonarCloud analysis should (will?) fail without tests (code coverage)

dzou:
> I think once you fix the unit tests failures you'll find SonarCloud analysis should (will?) fail without tests (code coverage)

Gotcha, will fix and add test.

dzou:
Will close this, Mike has a better fix.

PR #274, opened 2021-02-09
Fix some SonarCloud bugs

ttomsu:
Fixes https://sonarcloud.io/project/issues?id=GoogleCloudPlatform_spring-cloud-gcp&issues=AXZNaeudhMoO4rKzKcD-&open=AXZNaeudhMoO4rKzKcD- in the first commit 

and 

https://sonarcloud.io/project/issues?id=GoogleCloudPlatform_spring-cloud-gcp&issues=AXZNafzahMoO4rKzKcP9&open=AXZNafzahMoO4rKzKcP9 in the second

### DIFF
--- a/spring-cloud-gcp-security-firebase/src/main/java/com/google/cloud/spring/security/firebase/FirebaseJwtTokenDecoder.java
+++ b/spring-cloud-gcp-security-firebase/src/main/java/com/google/cloud/spring/security/firebase/FirebaseJwtTokenDecoder.java
@@ -21,7 +21,6 @@ import java.security.interfaces.RSAPublicKey;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.locks.ReentrantLock;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


meltsufin (approved):
When you commit, can you please add a comment about the change in locking behavior in the commit message? Thanks!

spring-cloud-gcp-security-firebase/src/main/java/com/google/cloud/spring/security/firebase/FirebaseJwtTokenDecoder.java
```diff
@@ -68,25 +67,20 @@
...
+
 	public FirebaseJwtTokenDecoder(RestOperations restClient, String googlePublicKeysEndpoint, OAuth2TokenValidator<Jwt> tokenValidator) {
 		this.restClient = restClient;
 		this.googlePublicKeysEndpoint = googlePublicKeysEndpoint;
 		this.tokenValidator = tokenValidator;
 	}
+
 	@Override
-	public Jwt decode(String token) throws JwtException {
+	public synchronized Jwt decode(String token) throws JwtException {
```
elefeint, line 80: We can decrease the synchronized surface by moving `synchronized` to the `refresh()` method -- that's what was attempted to be protected by the original lock.

The decoding itself can be done in parallel.

elefeint (changes requested):

spring-cloud-gcp-security-firebase/src/main/java/com/google/cloud/spring/security/firebase/FirebaseJwtTokenDecoder.java
ttomsu: Good call - done.

elefeint (approved):

PR #275, opened 2021-02-09
Cleanup in case of failures in PubSubAdmin constructor

meltsufin:
Fixes blocker SonarCloud finding:
"Use try-with-resources or close this "TopicAdminClient" in a "finally" clause."

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 the original author or authors.
+ * Copyright 2017-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


elefeint (dismissed):
Wait, so this will result in the admin clients getting closed immediately after getting created. I think Sonar might be wrong on this one.

We do want to close the client library objects in maybe `@PreDestroy`.

meltsufin:
> Wait, so this will result in the admin clients getting closed immediately after getting created. I think Sonar might be wrong on this one.
> 
> We do want to close the client library objects in maybe `@PreDestroy`.

Good catch! And the integration tests caught it too!

meltsufin:
> We do want to close the client library objects in maybe `@PreDestroy`.

The class already implements `AutoCloseable`. So, Spring will try to call the `close()` method already.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
```diff
@@ -54,11 +54,11 @@
 
 	private static final String NO_TOPIC_SPECIFIED = "No topic name was specified.";
 
-	private final String projectId;
+	private String projectId;
 
-	private final TopicAdminClient topicAdminClient;
+	private TopicAdminClient topicAdminClient;
```
ttomsu, line 59: IMO keeping these `final` is a greater benefit than the small amount of duplicate `Assert.notNull` code in the constructors, which, if I estimate correctly, may just be ensuring the project ID provider is not null and the project ID it provides is not null.

meltsufin: Agreed.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
```diff
@@ -68,19 +68,35 @@
...
-				TopicAdminClient.create(
-						TopicAdminSettings.newBuilder()
-								.setCredentialsProvider(credentialsProvider)
-								.build()),
-				SubscriptionAdminClient.create(
-						SubscriptionAdminSettings.newBuilder()
-						.setCredentialsProvider(credentialsProvider)
-						.build()));
+			CredentialsProvider credentialsProvider) throws PubSubException {
+		SubscriptionAdminClient subscriptionClient;
```
ttomsu, line 75:
```suggestion
```

ttomsu (approved):

PR #277, opened 2021-02-09
Handle NPE with getPersistentEntity in Datastore

meltsufin:


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/datastore/DatastoreKeyIdConverter.java
```diff
@@ -66,6 +66,11 @@ public String toRequestId(Serializable serializable, Class<?> aClass) {
 	public boolean supports(Class<?> entityType) {
 		// This ID converter only covers the Datastore key type. Returning false here causes the
 		// default converter from Spring Data to be used.
-		return this.datastoreMappingContext.getPersistentEntity(entityType).getIdProperty().getType().equals(Key.class);
+		try {
```
dzou, line 69: I think catching exceptions is expensive compared to just checking the values. Maybe it's preferable to just check the values directly?

```
result = this.datastoreMappingContext.getPersistentEntity(entityType)

if (result == null) {
  return false;
} else {
  return result.getIdProperty().getType().equals(Key.class);
}
```

meltsufin: `getIdProperty()` can also return null :(

dzou: Ahh ok.

dzou (approved):

PR #282, opened 2021-02-10
Fix SC: Complete switch case statements with all options or add default case

ttomsu:


spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/SpannerMutationFactoryImpl.java
```diff
@@ -177,7 +177,7 @@ private void verifyChildHasParentId(SpannerPersistentEntity parentEntity,
 	}
 
 	private WriteBuilder writeBuilder(Op op, String tableName) {
-		Mutation.WriteBuilder builder = null;
+		Mutation.WriteBuilder builder;
```
elefeint, line 180: Can't wait for switch statements returning values. Is spacing correct on the case statements? They look oddly aligned.

elefeint (approved):

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/SpannerMutationFactoryImpl.java
ttomsu: Yep - they're just tabs.

PR #285, opened 2021-02-10
Drop silly uses of String.format

ttomsu:


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/DefaultCloudSqlJdbcInfoProvider.java
```diff
@@ -54,7 +54,7 @@ public String getJdbcUrl() {
 				this.properties.getDatabaseName(),
 				this.properties.getInstanceConnectionName());
 		if (StringUtils.hasText(properties.getIpTypes())) {
-			jdbcUrl = String.format(jdbcUrl + "&ipTypes=%s", properties.getIpTypes());
+			jdbcUrl += "&ipTypes=" + properties.getIpTypes();
```
elefeint, line 57: Optional: I'd maybe add parentheses for readability.

elefeint (approved):

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/DefaultCloudSqlJdbcInfoProvider.java
ttomsu: Nack

PR #286, opened 2021-02-10
Add additional null checks for Spanner

dzou:
Add additional null checks for Spanner module.


elefeint (changes requested):
There might be an opportunity here for a helper method/class that takes a spanner mapping context, and either retrieves persistent entity or throws.

ttomsu:
+1 to Elena's comment - plus you may start getting dinged on repeating the same String and told to make it a constant

dzou:
> There might be an opportunity here for a helper method/class that takes a spanner mapping context, and either retrieves persistent entity or throws.

Done -- I introduced a method called `context.getPersistentEntityOrFail(..)`

elefeint (approved):
You can even just make it an overridden version of `.getPersistentEntity()`. Failure is always an option for any method.

dzou:
> You can even just make it an overridden version of `.getPersistentEntity()`. Failure is always an option for any method.

I tried to just override the `.getPersistentEntity()` directly at first, however it seems there times where we check if it is `null` or not; there was at least one test that failed. (I guess it's just the majority of cases (and not all) in which we call this method unsafely.)

PR #288, opened 2021-02-11
Remove extra parentheses around single arg lambdas

ttomsu:
Gnarly regex that I'm not proud of (I am a bit!)

Find: `\(([^\s^,^(.]+?)\) ->`
Replace: `$1 ->`

Breakdown - Lazily capture the contents of a set of parentheses that don't contain a space or a comma, but have at least once character (i.e. not `(foo, bar) ->`,  `(Foo foo) ->`, nor `() ->`). Took a couple of tries but I think I found more than even SonarCloud identified (I don't think SC applies this rule to tests).

### DIFF
--- a/docs/src/main/asciidoc/spring-stream.adoc
+++ b/docs/src/main/asciidoc/spring-stream.adoc
@@ -185,7 +185,7 @@ public class SinkExample {
         return args -> {
             // This will poll only once.
             // Add a loop or a scheduler to get more messages.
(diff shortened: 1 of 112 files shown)


elefeint (approved):
Using the chaotic power of regex for good, I see :)
Can save on some carets next time; leading caret negates all characters in a class `\(([^\s,(.]+?)\)`.

PR #295, opened 2021-02-11
Refactor fancy functional fluent flow

ttomsu:
I think it was a bit of force fitting some Functional magic in where a simpler solution is a little more understandable. Also fixes a SC identified issue.


dzou (approved):
Nice.

PR #296, opened 2021-02-11
Make various fields transient

ttomsu:
I'm not 100% sure on this one, but it's my understanding that `Serializable` objects must have all of their non-`transient` fields also be `Serializable`, and if they can't/aren't, they are marked `transient` and thus ignored. 

Since many of these objects are java language constructs (`Supplier`, `Optional`) or from dependent libraries (stubs), there's not much we can do to make the objects `Serializable`, thus I've marked them `transient`


dzou (approved):
We could just mark the issues as won't fix if `transient` is just adding noise. I doubt serialization works anyway for some of these objects (like ReactiveFirestoreTransactionManager probably won't be able to recreate the firestore stub after serialization).

elefeint:
I think java will throw `NotSerializableException` if a field of a serializable class is not serializable.

The abstract platform manager says 

> The state of this class is serializable, to allow for serializing the
> * transaction strategy along with proxies that carry a transaction interceptor.
> * It is up to subclasses if they wish to make their state to be serializable too.

So I agree that we can won't-fix those Sonar findings, and not support serialization.

ttomsu:
Dropping this PR and closing those issues as "won't fix"

PR #297, opened 2021-02-11
Fix remaining Spanner Null Pointer Errors

dzou:
Fixes the remaining NPE in the Spanner module.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
@@ -407,7 +407,7 @@ public class ConverterAwareMappingSpannerEntityWriter implements SpannerEntityWr
 
 		if (!valueSet) {
 			throw new SpannerDataException(String.format(
-					"Unsupported mapping for type: %s", propertyValue.getClass()));
+					"Unsupported mapping for type: %s", propertyType));
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -407,7 +407,7 @@ else if (property.getAnnotatedColumnItemType() != null) {
 
 		if (!valueSet) {
 			throw new SpannerDataException(String.format(
-					"Unsupported mapping for type: %s", propertyValue.getClass()));
```
elefeint, line 410: In principle, propertyType can be null. Not sure whether it would happen in practice.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SpannerQueryLookupStrategy.java
```diff
@@ -79,6 +79,7 @@ public RepositoryQuery resolveQuery(Method method, RepositoryMetadata metadata,
 			ProjectionFactory factory, NamedQueries namedQueries) {
 		SpannerQueryMethod queryMethod = createQueryMethod(method, metadata, factory);
 		Class<?> entityType = getEntityType(queryMethod);
+
```
elefeint, line 82: revert whitespace?

elefeint (approved):

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
dzou: Every property has to have a type right (meaning it would be non-null)? Not sure. But this will solve our NPE here from `.getClass()` though.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SpannerQueryLookupStrategy.java
dzou: Done.

PR #310, opened 2021-02-16
Break up the monster Spanner integration test

ttomsu:
Turns out the DB cleanup code was broken too - fixed that as well.

I suggest reviewing one commit at a time.

Context: Was fixing the error flagged by [Test methods should not contain too many assertions](https://sonarcloud.io/organizations/googlecloudplatform/rules?open=java%3AS5961&rule_key=java%3AS5961), but this unit test was heavy on the test but not the _unit_ part. It contained 89 assertions in this one method (max is set at 25).


dzou (approved):
Nice, this is much cleaner.

PR #313, opened 2021-02-16
Reduce usage of deprecated classes

ttomsu:
This gets some of the ["@Deprecated" code should not be used](https://sonarcloud.io/organizations/googlecloudplatform/rules?open=java%3AS1874&rule_key=java%3AS1874) violations, but not all of them. I've noted which ones I think are breaking changes for our users in the comments of each one: https://sonarcloud.io/project/issues?id=GoogleCloudPlatform_spring-cloud-gcp&resolved=false&rules=java%3AS1874

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/publisher/PubSubPublisherTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/publisher/PubSubPublisherTemplate.java
@@ -33,6 +33,8 @@ import org.springframework.util.Assert;
 import org.springframework.util.concurrent.ListenableFuture;
 import org.springframework.util.concurrent.SettableListenableFuture;
 
+import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
+
 /**
  * Default implementation of {@link PubSubPublisherOperations}.
  * <p>The main Google Cloud Pub/Sub integration component for publishing to topics.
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/publisher/PubSubPublisherTemplate.java
```diff
@@ -119,7 +121,7 @@ public void onSuccess(String result) {
 				settableFuture.set(result);
 			}
 
-		});
+		}, directExecutor());
```
dzou, line 124: Is `directExecutor()` equivalent to `Runnable::run`?

Maybe you can just use `Runnable::run` here and avoid the guava import.

ttomsu: It looks pretty similar, but in this case I just copied it straight from the deprecated implementation: https://github.com/googleapis/api-common-java/blob/master/src/main/java/com/google/api/core/ApiFutures.java#L58

Also note that this isn't Guava - it's `api-common-java`, which I assume comes with all of our client libraries. That being said, I've now added two entries to our checkstyle-exceptions, and I'm not able to find where this `IllegalImport` rule keeps catching these imports.

dzou (approved):

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/publisher/PubSubPublisherTemplate.java
dzou: Gotcha, sounds good.

PR #327, opened 2021-02-19
Use date in dependencies cache and refresh daily

ttomsu:
Following up from yesterdays conversation about the GHA cache - this changes the "purge cache" job to a "warm cache" job because, as we learned, we cannot _update_ an existing cache - we can only create wholly new ones. 

So this job runs right at the beginning of every day (UTC time, obv, we're not savages after all) and stores the cache based on the YYYY-MM-DD. I've also removed the backup `restore-keys` option as I have no idea what is in this `Linux-maven-` cache. It likely doesn't have much of what we need so lets save the time downloading it.

If this works out as expected I can make similar changes to our other jobs.

### DIFF
--- a/.github/workflows/unitTests.yaml
+++ b/.github/workflows/unitTests.yaml
@@ -18,16 +18,21 @@ jobs:
       matrix:
         java: [8, 11, 15]
     steps:
+    - name: Get current date
+      id: date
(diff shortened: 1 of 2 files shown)


dzou (approved):

meltsufin (approved):
Looks good! Let's give it a try.

PR #264, opened 2021-02-05
Update BOM version in README.adoc to 2.0.1

meltsufin:

PR #328, opened 2021-02-19
update BOM version in /docs readme

meltsufin:
Addressing https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/264#issuecomment-782169786.

@ttomsu I'm not really sure why we have a readme in /docs and how are we supposed to use it.


ttomsu:
It looks like the parent POM for the docs/ module invokes this class: https://github.com/spring-cloud/spring-cloud-build/blob/master/docs/src/main/java/org/springframework/cloud/internal/asciidoctor/ReadmeMain.java to (I'm guessing) resolve some references linked in the source .adoc so they work in the root .adoc

I also think having two of these is silly, and a comparison between the src file and the generated file showed little difference. I think we should remove the src README and only rely on the root one.

ttomsu (approved):

dzou (approved):

Issue #236, opened 2021-02-01
Allow custom subscription name for Pub/Sub healthcheck


elefeint:
Issue ported from spring-cloud/spring-cloud-gcp#2630.

Currently, the Pub/Sub healthcheck attempts a pull from a nonexistent subscription. This has two negative consequences:

1. The pull takes longer
2. Error logs get polluted

Allowing a pull from a known existing subscription would ameliorate issues like spring-cloud/spring-cloud-gcp#2628 and spring-cloud/spring-cloud-gcp#2570.

elefeint:
Here are the detailed steps that will need to be taken for this task:

1. Introduce a new properties class `GcpPubSubHealthIndicatorProperties` in the `com.google.cloud.spring.autoconfigure.pubsub.health` package. You can model the class after `GcpPubSubProperties`, but it will be much simpler, only containing one property.
1. Introduce a new wrapper type that will contain `PubSubTemplate` and the healthcheck subscription name from the properties. [` PubSubHealthIndicatorAutoConfiguration`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/fd96f3ab1d0df3c1dc0eca6ddc0db52b3fa765f1/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfiguration.java#L52) will need to use the new type as the second parameterized type, and it will need to create/provide a bean of that type.
1. Update [`PubSubHealthIndicator`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java) class to accept the new wrapper type instead of `PubSubTemplate`.
1.  Update/add tests for any lines of code affected.
1. Update the [healthcheck section](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/master/docs/src/main/asciidoc/pubsub.adoc#cloud-pubsub-health-indicator) in Pub/Sub documentation

elefeint:
cc/ @ThusithaDJ

ThusithaDJ:
@elefeint, thanks a lot for the information and the opportunity. I'll let you know when I'm having some doubts.

ThusithaDJ:
[...]

PR #330, opened 2021-02-21
Support custom subscription name for Pub/Sub health check

patpe:
First, the main issue that I want to address can be seen by looking at https://github.com/Predictly/gcp-pubsub-health where Kubernetes put's the pod in an restart loop since it fails to respond to `/actuator/health`.

Here are my thoughts that went into designing a solution:

1. Started investigating adding pubsub state to PubSubTemplate, PubSubSubscriberTemplate or PubSubPublisherTemplate, decided not to go down this route since it would add code and variables that would have nothing to do with the main flow of these classes
2. Decided to handle the issues I saw by pulling messages async in the existing `PubSubHealthIndicator` which will make it possible to return with an unknown state during the health check even if the application is under heavy load
3. Added support for customizing both health check subscription and timeout according to @elefeint's initial ideas

My [concerns ](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/236#issuecomment-782733200) related to this could be mitigated by documenting how these attributes in the new `PubSubHealthIndicatorProperties` should NOT be used

I have tested my code in the `gcp-pubsub-health` application and it removes the issue with Kubernetes restarting the application due to liveness timeouts, under load it responds based on the timeout configured in the new property `spring.cloud.gcp.pubsub.health.timeout-millis` with the following

```javascript
"pubSub": {
  "status": "UNKNOWN",
  "details": {
    "error": "java.util.concurrent.TimeoutException: null"
  }
}
```

If this looks good I can go ahead and update the docs as well. First time contributing so any feedback is welcome.


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthTemplate.java
```diff
@@ -0,0 +1,57 @@
...
+
+	private PubSubTemplate pubSubTemplate;
+
+	private String subscription;
+
+	private long timeoutMillis;
+
+	public PubSubHealthTemplate(PubSubTemplate pubSubTemplate, String subscription,
+			long timeoutMillis) {
+		super();
```
ttomsu, line 36: Not necessary to invoke `super()` here.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
```diff
@@ -16,55 +16,73 @@
...
+		catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
+			builder.withException(e).unknown();
+		}
+		catch (ExecutionException e) {
+			Throwable t = e.getCause();
+			if (t instanceof ApiException) {
+				ApiException aex = (ApiException) t;
+				Code errorCode = aex.getStatusCode().getCode();
+				if (errorCode == StatusCode.Code.NOT_FOUND || errorCode == Code.PERMISSION_DENIED) {
```
ttomsu, line 70: Hmm... Given that we're now encouraging "bring your own subscription to use for health checks", do we really want to say that a not found is ok? Seems to me like this would mask a misconfiguration error, and the error would be swallowed

ttomsu (commented):
Mostly looks good - great work for a first pass!

Left a comment for @meltsufin / @elefeint re: which return codes we want use for up/down signaling.

dzou:
Would it be better to keep the custom subscription as a field of the `PubSubHealthIndicator` or introduce a new wrapper class like in here (via `PubSubHealthTemplate`)?

On my first pass I thought keeping the custom subscription in the `PubSubHealthIndicator` as an optional string would be simpler to avoid introducing a new class, but am curious to hear everyone's thoughts.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
meltsufin: I don't know if we want to *require* users to bring their own subscription. This certainly would need to be documented well to clear up the confusion on why setting a custom subscription for health check if optional.

That being said, if the user provides a subscription, `NOT_FOUND` or `PERMISSION_DENIED` would indeed be error conditions.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
```diff
@@ -16,55 +16,73 @@
...
+		Assert.notNull(pubSubHealthTemplate, "PubSubHealthTemplate can't be null");
+		this.pubSubHealthTemplate = pubSubHealthTemplate;
 	}
 
 	@Override
 	protected void doHealthCheck(Health.Builder builder) {
+		ListenableFuture<List<AcknowledgeablePubsubMessage>> future = this.pubSubHealthTemplate.pullAsync();
 		try {
-			this.pubSubTemplate.pull("subscription-" + UUID.randomUUID().toString(), 1, true);
+			future.get(this.pubSubHealthTemplate.getTimeoutMillis(), TimeUnit.MILLISECONDS);
```
meltsufin, line 59: Wouldn't you want `builder.up()` after this line?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthTemplate.java
```diff
@@ -0,0 +1,57 @@
...
+		this.pubSubTemplate = pubSubTemplate;
+		this.subscription = subscription;
```
meltsufin, line 38: This subscription should be validated as existing as early as possible. (Fail fast)

meltsufin (changes requested):
I don't see much benefit in making `PubSubHealthTemplate` part of our public API.
It can either be subsumed under `PubSubHealthIndicator`, or just made package-private. I would be fine with either.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
elefeint: I agree that custom subscription should be optional -- for backwards compatibility, and for users who can't create a dedicated subscription due to permission structure.

So there are two paths through the healthcheck, which I'd probably handle as follows:
1) a random subscription expected to return NOT_FOUND or PERMISSION_DENIED. Any other exception would signal downtime.
2) a custom subscription, for which the expected "up" signal is *not getting any exception*. Either receiving messages or pulling an empty batch would be "up". But getting not found OR permission denied should be unexpected. My reasoning is that if a team/org has enough permissions to create a dedicated subscription, they are likely able to set permissions appropriately.

elefeint: The if/else will become even more unwieldy with my suggestions, btw, so you may want a helper method.

patpe:
Continued working on this and got to a point where it started to get a bit confusing. Consider the following scenario:

`PubSubHealthIndicatorAutoConfiguration#pubSubHealthContributor` is called with a map of more than one `PubSubTemplate` and the properties contains a specified subscription. What does this setup mean?

Will it be possible to have several `PubSubTemplate` beans each pointing to different projects? If so, how should we use the specified subscription? Throw `IllegalArgumentException` if `subscription` specified and there are more than one `PubSubTemplate` bean?

Validate in `PubSubHealthIndicatorAutoConfiguration#pubSubHealthContributor` which of the `PubSubTemplate` beans can read from the specified subscription, let the rest pull from `UUID.randomUUID()`?

meltsufin:
@patpe One thing to note is that subscription name could be relative or fully-qualified as in `projects/<project_name>/subscriptions/<subscription_name>}`. If a relative one is used, it will automatically be routed to the project of the `PubSubTemplate` bean. Otherwise, the user is explicitly asking to connect to a specific project, which will work seamlessly regardless of which project is defined for the specific `PubSubTemplate` bean.
So, I don't really see much of an issue here.

patpe:
@meltsufin I understand your point, but not sure . My concern is the scenario where multiple `PubSubTemplate` beans are configured and they have separate `CredentialProvider` that does eventually not grant access to the same project. This is, in theory, possible to do and if there are more than one `PubSubTemplate` bean a likely reason. Think of a proxy service that reads from one project and writes to the project it is itself running in, cross organization or department in a business.

Which of the `PubSubTemplate` (i.e. GCP projects) should the health check assume that the configured subscription is in? Or am I missing something fundamental about how the GCP libraries work under the hood?

patpe:
Updated PR which now supports the following

1. User can configure a custom subscription and timeout for health check
2. If specified, the subscription has the exist in all available pubSubTemplates or the application will not start (`BeanInitializationException` will be thrown)
3. Health checks against PubSub done async and expected to return response within specified timeout (1000ms default)
4. If a custom subscription has been specified, the async pull against pubsub will signal unknown for `TimeoutException` and down for any other exception
5. If no custom subscription has been specified, the async pull is done against a random subscription name and will signal ok for `NOT_FOUND` and `PERMISSION_DENIED` in ApiException, unknown for TimeoutException and down for anything else

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthTemplate.java
```diff
@@ -0,0 +1,95 @@
...
+	 */
+	private String subscription;
+
+	/**
+	 * Timeout when performing health check.
+	 */
+	private long timeoutMillis;
+
+	PubSubHealthTemplate(PubSubTemplate pubSubTemplate, String userSubscription, long timeoutMillis) {
+		this.pubSubTemplate = pubSubTemplate;
```
meltsufin, line 63: Assert that `pubSubTemplate` is not null?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthTemplate.java
```diff
@@ -0,0 +1,95 @@
...
+		this.userSubscription = userSubscription;
```
meltsufin, line 64: Looks like `userSubscription` could just be a boolean flag, since you're storing it in `subscription` anyway.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthTemplate.java
```diff
@@ -0,0 +1,95 @@
...
+		this.subscription = userSubscription;
+		if (!StringUtils.hasText(this.subscription)) {
+			this.subscription = UUID.randomUUID().toString();
+		}
+		this.timeoutMillis = timeoutMillis;
+	}
+
+	void pullAndAckAsync() throws InterruptedException, ExecutionException, TimeoutException {
```
meltsufin, line 72: Since you're doing `fugure.get()`, I think it's confusing to call this method `...Async`.

Alos, I would probably just call it `probeHealth()` or something like that to keep the implementation details (which might change in the future) leaking into the method name.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthTemplate.java
```diff
@@ -0,0 +1,95 @@
...
+		ListenableFuture<List<AcknowledgeablePubsubMessage>> future = pubSubTemplate.pullAsync(this.subscription, 1, true);
+		List<AcknowledgeablePubsubMessage> messages = future.get(timeoutMillis, TimeUnit.MILLISECONDS);
+		messages.forEach(AcknowledgeablePubsubMessage::ack);
```
meltsufin, line 75: What's the purpose of acking the message(s)?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthTemplate.java
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	boolean isExpectedExecutionException(ExecutionException e) {
```
meltsufin, line 78: Maybe name it something like `isHealthyException()`?

meltsufin (changes requested):
Was the intent to make the health-check async? If so, currently it doesn't seem to be doing it because of the `future.get()`. You would instead need to register a success and a failure callback on the future, which will in turn set the heath status. This will probably be easier to do if you merged the `PubSubHealthTemplate` with the `PubSubHealthIndicator` into a single  class.

ttomsu:
Also please take a look at the Sonar Cloud report for the [5 code smells](https://sonarcloud.io/project/issues?id=GoogleCloudPlatform_spring-cloud-gcp&pullRequest=330&resolved=false&types=CODE_SMELL) - they should be pretty simple to fix.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthTemplate.java
patpe: My reason for acking them is just hygiene. If a user sets up a user subscription for health checks and actually publishes messages to it, I would assume that they expect the message to be acked by the health check as well.

And if we don't ack, the subscription will start to show `expired` in GCP stats as well which will draw unnecessary attention to a problem that has been artifically created by the health check.

patpe:
Thanks for the feedback, incorporated your suggestions. ⛷️ for a week, hence late reply.

> Was the intent to make the health-check async? If so, currently it doesn't seem to be doing it because of the `future.get()`.

The intent was never to make the health check async and this is just poor naming of the PR from me. The intent was to use the async functionality of the PubSub API to guarantee a response from the health check in a configurable amount of time, thereby avoiding not returning a response.

That being said, if making the health probe async is a better solution I can continue working with that as a target. My gut feeling about making it async was that the use case did not merit introducing that level of complexity.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
```diff
@@ -16,55 +16,66 @@
...
  *
  * @since 1.2.2
  */
 public class PubSubHealthIndicator extends AbstractHealthIndicator {
 
-	private final PubSubTemplate pubSubTemplate;
+	private final PubSubHealthTemplate pubSubHealthTemplate;
 
-	public PubSubHealthIndicator(PubSubTemplate pubSubTemplate) {
+	public PubSubHealthIndicator(PubSubHealthTemplate pubSubHealthTemplate) {
```
meltsufin, line 50: Since `PubSubHealthTemplate` is now a package-private class, we can't make it part of the public interface here. So, we'll have to accept a bean of `PubSubHealthTemplate` like before and just create the `PubSubHealthTemplate` internally. Alternatively, you can just move the methods from the `PubSubHealthTemplate` class into this class and remove `PubSubHealthTemplate` altogether.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
```diff
@@ -16,55 +16,66 @@
...
- * Default implemenation of
+ * Default implementation of
  * {@link org.springframework.boot.actuate.health.HealthIndicator} for Pub/Sub. Validates
- * if connection is successful by looking for a random generated subscription name that
- * won't be found. If no subscription is found we know the client is able to connect to
- * GCP Pub/Sub APIs.
+ * if connection is successful by pulling message asynchronously from the pubSubHealthTemplate.
+ *
+ * If a custom subscription has been specified, this health indicator will only signal up
+ * if messages are successfully pulled and acknowledged.
```
meltsufin, line 34: I think it's worth noting here that the custom subscription doesn't have to have messages on it. Also, it's important for users to realize that the topic that the custom subscription is for needs to be dedicated to the health check. I'm very worried that users will accidentally create a health subscription to a topic that is also used for other purposes and potentially lose messages.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfiguration.java
```diff
@@ -49,13 +55,42 @@
...
+		String subscription = gcpPubSubHealthProperties.getSubscription();
+		Long timeoutMillis = gcpPubSubHealthProperties.getTimeoutMillis();
+
+		Map<String, PubSubHealthTemplate> pubSubHealthTemplates = new HashMap<>();
+		pubSubTemplates.forEach((k, v) -> pubSubHealthTemplates.put(k, new PubSubHealthTemplate(v, subscription, timeoutMillis)));
+		pubSubHealthTemplates.forEach(this::validatePubSubHealthTemplate);
+		return createContributor(pubSubHealthTemplates);
+	}
+
+	private void validatePubSubHealthTemplate(String name, PubSubHealthTemplate pubSubHealthTemplate) {
```
meltsufin, line 75: Can this bit of code duplication with the `PubSubHealthIndicator` be avoided?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthTemplate.java
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.util.StringUtils;
+import org.springframework.util.concurrent.ListenableFuture;
+
+/**
+ * Template for performing health checks against GCP Pub/Sub.
+ *
+ * @author Patrik Hörlin
+ *
+ */
+class PubSubHealthTemplate {
```
meltsufin, line 41: I know at first I said it probably didn't matter that we're introducing this additional class, as long as it's package-private, but I'm now seeing more complexity introduced due to it. What's the benefit you're seeing from having this class?

meltsufin (changes requested):

patpe:
Refactored so that `PubSubHealthTemplate` isn't needed anymore and added javadoc to `PubSubHealthIndicator` about the behaviour when pulling messages.

I have some thoughts about the way the health indicator is validated. Should it 
1. be moved to the `PubSubHealthIndicator` constructor?
2. be removed (no other HealthIndicator seems to have this behaviour)?

@meltsufin I agree with your concerns and I have some doubts about the current implementation in this PR. If forced to choose between the following two scenarios 
1. pulling and acking a message from a business subscription which has been accidentally configured as health check subscription
2. not ack:ing messages on a health check subscription that will produce side effects in GCP statistics (expired metric)

I would prefer to trigger option 2 since it is the lesser of the two evils. Should we perhaps remove the `ack`?

meltsufin:
> I have some thoughts about the way the health indicator is validated. Should it
> 
> 1. be moved to the `PubSubHealthIndicator` constructor?
> 2. be removed (no other HealthIndicator seems to have this behaviour)?
> 

The nice thing about validation in the constructor is that we force any problems to surface during app startup, but on the other hand, if no other health indicator does it, I think it should be fine to just remove it.

> @meltsufin I agree with your concerns and I have some doubts about the current implementation in this PR. If forced to choose between the following two scenarios
> 
> 1. pulling and acking a message from a business subscription which has been accidentally configured as health check subscription
> 2. not ack:ing messages on a health check subscription that will produce side effects in GCP statistics (expired metric)
> 
> I would prefer to trigger option 2 since it is the lesser of the two evils. Should we perhaps remove the `ack`?

What do you think of a compromise where we just add a configuration property for acking and default it to `false`?

meltsufin:
@patpe I would like to help you polish this PR. Would you mind giving me access to your fork?

patpe:
> @patpe I would like to help you polish this PR. Would you mind giving me access to your fork?

Thanks @meltsufin, I have given you access and you should be receiving a notification about it soon. Combining this PR with work proved a greater challenge than I anticipated, will dive into it again right now.

patpe:
@meltsufin Added support for toggling acknowledge of messages on or off, off by default.

meltsufin (approved):
@patpe I've polished up the pull request and added some reference documentation. Please see if looks good to you.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
```diff
@@ -16,55 +16,174 @@
...
+	 */
+	private final String subscription;
+
+	/**
+	 * Timeout when performing health check.
+	 */
+	private final long timeoutMillis;
+
+	/**
+	 * If messages pull should be acknowledged or not.
```
elefeint, line 86:
```suggestion
	 * Whether pulled messages should be acknowledged.
```

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
```diff
@@ -16,55 +16,174 @@
...
 		super("Failed to connect to Pub/Sub APIs. Check your credentials and verify you have proper access to the service.");
-		Assert.notNull(pubSubTemplate, "PubSubTemplate can't be null");
+		Assert.notNull(pubSubTemplate, "pubSubTemplate can't be null");
 		this.pubSubTemplate = pubSubTemplate;
+		this.specifiedSubscription = StringUtils.hasText(healthCheckSubscription);
+		if (this.specifiedSubscription) {
+			this.subscription = healthCheckSubscription;
+		}
+		else {
+			this.subscription = UUID.randomUUID().toString();
```
elefeint, line 99: Subscription names must start with a letter. I'd suggest a prefix of "spring-cloud-gcp-healthcheck".
https://googleapis.dev/java/google-cloud-pubsub/latest/com/google/pubsub/v1/Subscription.Builder.html#setName-java.lang.String-

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
```diff
@@ -16,55 +16,174 @@
...
+			pullMessage();
+		}
+		catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
+			validationFailed(e);
+		}
+		catch (ExecutionException e) {
+			if (!isHealthyException(e)) {
+				validationFailed(e);
+			}
```
elefeint: I'd add a comment after the `if` block for `// ignore expected exceptions` in lieu of the `else` clause.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
```diff
@@ -46,37 +64,131 @@
 			.withBean(GcpProjectIdProvider.class,  () -> () -> "fake project")
 			.withBean(CredentialsProvider.class, () -> () -> mock(Credentials.class));
 
+	@SuppressWarnings("unchecked")
 	@Test
-	public void healthIndicatorPresent() {
+	public void healthIndicatorPresent() throws Exception {
+		PubSubTemplate mockPubSubTemplate = mock(PubSubTemplate.class);
+		ListenableFuture<List<AcknowledgeablePubsubMessage>> future = mock(ListenableFuture.class);
```
elefeint, line 96: Don't need to mock it; use `ApiFutures.immediateFuture()`. It might help with unchecked warnings, too, since you can return a typed list.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
```diff
@@ -46,37 +64,131 @@
...
-				.withPropertyValues("management.health.pubsub.enabled=true")
+				.withBean("pubSubTemplate", PubSubTemplate.class, () -> mockPubSubTemplate)
+				.withPropertyValues(
+						"management.health.pubsub.enabled=true",
+						"spring.cloud.gcp.pubsub.health.subscription=test",
+						"spring.cloud.gcp.pubsub.health.timeout-millis=1500",
+						"spring.cloud.gcp.pubsub.health.acknowledgeMessages=true")
 				.run(ctx -> {
 					PubSubHealthIndicator healthIndicator = ctx.getBean(PubSubHealthIndicator.class);
 					assertThat(healthIndicator).isNotNull();
```
elefeint, line 110: Since this is testing context loading, might want to verify that the health indicator validation got invoked (that template.pullAsync() got called)

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
```diff
@@ -46,37 +64,131 @@
...
 
+	@SuppressWarnings("unchecked")
+	@Test
+	public void apiExceptionWhenValidating_userSubscriptionSpecified_healthAutoConfigurationFails() throws Exception {
+		PubSubHealthIndicatorProperties properties = new PubSubHealthIndicatorProperties();
+		PubSubHealthIndicatorAutoConfiguration p = new PubSubHealthIndicatorAutoConfiguration(properties);
+		properties.setSubscription("test");
+
+		PubSubTemplate mockPubSubTemplate = mock(PubSubTemplate.class);
+		ListenableFuture<List<AcknowledgeablePubsubMessage>> future = mock(ListenableFuture.class);
```
elefeint, line 156: `ApiFutures.immediateFailedFuture()` will help.

elefeint (changes requested):
My only substantial comment is about subscription names needing to start with a letter.

Otherwise lgtm with minor suggestions.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
```diff
@@ -16,55 +16,171 @@
...
+		}
 		catch (Exception e) {
-			builder.withException(e).down();
+			down.accept(e);
+		}
+	}
+
+	private void pullMessage() throws InterruptedException, ExecutionException, TimeoutException {
+		ListenableFuture<List<AcknowledgeablePubsubMessage>> future = pubSubTemplate.pullAsync(this.subscription, 1, true);
+		List<AcknowledgeablePubsubMessage> messages = future.get(timeoutMillis, TimeUnit.MILLISECONDS);
```
dzou, line 147: I think using the `future.get(..)` is blocking right? Should you register a callback here?

dzou (dismissed):

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
meltsufin: We decided to keep it synchronous.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
meltsufin: Since I actually want to verify calls on the future, I think I'll keep the mock.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
dzou: I see. Is there a benefit to using pullAsync vs. the Sync pull then?

elefeint (approved):

dzou (approved):

PR #331, opened 2021-02-22
Simply suggested BOM use in gradle

jasiekmiko:
The existing instructions seemed to refer to times before gradle supported BOMs. Since there is now a go-to way for BOM use built into gradle, it seems counter-intuitive to refer potential users to an out-dated solution.

### DIFF
--- a/docs/src/main/asciidoc/getting-started.adoc
+++ b/docs/src/main/asciidoc/getting-started.adoc
@@ -34,7 +34,7 @@ Instead, the version of the BOM you’re using determines the versions of the us
 </dependencyManagement>
 ----
 
-Or, if you're a gradle user:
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -34,13 +34,19 @@ Instead, the version of the BOM you’re using determines the versions of the us
 </dependencyManagement>
 ----
 
+Or, if you're a gradle user:
```
meltsufin, line 37:
```suggestion
Or, if you're a Gradle user:
```

meltsufin (changes requested):
Thank you!!

meltsufin (approved):

PR #332, opened 2021-02-22
Migrate warmCache functionality into daily/first runs of each job

ttomsu:
Thinking through this more, since the capability introduced in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/327 still requires artifacts to be downloaded for each particular job (unit test, integration test, docs, etc), and we already have most of these jobs running once per day - lets just use that run to create the **specific cache for that job**. AFAIK GHA doesn't limit you on the number of caches you can create, so we can just suffix the cache with the `job_id` (aka `github.job`). 

So for example, our ~5:35AM run of integration tests will create a cache named `Linux-maven-2021-02-22-integrationTests` that will be used by all merges that day - it should have all the dependencies needed for `compile`, `test`, `package`, `install`, and `verify`. Prior to this PR, it would spend some time downloading dependencies needed for `verify` because those weren't captured during the old `warmCache` job.


dzou (approved):
Cool!

PR #334, opened 2021-02-23
Add IT name to integration test cache

ttomsu:
even MOAR optimization


dzou (approved):
Sweeeeettttt more optimization :ok_hand:

meltsufin (commented):
Did you observe a significant difference in the caches for each of the IT modules?
One thing I noticed what that the cache loading was much slower for the first IT module and faster for the consequent ones. Would this make all of them as slow as the first?

PR #336, opened 2021-02-25
Add lazy `@Interleaved` example to Spanner sample

meltsufin:


spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/main/java/com/example/Trader.java
```diff
@@ -50,6 +51,9 @@
 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MMM-dd HH:mm:ss z")
 	private List<java.sql.Timestamp> modifiedOn;
 
+	@Interleaved(lazy = true)
+	private List<Trade> trades;
```
dzou, line 55: Should we leave a comment that it should be left uninitialized? Or should we just do that fix you suggested where we initialize it if the list is empty?

dzou (approved):

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/main/java/com/example/Trader.java
meltsufin: I haven't really looked into that possible fix. It's not common to initialize like that in entities. So, the comment might be distracting.

dzou: Hm yeah makes sense. Maybe making this mistake is fairly rare.

PR #340, opened 2021-02-26
Set up unified maven cache for GH Actions

meltsufin:
Moving away from a cache per job and integration test module.

This eliminates a lot of mostly overlapping maven caches, thereby reducing storage space and daily cache write bandwidth from ~30GB to ~1.5GB.

This also allows us to be within the [5GB cache storage limit](https://github.com/actions/cache#cache-limits).

### DIFF
--- a/.github/workflows/sonar.yaml
+++ b/.github/workflows/sonar.yaml
@@ -34,7 +34,7 @@ jobs:
         id: mvn-cache
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ steps.date.outputs.date }}
+          key: ${{ runner.os }}-maven-unified-${{ steps.date.outputs.date }}
       - name: Maven go offline
         id: mvn-offline
         if: steps.mvn-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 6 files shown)


.github/workflows/sonar.yaml
```diff
@@ -31,9 +31,14 @@ jobs:
           path: ~/.sonar/cache
           key: ${{ runner.os }}-sonar-${{ steps.date.outputs.date }}
       - uses: actions/cache@v2
+        id: mvn-cache
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ steps.date.outputs.date }}-unitTests
+          key: ${{ runner.os }}-maven-unified-${{ steps.date.outputs.date }}
+      - name: Maven go offline
+        id: mvn-offline
```
dzou, line 39: What does `go-offline` do again?

dzou (approved):
Looks fine -- So any consensus on which approach we ultimately prefer to settle on? (unified cache vs. module-based?)

.github/workflows/sonar.yaml
meltsufin: downloads all maven dependencies including plugin etc for your project

ttomsu (approved):

Issue #210, opened 2021-01-19
PathResourceResolver can't resolve a GoogleStorageResource due to no Google Storage UrlStreamHandler


jhordies:
When trying to use a Google Storage Resource as the spring static resource location the PathResourceResolver calls the method isResourceUnderLocation that calls the GoogleStorageResource.toURL() that throws a MalformedURLException because no URL stream handler is registered for the gs protocol (as stated by the documentation)

**Sample**
Add the following dependecy to your project:
```xml
		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-gcp-starter-storage</artifactId>
		</dependency>
```
Set the following property in you project:
spring.resources.static-locations=gs://[YOUR_GCS_BUCKET]/

If this is a public bucket (allUsers with Storage read permission) and no credentials are needed, use the following config class:
```java
@Configuration
public class BucketResourceConfig {

    @Bean
    public Storage storage() throws IOException {
        return StorageOptions.newBuilder()
                .setHeaderProvider(
                        new UserAgentHeaderProvider(GcpStorageAutoConfiguration.class))
                .build().getService();
    }

    @Bean
    public CredentialsProvider googleCredentials() throws Exception {
        return new NoCredentialsProvider();
    }
}
```
Assuming the root of the bucket contains a index.html
Call your server with http://host:port/index.html 
Put a breakpoint in PathResourceResolver.isResourceUnderLocation(...)
and see it fail to call resource.getURL() even though resource.getInputStream() would return the index.html

ttomsu:
Can you provide more details of your environment, specifically which version of Spring Cloud GCP, which Spring Cloud version, and which Spring Boot version you're using?

ttomsu:
Could you also point me to what documentation you're referring?

jhordies:
Hi,
> Could you also point me to what documentation you're referring?
I meant the javadoc of GoogleStorageResource
	/**
	 * Since the gs: protocol will normally not have a URL stream handler registered,
	 * this method will always throw a {@link java.net.MalformedURLException}.
	 * @return the URL for the GCS resource, if a URL stream handler is registered for the gs protocol.
	 */
	@Override
	public URL getURL() throws IOException {
		return getURI().toURL();
	}


> Can you provide more details of your environment, specifically which version of Spring Cloud GCP, which Spring Cloud version, and which Spring Boot version you're using?

[...]

PR #341, opened 2021-02-26
GoogleStorageResource: Returns the self-link of an object or bucket i…

ttomsu:
…nstead of always throwing a Malformed URL exception.

Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/210

This PR changes primarily changes `GoogleStorageResource.getURL()` from always throwing a `MalformedUrlException` to returning the object's `selfLink`, which is always a valid `https` URL. This URL just gives one back the object metadata, such as

https://www.googleapis.com/storage/v1/b/halconfig/o/versions.yml
```json
{
  "kind": "storage#object",
  "id": "halconfig/versions.yml/1614118188143610",
  "selfLink": "https://www.googleapis.com/storage/v1/b/halconfig/o/versions.yml",
  "mediaLink": "https://www.googleapis.com/download/storage/v1/b/halconfig/o/versions.yml?generation=1614118188143610&alt=media",
  "name": "versions.yml",
  "bucket": "halconfig",
  "generation": "1614118188143610",
  "metageneration": "1",
  "contentType": "application/text",
  "storageClass": "STANDARD",
  "size": "1536",
  "md5Hash": "VsQ39mqqIOaCINWZfiTs3w==",
  "crc32c": "SmfKkg==",
  "etag": "CPqn6+mCge8CEAE=",
  "timeCreated": "2021-02-23T22:09:48.221Z",
  "updated": "2021-02-23T22:09:48.221Z",
  "timeStorageClassUpdated": "2021-02-23T22:09:48.221Z"
}
```

But at least it's a valid URL. When this method always throwns a URL exception, we run into issues with other parts of the Spring ecosystem like the user's bug report. I tested out the user's config and it works as expected (serving index.html) when a bucket is specified. It doesn't look like it works if a bucket + prefix (subdirectory) is specified, but that may be a configuration issue on my end.

Secondarily, this PR cleans up several places where a checked `IOException` is in the method signature, but it appears the underlying library has started throwing a runtime `StorageException` instead, thus IntelliJ and Sonar were throwing up warnings to clean it up.

### DIFF
--- a/docs/src/main/asciidoc/storage.adoc
+++ b/docs/src/main/asciidoc/storage.adoc
@@ -90,7 +90,7 @@ However, creating signed URLs requires that the resource was created using servi
 ====
 As of v2.0.2+, the `GoogleStorageResource.getURL()` method returns the `Bucket` or `Blob` 's `selfLink` value, rather than attempting to convert the `URI` a `URL` object that nearly-always threw a `MalformedURLException`.
 This value is notably different from `GoogleStorageResource.getURI()`, which returns the more commonly used `gs://my-bucket/my-object` identifier.
-See https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/341[here] for more details.
+Returning a valid URL is necessary to support some features in the Spring ecosystem, such as `spring.resources.static-locations`.
 ====
 
 The Spring Boot Starter for Google Cloud Storage auto-configures the `Storage` bean required by the `spring-cloud-gcp-storage` module, based on the `CredentialsProvider` provided by the Spring Boot GCP starter.
--- a/docs/src/main/asciidoc/storage.adoc
+++ b/docs/src/main/asciidoc/storage.adoc
@@ -86,6 +86,13 @@ This type represents a GCS file, which has associated https://cloud.google.com/s
 The `createSignedUrl` method can also be used to obtain https://cloud.google.com/storage/docs/access-control/signed-urls[signed URLs] for GCS objects.
 However, creating signed URLs requires that the resource was created using service account credentials.
 
+[CAUTION]
+====
+As of v2.0.2+, the `GoogleStorageResource.getURL()` method returns the `Bucket` or `Blob` 's `selfLink` value, rather than attempting to convert the `URI` a `URL` object that nearly-always threw a `MalformedURLException`.
+This value is notably different from `GoogleStorageResource.getURI()`, which returns the more commonly used `gs://my-bucket/my-object` identifier.
+See https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/341[here] for more details.
+====
+
 The Spring Boot Starter for Google Cloud Storage auto-configures the `Storage` bean required by the `spring-cloud-gcp-storage` module, based on the `CredentialsProvider` provided by the Spring Boot GCP starter.
 
 ==== Setting the Content Type
--- a/spring-cloud-gcp-storage/src/main/java/com/google/cloud/spring/storage/GoogleStorageResource.java
+++ b/spring-cloud-gcp-storage/src/main/java/com/google/cloud/spring/storage/GoogleStorageResource.java
@@ -37,6 +37,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.core.io.WritableResource;
+import org.springframework.lang.NonNull;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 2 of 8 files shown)


dzou (approved):

meltsufin (approved):
I'll take a look at it in a bit. Please don't merge yet.

spring-cloud-gcp-storage/src/main/java/com/google/cloud/spring/storage/GoogleStorageResource.java
```diff
@@ -113,14 +114,13 @@ public boolean isAutoCreateFiles() {
...
+	 * @throws StorageException if an issue occurs getting the Bucket or Blob.
+	 * @return Returns true if the bucket or object exists.
+	 */
 	@Override
 	public boolean exists() {
-		try {
-			return isBucket() ? getBucket() != null : getBlob() != null;
-		}
-		catch (IOException ex) {
-			return false;
```
meltsufin, line 122: This might be a breaking change. I imagine that previously, if there was a permissions issue, it would return `false`, but now it will throw an exception.

spring-cloud-gcp-storage/src/main/java/com/google/cloud/spring/storage/GoogleStorageResource.java
```diff
@@ -134,27 +134,38 @@ public boolean isOpen() {
...
+			Assert.notNull(b, "The object at " + this.getURI() + " does not exist.");
+			url = getBlob().getSelfLink();
+		}
+		return new URL(url);
 	}
 
 	@Override
+	@NonNull
 	public URI getURI() {
 		return this.location.uri();
```
meltsufin, line 159: Hmmm, URI will be different than URL, which could be confusing. Maybe we should document this a bit.

meltsufin (changes requested):
Looks pretty good, but I'm just unsure about the breaking change in `exists()`.
Also, I think we need to document the behavior of `getURL()`.

spring-cloud-gcp-storage/src/main/java/com/google/cloud/spring/storage/GoogleStorageResource.java
ttomsu: Tested both with and without the change using a service account without permission, and both ended up throwing a `StorageException` that is not caught (because `StorageException` does not extend `IOException`), so _that_ behavior is the same.

Exception from the logs:
```
Request processing failed; nested exception is com.google.cloud.storage.StorageException: 897650840564-jriaec6tjb5n9od3fpgu63lgib25kqhk@developer.gserviceaccount.com does not have storage.objects.get access to the Google Cloud Storage object.
```

I'll add some docs.

spring-cloud-gcp-storage/src/main/java/com/google/cloud/spring/storage/GoogleStorageResource.java
ttomsu: From https://www.guru99.com/url-vs-uri-difference.html#:~:text=URI-,URL%20stands%20for%20Uniform%20Resource%20Locator.,Uniform%20Resource%20Name)%20or%20both.:

> Not all URIs are URLs since a URI can be a name instead of a locator.

In this case, it's URI doesn't conform to a URL syntax, so the self-link is the best we can do without adding some hacky reflection-based `UrlStreamHandlerFactory` impl.

meltsufin: Yeah I agree with the workaround. I just think it would be helpful to document it.

docs/src/main/asciidoc/storage.adoc
```diff
@@ -86,6 +86,13 @@ This type represents a GCS file, which has associated https://cloud.google.com/s
 The `createSignedUrl` method can also be used to obtain https://cloud.google.com/storage/docs/access-control/signed-urls[signed URLs] for GCS objects.
 However, creating signed URLs requires that the resource was created using service account credentials.
 
+[CAUTION]
+====
+As of v2.0.2+, the `GoogleStorageResource.getURL()` method returns the `Bucket` or `Blob` 's `selfLink` value, rather than attempting to convert the `URI` a `URL` object that nearly-always threw a `MalformedURLException`.
+This value is notably different from `GoogleStorageResource.getURI()`, which returns the more commonly used `gs://my-bucket/my-object` identifier.
+See https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/341[here] for more details.
```
meltsufin, line 93: I don't know if we should link to the PR in the refdoc. It seems too low-level.
Maybe just mention that we're doing it to allow for `spring.resources.static-locations`.

meltsufin (changes requested):

docs/src/main/asciidoc/storage.adoc
ttomsu: Done.

meltsufin (approved):

Issue #158, opened 2020-12-01
Add Support for Pub/Sub Dead Letter Queues


meltsufin:
Migrated from: https://github.com/spring-cloud/spring-cloud-gcp/issues/2328

Issue #85, opened 2020-10-14
Add support for Pub/Sub message ordering


meltsufin:
`PubSubAdmin.createSubscription()` will need to be expanded to allow enabling ordering, and whenever we create a `PubsubMessage`, we need to be able to set the ordering key for it.
Also, the Pub/Sub Cloud Stream Binder will need to be updated to support the option.
See: https://cloud.google.com/pubsub/docs/ordering
See: https://cloud.google.com/pubsub/docs/publisher#using_ordering_keys

Q: Does it makes sense to expose the keys as partitions in a Kafka-binder-like API?

meltsufin:
@ttomsu What do you think of focusing on this one after DLQ?

DerMorb:
The `PubSubMessageConverter.toPubSubMessage(Object payload, Map<String, String> headers)` method might also need a way to set the ordering key or we need another way to provide the ordering key.
Right now the `payload` is ending up in the `data` and the `headers` in the `attributes` of a `PubSubMessage` but the ordering key is a separate field.

Or do we need to write our own message converter somehow like this?

```
public class OrderMessagePubSubMessageConverter extends JacksonPubSubMessageConverter {

    public OrderMessagePubSubMessageConverter(ObjectMapper objectMapper) {
        super(objectMapper);
    }

    @Override
    public PubsubMessage toPubSubMessage(Object payload, Map<String, String> headers) {
        PubsubMessage pubsubMessage = super.toPubSubMessage(payload, headers);
        return PubsubMessage.newBuilder()
                            .mergeFrom(pubsubMessage)
                            .setOrderingKey(???)
                            .build();
    }
}
```

meltsufin:
@DerMorb I think we just need to pick a name for the header by adding it to `GcpPubSubHeaders` and then extend all implementations of `PubSubMessageConverter` to marshal that header back and forth between headers and `PubsubMessage.orderingKey`. It's kind of what you're suggesting, but more general.

meltsufin:
Now that #85 is in, what's left?

1.  In Spring Cloud Stream Binder for Pub/Sub, we should add the ability to automatically create subscriptions that support message ordering. This will involve adding `enableMessageOrdering` property to `PubSubConsumerProperties` and updating `PubSubChannelProvisioner.createSubscription`.
2. In Spring Integration Outbound Channel Adapter AKA `PubSubMessageHandler` we can add a property to specify a default ordering key.
[...]

PR #343, opened 2021-03-01
Add createSubscription method with single Subscription.Builder argume…

ttomsu:
…nt as a catch-all.

This was originally intended to help solve https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/158 (DLQ support), but may also enable basic support of message ordering (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/85), replay, and filtering.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
@@ -235,8 +235,8 @@ public class PubSubAdmin implements AutoCloseable {
 		Assert.hasText(topicName, NO_TOPIC_SPECIFIED);
 
 		Subscription.Builder builder = Subscription.newBuilder()
-				.setName(PubSubSubscriptionUtils.toProjectSubscriptionName(subscriptionName, this.projectId).toString())
-				.setTopic(PubSubTopicUtils.toTopicName(topicName, this.projectId).toString());
+				.setName(subscriptionName)
+				.setTopic(topicName);
 
 		builder.setAckDeadlineSeconds(this.defaultAckDeadline);
 		if (ackDeadline != null) {
@@ -260,6 +260,13 @@ public class PubSubAdmin implements AutoCloseable {
 	 * @return the created subscription
 	 */
 	public Subscription createSubscription(Subscription.Builder builder) {
+		Assert.notNull(builder, "Builder cannot be null");
+		Assert.hasText(builder.getName(), "Subscription name must not be null or empty");
+		Assert.hasText(builder.getTopic(), "Topic name must not be null or empty");
+
+		builder.setName(PubSubSubscriptionUtils.toProjectSubscriptionName(builder.getName(), this.projectId).toString());
+		builder.setTopic(PubSubTopicUtils.toTopicName(builder.getTopic(), this.projectId).toString());
+
 		return this.subscriptionAdminClient.createSubscription(builder.build());
 	}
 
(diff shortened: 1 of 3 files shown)


dzou (approved):

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
```diff
@@ -235,22 +234,33 @@ public Subscription createSubscription(String subscriptionName, String topicName
 		Assert.hasText(subscriptionName, "No subscription name was specified.");
 		Assert.hasText(topicName, NO_TOPIC_SPECIFIED);
 
-		int finalAckDeadline = this.defaultAckDeadline;
+		Subscription.Builder builder = Subscription.newBuilder()
+				.setName(PubSubSubscriptionUtils.toProjectSubscriptionName(subscriptionName, this.projectId).toString())
+				.setTopic(PubSubTopicUtils.toTopicName(topicName, this.projectId).toString());
+
+		builder.setAckDeadlineSeconds(this.defaultAckDeadline);
```
meltsufin, line 241: Hmmm, previously we would only set ack deadline if the user specified a custom one. Was that a bug?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
```diff
@@ -235,22 +234,33 @@ public Subscription createSubscription(String subscriptionName, String topicName
...
-				finalAckDeadline);
+	/**
+	 * Create a new subscription on Google Cloud Pub/Sub.
+	 *
+	 * @param builder a Subscription.Builder straight from the client API library that exposes all available knobs and
+	 * levers
+	 * @return the created subscription
+	 */
+	public Subscription createSubscription(Subscription.Builder builder) {
+		return this.subscriptionAdminClient.createSubscription(builder.build());
```
meltsufin, line 263: Should we set the project on the subscription, if it wasn't set?

meltsufin (changes requested):

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
ttomsu: It's not true that we previously didn't set the ack deadline - see the actual `createSubscription` method call in the previous code - we were always passing the default or the user override.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
ttomsu: I would say no - I think the point of this method should be straight passthrough. It's main value add is hooking into the credential, not to change the user's inputs.

ttomsu: After now working with this, yes, we probably should look for the `projects/` prefix and put one on if missing

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
meltsufin: You're right!

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/PubSubAdmin.java
meltsufin: They require a fully qualified subscription name, right?
If we want to allow "relative" subscription names and topic names, we would need to look for the absense of the `projects/` prefix and add in the configured project to it.

meltsufin (approved):

Issue #342, opened 2021-03-01
Propagate `storeOnDisk` property to the configuration


artemptushkin:
### Problem.

Some Datastore emulator runs require `--no-store-on-disk` start-up property, but currently, there is no way to pass it

### Solution

I would suggest to create an additional property for `org.springframework.cloud.gcp.autoconfigure.datastore.EmulatorSettings` and pass it to `LocalDatastoreHelper`

meltsufin:
Sounds like a good idea.

artemptushkin:
I will take this

PR #344, opened 2021-03-02
GH-342: Apply additional emulator properties

artemptushkin:
#342 Currently, some properties are missing to propagate from the application properties to Datastore emulator.

I propagate `storeOnDisk` and `dataDir` here and adjust the javadoc. The task supposes to add only the first property but consistency order, I add the second one.

Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/342.


dzou:
Nice contribution! Looks good. I pushed a commit to fix the checkstyle.

I'll wait a little to let other team members take a look and then merge it in.

dzou (approved):

ttomsu (approved):

meltsufin (commented):
Can we also update the `datastore.adoc` with the new properties? Thanks!

dzou:
Yup I can do that.

PR #348, opened 2021-03-03
Add sample demonstrating the difference between PubSub dead letter queue and Spring Integration error channel

ttomsu:
Refdoc update coming (including a fancy picture), but getting the sample out for some feedback first.


meltsufin (commented):
What do you think of focusing on extending the reference documentation? We already have a section on [error handling](https://googlecloudplatform.github.io/spring-cloud-gcp/2.0.1/reference/html/index.html#error-handling). I feel like we're just missing a new section on the use of dead-letter queues. You can actually move a lot of this code into an integration test and pull code snippets from it into the refdoc.

If we were to add the sample, I think it might be more clear if it was just focused on one approach -- the dead letter queue. Maybe the error-channel approach can be added to the existing integration-pubsub-sample. In any case, I would be satisfied even if just the refdoc was updated. The sample is a bonus.

elefeint (commented):
I agree that the DLQ approach is the most useful part to demonstrate to the users. We should definitely keep this sample, though -- it shows a part of Pub/Sub capabilities that's hard to mentally tie in with the Spring Integration ecosystem. This sample addresses the feedback that we saw in the readout today -- that the documentation samples are too basic and don't teach you real-world skills.

For `SpringIntegrationErrorChannelApproach`, @ttomsu did you mean to add an application-managed simulated DLQ? In principle, we demonstrate something similar (although without error thrown by error handler) in the Spring Cloud Stream binder sample here -- https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/2c78357734bd1bf9914e9e88ab486a7d0c0bcf84/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java#L41

ttomsu:
@elefeint I was attempting to demonstrate that if there is an error channel configured and it completes successfully, the message will be ack'ed, and if the _error channel_ also throws an exception, only _then_ it will be nack'ed. I added this because it wasn't clear to me until you explained it why messages weren't being redelivered when funneled to the error channel.

ttomsu:
Closing in favor of https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/358

PR #350, opened 2021-03-04
Adds section in pubsub docs on using the Subscription.Builder to set …

ttomsu:
…up dead letter topics.

I cleaned up a bit of stuff that I think makes some sections a bit cleaner/clearer to understand.

Personally, as a side note, I don't see that much value in most of the sections under `Pub/Sub management` that wouldn't be better served by pointing to the Javadoc. "PubSubAdmin implements a method to [CRUD operation]" IMO doesn't really need to be explained.

### DIFF
--- a/docs/src/main/asciidoc/pubsub.adoc
+++ b/docs/src/main/asciidoc/pubsub.adoc
@@ -234,7 +234,7 @@ NOTE: All `ack()`, `nack()`, and `modifyAckDeadline()` methods on messages, as w
 ===== Dead Letter Topics
 
 Your application may occasionally receive a message it cannot process.
-If you <<creating-a-subscription,create your `Subscription`>> using the `Subscription.Builder` option, you can specify a `DeadLetterPolicy` that will forward all `nack()`-ed and non-`ack()`-ed message after a configurable amount of redelivery attempts.
+If you <<creating-a-subscription,create your `Subscription`>> passing the `Subscription.Builder` argument, you can specify a `DeadLetterPolicy` that will forward all `nack()`-ed and non-`ack()`-ed messages after a configurable amount of redelivery attempts.
 See https://cloud.google.com/pubsub/docs/dead-letter-topics[here] for more information.
 
 // TODO(ttomsu): make this an include to working code.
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -231,6 +231,32 @@ Using these methods for acknowledging messages in batches is more efficient than
 
 NOTE: All `ack()`, `nack()`, and `modifyAckDeadline()` methods on messages, as well as `PubSubSubscriberTemplate`, are implemented asynchronously, returning a `ListenableFuture<Void>` to enable asynchronous processing.
 
+===== Dead Letter Topics
+
+Your application may occasionally receive a message it cannot process.
+If you <<creating-a-subscription,create your `Subscription`>> using the `Subscription.Builder` option, you can specify a `DeadLetterPolicy` that will forward all `nack()`-ed and non-`ack()`-ed message after a configurable amount of redelivery attempts.
```
elefeint, line 237:
```suggestion
If you <<creating-a-subscription,create your `Subscription`>> passing the `Subscription.Builder` argument, you can specify a `DeadLetterPolicy` that will forward all `nack()`-ed and non-`ack()`-ed messages after a configurable amount of redelivery attempts.
```

meltsufin (approved):
Looks good after you address that `TODO`.
What we've learned from UXR is that sample code is typically far better than these paragraphs with references to classes and methods.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -371,46 +395,30 @@ include::{project-root}/spring-cloud-gcp-autoconfigure/src/test/java/com/google/
...
-}
+public Subscription createSubscription(String subscriptionName, String topicName, Integer ackDeadline, String pushEndpoint)
+
+public Subscription createSubscription(Subscriber.Builder builder)
 ----
 
-Alternative methods with default settings are provided for ease of use.
 The default value for `ackDeadline` is 10 seconds.
 If `pushEndpoint` isn’t specified, the subscription uses message pulling, instead.
+You can also pass a full `Subscription.Builder` for full control over any options or features available in the client library.
```
elefeint, line 413:
```suggestion
You can also pass a `Subscription.Builder` for full control over any options or features available in the client library.
```

elefeint (approved):

ttomsu:
Thanks for those edits - writing is hard :pen:

Issue #309, opened 2021-02-15
deprecate pubsub PullRequest returnImmediately flag 


lukasGemela:
**Is your feature request related to a problem? Please describe.**
whilst investigating problems with delayed  pubsub messages we found out that our problems were caused by returnImmediately flag set to true. According to documentation this flag is actually deprecated server-side and they heavily discourage from using it now:
https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#pullrequest 

**Describe the solution you'd like**
Either remove it or deprecate it using annotation.

**Describe alternatives you've considered**
No alternatives.

**Additional context**
We could also do the PR but I open this report to run decision whether keep the flag in deprecated state or remove it completely.

dzou:
Thanks for the report. We had discussions of deprecating this in the past, though had trouble coordinating this change with the Pub/Sub team. Will follow-up and ask what is our long-term plan for this. @meltsufin @elefeint

meltsufin:
Related to: https://github.com/spring-cloud/spring-cloud-gcp/issues/2247.
If I recall correctly, we decided to keep it last time we looked at it because there was no directly replacement for it.

lukasGemela:
I think my biggest issue is that setting returnImmediately to true has some serious side effect for us (message delivery delayed for hours), the flag itself is deprecated and there is no mentioning in documentation or javaDoc. Spring template documentation should at least discourage people from using this flag

meltsufin:
> I think my biggest issue is that setting returnImmediately to true has some serious side effect for us (message delivery delayed for hours)

This sounds like a bug potentially. Can you elaborate how the use of this flag delays message delivery?

lukasGemela:
> > I think my biggest issue is that setting returnImmediately to true has some serious side effect for us (message delivery delayed for hours)
> 
> This sounds like a bug potentially. Can you elaborate how the use of this flag delays message delivery?

sure, please read through this ticket 
https://issuetracker.google.com/issues/179915783

meltsufin:
@lukasGemela Thanks for the reference to the ticket. We'll work internally to figure out how to proceed and update this issue soon.

meltsufin:
@lukasGemela The word from the Pub/Sub team is that the way `returnImmediately=true` works at the moment is that it waits up to 1 second before returning the pull call, and it's only kind of deprecated because there are still users who depend on its behavior and there is no direct replacement yet.
[...]

PR #354, opened 2021-03-05
Add warning about use of `returnImmediately=true`

meltsufin:
Updates javadocs and reference documentation with a warning.
However, we decided not to deprecate the field for now.

Addresses: #309.

### DIFF
--- a/docs/src/main/asciidoc/pubsub.adoc
+++ b/docs/src/main/asciidoc/pubsub.adoc
@@ -219,9 +219,9 @@ h|pullNext | Allows for a single message to be pulled and automatically acknowle
 h|pullAndConvert | Works the same as the `pull` method and, additionally, converts the Pub/Sub binary payload to an object of the desired type, using the converter configured in the template.
 |===
 
-WARNING: Setting `returnImmediately` to `true` is not recommended as it may result in long delays in message delivery.
-This is due to the fact that "immediately" really means 1 second, and if no messages can be retrieved in that time from the Pub/Sub backend, no messages will be returned.
-This may happen repeatedly despite having messages queued up on the topic. The recommended alternative is to set `returnImmediately` to `false`, or to use `subscribe` methods described in the previous section.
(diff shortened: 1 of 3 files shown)


docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -219,6 +219,10 @@ h|pullNext | Allows for a single message to be pulled and automatically acknowle
 h|pullAndConvert | Works the same as the `pull` method and, additionally, converts the Pub/Sub binary payload to an object of the desired type, using the converter configured in the template.
 |===
 
+WARNING: Setting `returnImmediately` to `true` is not recommended as it may result in long delays in message delivery.
```
ttomsu, line 222: I'd prefer more active voice:

- We recommend setting `returnImmediately` to `true`, otherwise it may result in delayed message delivery.

- The implementation really means 1 second, and if the client cannot retrieve any messages in that time from the Pub/Sub backend, it will not retrieve any message at all (may need to clarify this sentence?)

 - We recommend setting `returningImmediately` to false, or use one of the `subscribe` methods from the previous section.

meltsufin: Thanks! How about now?

ttomsu: Much better, though you may want to say "if the Pub/Sub _client_ cannot  get messages from the backend ... " because to me, Pub/Sub is the _service_ so it doesn't make sense that the service can't get messages from itself.

ttomsu (approved):

elefeint (approved):

docs/src/main/asciidoc/pubsub.adoc
meltsufin: Based on the conversation with the Pub/Sub team, it seems more like the latter because of the distributed nature of Pub/Sub. So, omission of the word "client" was intentional.

PR #355, opened 2021-03-05
Fix Pub/Sub Integration Snippet in doc

dzou:
I changed it to inject `PubSubTemplate` instead of `PubSubSubscriberOperations`; turns out there are 2 beans in the environment that implement `PubSubSubscriberOperations` making it non-unique -- `PubSubTemplate` and `PubSubSubscriberTemplate`.


meltsufin (approved):
You can technically just add a bean qualifier, but this is simpler.

PR #358, opened 2021-03-10
Cloud Stream Dead Letter Topic support

ttomsu:
Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/158
Two commits here - one for the production code and one for the sample.

Since Cloud Stream handles all things related to topic/subscription creation, we need a way to support dead letter topics. This PR expands the `PubSubConsumerProperties` to allow setting `deadLetterPolicy.deadLetterTopic` and `deadLetterPolicy.maxDeliveryAttempts`

One of the primary changes is consolidating the `createSubscription` call into its own function which applies the above policy (if available).  Because we're potentially creating _another_ topic, I removed the `Optional` return type from `ensureTopicExists` and instead have it call itself recursively to ensure we nearly always get a `Topic` object back and throw an exception in the extremely rare case where we first get an `AlreadyExistsException` but then still get `null` back from `getTopic()`. This is a slight change in behavior because were previously assuming everything was fine and just logging a warning, which we're no longer doing - like I said, though, this should be a pretty rare case.

The tests were refactored to remove the JUnit-4-only `@Rule` for exception catching, and use AssertJ instead.

---

The sample code is a combination of the simplicity of the legacy (annotation-based) style and the current functional style, but without the module split out of the other functional sample.

It ends up with 3 functions - one to publish the message, one to receive the message from the main topic, and one to receive the message from the dead letter topic. An integration test ensures this, like all the other tests, starts and runs the application passes as expected.

This PR is already pretty big, so I'll follow up with another PR to pull in some code samples into the refdoc to fix a TODO I put in there earlier.


spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/properties/PubSubConsumerProperties.java
```diff
@@ -57,4 +59,34 @@ public String getSubscriptionName() {
 	public void setSubscriptionName(String subscriptionName) {
 		this.subscriptionName = subscriptionName;
 	}
+
+	public DeadLetterPolicy getDeadLetterPolicy() {
+		return deadLetterPolicy;
+	}
+
+	public void setDeadLetterPolicy(DeadLetterPolicy deadLetterPolicy) {
```
dzou, line 67: Would it be better to use the [`DeadLetterPolicy` proto object in the client libs](https://github.com/googleapis/java-pubsub/blob/7d5bc463cf62395137b88261e01ceb792d379001/samples/snippets/src/main/java/pubsub/UpdateDeadLetterPolicyExample.java#L61) instead of creating a new custom wrapper?

The proto object is a little more complex, but it would reduce code duplication and also stay up to date if new settings are introduced to dead letter policy.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -124,23 +128,47 @@ public void afterUnbindConsumer(ConsumerDestination destination) {
 		}
 	}
 
-	private Optional<Topic> ensureTopicExists(String topicName, boolean autoCreate) {
+	Topic ensureTopicExists(String topicName, boolean autoCreate) {
```
dzou, line 131: I like the recursion, clever :+1:

Minor refactor suggestion: I would suggest first checking if `topic != null` and then returning early instead of doing the `topic == null` check. To me it looks a little bit simpler.

```
Topic topic = this.pubSubAdmin.getTopic(topicName);

if (topic != null) {
  return topic;
}

if (autoCreate) {
	try {
		return this.pubSubAdmin.createTopic(topicName);
	}
	catch (AlreadyExistsException alreadyExistsException) {
		return ensureTopicExists(topicName, false);
	}
}
else {
	throw new ProvisioningException("Non-existing '" + topicName + "' topic.");
}

```

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-deadLetterTopic-sample/src/main/java/com/example/SinkExample.java
```diff
@@ -0,0 +1,62 @@
...
+ * Example of a sink for the sample app.
+ *
+ * @author Travis Tomsu
+ */
+@Configuration
+@Slf4j
+public class SinkExample {
+
+	@Bean
+	public Consumer<Message<UserMessage>> logUserMessage() {
```
dzou, line 41: Should this also be a `Consumer<UserMessage>` like the below `deadLetterMessages()` for consistency? Or does one way only take the wrapped message object?

ttomsu: Unfortunately no, because I need the ability to `nack()` the message. I've added a utility method that I'd like to get your opinion on for more reliably getting an object I can nack - though I'm not quite sure how this interacts with the acking/nacking happening within the framework. Maybe @elefeint would have some insight.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/properties/PubSubConsumerProperties.java
dzou: :+1:

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-deadLetterTopic-sample/src/main/java/com/example/SourceExample.java
```diff
@@ -0,0 +1,68 @@
...
+	@Configuration
+	public static class SourceConfig {
+
+		@Bean
+		public Sinks.Many<UserMessage> postOffice() {
+			return Sinks.many().unicast().onBackpressureBuffer();
+		}
+
+		@Bean
+		Supplier<Flux<UserMessage>> generateUserMessages(Sinks.Many<UserMessage> postOffice) {
```
meltsufin, line 64: Can we do the same without introducing reactive? Would it be simpler?

meltsufin (changes requested):
Very cool! Just some minor comments.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -73,7 +74,10 @@ public ConsumerDestination provisionConsumerDestination(String topicName, String
 
 		// topicName may be either the short or fully-qualified version.
 		String topicShortName = TopicName.isParsableFrom(topicName) ? TopicName.parse(topicName).getTopic() : topicName;
-		Optional<Topic> topic = ensureTopicExists(topicName, properties.getExtension().isAutoCreateResources());
+		Topic topic = ensureTopicExists(topicName, properties.getExtension().isAutoCreateResources());
+
+		PubSubConsumerProperties.DeadLetterPolicy deadLetterPolicy = properties.getExtension().getDeadLetterPolicy();
+		boolean autoCreate = properties.getExtension().isAutoCreateResources();
```
elefeint, line 80: Might want to move it to above line 77, and just use that variable. 
More broadly, perhaps none of this logic needs to run if `autoCreateResources` is off?

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -124,23 +128,45 @@ public void afterUnbindConsumer(ConsumerDestination destination) {
...
+		}
+
+		if (autoCreate) {
+			try {
+				return this.pubSubAdmin.createTopic(topicName);
 			}
-			else {
-				throw new ProvisioningException("Non-existing '" + topicName + "' topic.");
+			catch (AlreadyExistsException alreadyExistsException) {
+				return ensureTopicExists(topicName, false);
```
elefeint, line 142: Maybe add a comment that this addresses a race condition between two microservices trying to autocreate?

elefeint (approved):

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-dead-letter-sample/README.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+4. Run the `mvn clean spring-boot:run` in the root of the code sample to get the app running.
+
+5. Browse `localhost:8080`, type in a message, a user name, and press the `Post it!` button.
+
+6. Verify in your app's logs that a similar message was posted:
++
+```
+Publishing message from batman
+Nacking message 1 from batman at 2021-03-10T15:33:00.479420: to the batcave!
```
meltsufin, line 31: It's going to say `(attempt 1)` now.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-dead-letter-sample/src/main/resources/application.properties
```diff
@@ -0,0 +1,23 @@
...
+spring.cloud.stream.bindings.generateUserMessages-out-0.destination=my-main-topic
+
+# The applications' Sink receives messages throug the logUserMessage Consumer. The anonymous subscriber is configured to
+# send any unprocessable messages to the dead letter topic after a set number of retry attempts.
+spring.cloud.stream.bindings.logUserMessage-in-0.destination=my-main-topic
+spring.cloud.stream.gcp.pubsub.bindings.logUserMessage-in-0.consumer.deadLetterPolicy.deadLetterTopic=my-dead-letter-topic
+spring.cloud.stream.gcp.pubsub.bindings.logUserMessage-in-0.consumer.deadLetterPolicy.maxDeliveryAttempts=6
+# We set this to MANUAL ackMode to take control over the ack()/nack()-ing of messages for this demo.
+# You do not need to do this in a normal application.
+spring.cloud.stream.gcp.pubsub.bindings.logUserMessage-in-0.consumer.ackMode=MANUAL
```
meltsufin, line 14: I'm surprised it worked without this setting for you.

meltsufin (approved):

dzou (approved):

ttomsu:
> I'm surprised it worked without this setting for you.

it worked because I `nack`ed the message before the framework could `ack` it. I'm not sure if Pub/Sub threw any errors for attempting to ack and already nacked message, but it apparently redelivered it all the same.

PR #375, opened 2021-03-12
Skip integration tests for dependabot

meltsufin:
This is because secrets are not available for dependabot.
See: https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/

### DIFF
--- a/.github/workflows/integrationTests.yaml
+++ b/.github/workflows/integrationTests.yaml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   integrationTests:
-    if: github.repository == 'GoogleCloudPlatform/spring-cloud-gcp'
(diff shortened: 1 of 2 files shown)


ttomsu (commented):
Sonar doesn't need to be changed right? It doesn't use any GH Secret?

meltsufin:
> Sonar doesn't need to be changed right? It doesn't use any GH Secret?

It seems to be using `secrets.GITHUB_TOKEN`, but maybe that's not needed? Not sure.

ttomsu:
It may be used to post the comment with the result? Otherwise I don't know
what it would be used for.

meltsufin:
@ttomsu I looked at the previous runs with dependabot and it seems fine. So, removing that change.

elefeint (approved):

PR #376, opened 2021-03-12
Make ITs wait their turn across all branches

meltsufin:
By default, turntstyle will only block concurrent jobs on the same branch.
That's not what we want.


ttomsu (approved):
This was what I was missing!! I always thought it wasn't doing exactly what I thought but there were other times I did see it pause and wait and was confused why.

PR #390, opened 2021-03-16
Fix SpannerRepositoryInsertIntegrationTests by restoring #92

meltsufin:
Restores change from #92.


elefeint (approved):

elefeint (approved):
Want to leave a comment why we need it, so it does not get removed in the future?

meltsufin:
> Want to leave a comment why we need it, so it does not get removed in the future?

Good idea! Done.

PR #392, opened 2021-03-16
Update unitTests.yaml

ttomsu:


.github/workflows/unitTests.yaml
```diff
@@ -64,8 +64,6 @@ jobs:
           install
 
     - name: Unit Tests
-      id: unitTest1
-      continue-on-error: true
```
meltsufin, line 68: What about other places where you have `continue-on-error: true`?

ttomsu: Those should be OK - this was accidentally left over from when I was retrying unit tests on failure.

meltsufin: ah ok

meltsufin (approved):

PR #393, opened 2021-03-17
Add conditionals on PubSubBinderConfig, including a property to expli…

ttomsu:
…citly disable it.

Originally filed issue: https://github.com/spring-cloud/spring-cloud-gcp/issues/2653

Thing to discuss is the `testPubSubDisabled_noReplacements` test, which as written will not fail on startup even if that particular configuration doesn't make a whole lot of sense.

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/config/PubSubBinderConfigurationTests.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/config/PubSubBinderConfigurationTests.java
@@ -28,6 +28,7 @@ import org.junit.Test;
 import org.springframework.boot.autoconfigure.AutoConfigurations;
 import org.springframework.boot.test.context.runner.ApplicationContextRunner;
 import org.springframework.context.annotation.Bean;
+import org.springframework.test.annotation.DirtiesContext;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
@@ -35,6 +36,7 @@ import static org.mockito.Mockito.mock;
 /**
  * Tests for PubSubBinderConfiguration and its interaction with other autoconfiguration classes.
  */
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
 public class PubSubBinderConfigurationTests {
 
 	@Test
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/config/PubSubBinderConfigurationTests.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/config/PubSubBinderConfigurationTests.java
@@ -20,6 +20,7 @@ import com.google.api.gax.core.CredentialsProvider;
 import com.google.auth.Credentials;
 import com.google.cloud.spring.autoconfigure.core.GcpContextAutoConfiguration;
 import com.google.cloud.spring.autoconfigure.pubsub.GcpPubSubAutoConfiguration;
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/config/PubSubBinderConfigurationTests.java
```diff
@@ -0,0 +1,131 @@
...
+		ApplicationContextRunner baseContext = new ApplicationContextRunner()
+				.withPropertyValues(
+						"spring.cloud.gcp.pubsub.enabled=false",
+						"spring.cloud.gcp.pubsub.binder.enabled=true")
+				.withConfiguration(AutoConfigurations.of(
+						GcpContextAutoConfiguration.class,
+						GcpPubSubAutoConfiguration.class,
+						PubSubBinderConfiguration.class))
+				.withUserConfiguration(TestConfiguration.class);
+		baseContext.run(ctx -> assertThat(ctx).doesNotHaveBean(PubSubChannelProvisioner.class));
```
elefeint, line 93: Are there any warnings logged from Spring Cloud Stream perhaps?

elefeint (changes requested):
Could you add a refdoc mention of the new gating property?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/config/PubSubBinderConfigurationTests.java
dzou, line 95: For this test, is it possible to try get the Binder bean in the `baseContext.run(ctx -> ...)` from the context and then checking if an exception is thrown? Maybe this is what you want to check. I think if you never try retrieving/injecting a bean that misses dependencies in the app, then you will never run into the error since it won't try to create the bean.

But I think in practice if you can get the exception to throw when the bean is retrieved/injected somewhere in the app, then this is good enough; it is how it works in other dependency injection systems too like Guice I think.

ttomsu: AFAIK that's not the case w Spring - beans are created even if never injected.

dzou: Ah I see.

dzou: Maybe I am missing something, I'm not quite following why the test doesn't throw an exception then. Doesn't `PubSubBinderConfiguration` create the `PubSubChannelProvisioner` bean which should throw an error if `pubsub.enabled=false`?

meltsufin (approved):
Very thorough tests! :)

We just need to add the property to `additional-spring-configuration-metadata.json` and to the refdoc.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/config/PubSubBinderConfigurationTests.java
ttomsu:
```
A component required a bean of type 'org.springframework.cloud.stream.binder.Binder' that could not be found.
```

Obviously if there are multiple binders this wouldn't happen, but at least there is some check that if this is the only one there has been a misconfiguration somewhere.

dzou (approved):

elefeint (approved):

ttomsu:
Property added to javadoc and `additional-spring-configuration-metadata.json`

PR #400, opened 2021-03-18
(testing pr) experimenting with changes to fix apiary client auth

dzou:
Still not sure if it is a change needed on our end or in the auth libraries.

Some discussion of this here: https://github.com/googleapis/google-auth-library-java/pull/601#pullrequestreview-615870841

But maybe we need to make a change on our end.


dzou:
I printed out the auth scopes to verify that they were being attached to the credential: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/runs/2143797505

I think we have it correct on our end at least. Curious to see if the new version will fix the problem for us.

arithmetic1728:
> I printed out the auth scopes to verify that they were being attached to the credential: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/runs/2143797505
> 
> I think we have it correct on our end at least. Curious to see if the new version will fix the problem for us.

In auth lib v0.25.0 we ignored the scopes and it was corrected in the fix. Sorry for the trouble.

It should go back to normal after libraries-bom PR is merged or released: https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1992

meltsufin:
Just upgraded. So, closing this.

PR #408, opened 2021-03-24
Enable Pub/Sub message ordering for publishing

meltsufin:
Introduces `GcpPubSubHeaders.ORDERING_KEY` header that will be automatically applied to the `PubsubMessage` up conversion.
Also, introduces `spring.cloud.gcp.pubsub.publisher.enable-message-ordering` property to enable message ordering on the `Publisher`.

First step in addressing: #85.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubProperties.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubProperties.java
@@ -126,6 +126,11 @@ public class GcpPubSubProperties implements CredentialsSupplier {
 		 */
 		private final Batching batching = new Batching();
 
+		/**
+		 * Enable message ordering setting.
+		 */
+		private Boolean enableMessageOrdering;
+
 		public Batching getBatching() {
 			return this.batching;
 		}
@@ -140,6 +145,14 @@ public class GcpPubSubProperties implements CredentialsSupplier {
 
 		public void setExecutorThreads(int executorThreads) {
 			this.executorThreads = executorThreads;
+		}
+
+		public Boolean getEnableMessageOrdering() {
+			return enableMessageOrdering;
+		}
+
+		public void setEnableMessageOrdering(Boolean enableMessageOrdering) {
+			this.enableMessageOrdering = enableMessageOrdering;
 		}
 	}
 
--- a/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubTemplateDocumentationIntegrationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubTemplateDocumentationIntegrationTests.java
@@ -34,6 +34,7 @@ import com.google.cloud.spring.pubsub.PubSubAdmin;
 import com.google.cloud.spring.pubsub.core.PubSubTemplate;
 import com.google.cloud.spring.pubsub.integration.outbound.PubSubMessageHandler;
 import com.google.cloud.spring.pubsub.support.AcknowledgeablePubsubMessage;
+import com.google.cloud.spring.pubsub.support.GcpPubSubHeaders;
 import com.google.cloud.spring.pubsub.support.converter.ConvertedAcknowledgeablePubsubMessage;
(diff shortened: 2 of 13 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubTemplateDocumentationIntegrationTests.java
```diff
@@ -93,6 +95,35 @@ public void testCreatePublishPullNextAndDelete() {
 		});
 	}
 
+	@Test
+	public void testCreatePublishPullNextAndDelete_ordering() {
+		pubSubTest((AssertableApplicationContext context, PubSubTemplate pubSubTemplate, String subscriptionName, String topicName) -> {
+			//tag::publish_ordering[]
+			Map<String, String> headers = Collections.singletonMap(GcpPubSubHeaders.ORDERING_KEY, "key1");
+			pubSubTemplate.publish(topicName, "message1", headers).get();
+			pubSubTemplate.publish(topicName, "message2", headers).get();
```
elefeint, line 104: Maybe add a comment to explain that messages with the same ordering key will be delivered in the same order they were sent.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -126,6 +126,11 @@ public void setKeepAliveIntervalMinutes(int keepAliveIntervalMinutes) {
 		 */
 		private final Batching batching = new Batching();
 
+		/**
+		 * Enable message ordering setting.
+		 */
+		private Boolean enableMessageOrdering;
```
elefeint, line 132: why not a primitive boolean? Default state is false; would be no need to mess with nulls.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultPublisherFactory.java
```diff
@@ -163,6 +151,36 @@ public Publisher createPublisher(String topic) {
...
+
+		if (this.retrySettings != null) {
+			publisherBuilder.setRetrySettings(this.retrySettings);
+		}
+
+		if (this.batchingSettings != null) {
+			publisherBuilder.setBatchingSettings(this.batchingSettings);
+		}
+
+		if (this.enableMessageOrdering != null) {
```
elefeint, line 179: If it's primitive, then we don't need to check for null here, just straight set the value, or, to be extra safe, check if true.

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/integration/outbound/PubSubMessageHandlerTests.java
```diff
@@ -227,4 +229,18 @@ public void testSetHeaderMapperWithNull() {
...
+
+	@Test
+	public void testPublishWithOrderingKey() {
+		this.message = new GenericMessage<byte[]>("testPayload".getBytes(),
+				new MapBuilder<String, Object>()
+						.put(GcpPubSubHeaders.ORDERING_KEY, "key1")
+						.build());
+
+		this.adapter.handleMessage(this.message);
+		verify(this.pubSubTemplate, times(1))
```
elefeint, line 241: can omit `times(1)`

elefeint (approved):
Looks good, couple of minor things.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubProperties.java
dzou: +1 yeah would also prefer the primitive boolean to avoid implying that not setting it is different than setting it to `false`.

dzou (approved):

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubProperties.java
meltsufin: Sometimes we follow the pattern that is to allow not setting it, so that the client library default can be used automatically. Otherwise, we have to maintain the default, which in this case probably doesn't matter.

Issue #88, opened 2020-10-15
Investigate Support for Pub/Sub Lite


meltsufin:
[Pub/Sub Lite](https://cloud.google.com/pubsub/lite/docs?hl=en) is a new product that is a partitioned, append-only log.

breun:
Google provides a basic Java library for Pub/Sub Lite: https://cloud.google.com/pubsub/lite/docs/reference/libraries#client-libraries-install-java Spring Cloud GCP could auto-configure some beans based on that?

PR #427, opened 2021-04-01
Modifications required to enable Pub/Sub Lite basic usage.

dpcollins-google:
fixes: #88

### DIFF
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -71,6 +71,14 @@
             <optional>true</optional>
         </dependency>
 
+        <!-- Pub/Sub Lite -->
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>spring-cloud-gcp-pubsublite</artifactId>
+            <version>2.0.3-SNAPSHOT</version>
+            <optional>true</optional>
+        </dependency>
+
         <!-- Cloud SQL -->
         <dependency>
             <groupId>org.springframework</groupId>
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsublite/GcpPubSubLiteAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsublite/GcpPubSubLiteAutoConfiguration.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.spring.autoconfigure.pubsublite;
 
+import java.io.IOException;
+
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.cloud.pubsublite.CloudZone;
 import com.google.cloud.pubsublite.cloudpubsub.FlowControlSettings;
@@ -29,7 +31,7 @@ import com.google.cloud.spring.pubsublite.support.DefaultPublisherFactory;
 import com.google.cloud.spring.pubsublite.support.DefaultSubscriberFactory;
 import com.google.cloud.spring.pubsublite.support.PublisherFactorySettings;
 import com.google.cloud.spring.pubsublite.support.SubscriberFactorySettings;
-import java.io.IOException;
+
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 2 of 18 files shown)


spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -71,6 +71,14 @@
             <optional>true</optional>
         </dependency>
 
+        <!-- Pub/Sub Lite -->
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>spring-cloud-gcp-pubsublite</artifactId>
+            <version>2.0.3-SNAPSHOT</version>
```
ttomsu, line 78: The version should be pulled in automatically, so you shouldn't need to specify here.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsublite/GcpPubSubLiteAutoConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+		this.finalCredentialsProvider = properties.getCredentials().hasKey()
+				? new DefaultCredentialsProvider(properties)
+				: credentialsProvider;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public PublisherFactory defaultPublisherFactory() {
+		PublisherFactorySettings settings = PublisherFactorySettings.newBuilder()
+				.setLocation(CloudZone.parse(properties.getLocation()))
```
ttomsu, line 71: We have `matchIfMissing=true` up top, but from what I can tell the user must specify a location (is there a default?), so this may cause an NPE.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -52,12 +52,13 @@
...
 	 * @since 1.1
 	 */
+	@Override
 	Subscriber subscribe(String subscription, Consumer<BasicAcknowledgeablePubsubMessage> messageConsumer);
 
 	/**
 	 * Add a callback method to an existing subscription that receives Pub/Sub messages converted to the requested
 	 * payload type.
-	 * <p>The created {@link Subscriber} is returned so it can be stopped.
+	 * <p>The created {@link SubscriberInterface} is returned so it can be stopped.
```
ttomsu, line 61: It looks like the return type wasn't changed - did you me to also return a `SubscriberInterface` instead of a `Subscriber`?

If so, this may be a small breaking change since this is in our public API

spring-cloud-gcp-pubsublite/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+
+  <artifactId>spring-cloud-gcp-pubsublite</artifactId>
+  <name>Spring Cloud GCP Module - Pub/Sub Lite</name>
+  <description>Spring Cloud GCP Pub/Sub Lite Module</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-pubsublite</artifactId>
+      <version>0.13.0</version>
```
ttomsu, line 19: Is Pubsub Lite included in the Google libraries BOM? If so, I think we'd prefer to use that version vs. hardcoding here.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberOperations.java
dpcollins-google: I intentionally tried to avoid this: I'll revert the comment change, but I created a lower level interface and overrode the return type in existing classes. This should not be a breaking change.

spring-cloud-gcp-pubsublite/pom.xml
dpcollins-google: It is not currently.

spring-cloud-gcp-autoconfigure/pom.xml
dpcollins-google: It causes build errors if I don't, I'm unsure why.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsublite/GcpPubSubLiteAutoConfiguration.java
dpcollins-google: Are you suggesting I remove `matchIfMissing=true`? I just copied these annotations from pubsub, there is no thought put into them.

> the user must specify a location
This is correct, pubsub lite is a zonal service, there is no default.

spring-cloud-gcp-autoconfigure/pom.xml
dzou: You'll have to add pubsublite as an entry to [spring-cloud-gcp-dependencies](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-dependencies/pom.xml#L89) so that the version gets registered. It should allow you to remove hard-coding the version here after that.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsublite/GcpPubSubLiteAutoConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureAfter(GcpContextAutoConfiguration.class)
+@ConditionalOnProperty(value = "spring.cloud.gcp.pubsublite.enabled", matchIfMissing = true)
+@ConditionalOnClass(PublisherFactorySettings.class)
+@EnableConfigurationProperties(GcpPubSubLiteProperties.class)
+public class GcpPubSubLiteAutoConfiguration {
```
meltsufin, line 48: I think you intend to run this auto configuration before `PubSubAutoConfiguration` because you are trying to overwrite the `PublisherFactory` and `SubscriberFactory` beans.
In that case, use `@AutoConfigureBefore(PubSubAutoConfiguration.class)`.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsublite/GcpPubSubLiteAutoConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+				.setLocation(CloudZone.parse(properties.getLocation()))
+				.setProjectIdProvider(finalProjectIdProvider)
+				.setCredentialsProvider(finalCredentialsProvider)
+				.build();
+		return new DefaultPublisherFactory(settings);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public StreamingSubscriberFactory defaultSubscriberFactory() {
```
meltsufin, line 80: What about `PubSubSubscriberOperations`? Can that be supported with Pub/Sub Lite?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubStreamingSubscriberOperations.java
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spring.pubsub.core.subscriber;
+
+import java.util.function.Consumer;
+
+import com.google.cloud.pubsub.v1.SubscriberInterface;
```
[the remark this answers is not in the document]
dpcollins-google: Yes. Subscriber has implementation details about Pub/Sub and implements SubscriberInterface (which has all the non-static methods)

meltsufin (commented):
What I'm seeing here from the high level design is this:

You are trying to re-use some of the interfaces of Pub/Sub for Pub/Sub Lite. This might be OK, but there is a slight problem in that only subscribing to messages is available in Pub/Sub Lite, and no explicit pulling. Did I get that right?

You are trying to combine Pub/Sub with Pub/Sub Lite autoconfiguration. I see this as potentially problematic because Pub/Sub has its own set of configuration properties which would be ignored. Also, it would make it hard to use Pub/Sub and Pub/Sub Lite in the same application.

I would instead look into keeping Pub/Sub and Pub/Sub Lite autoconfiguration completely separate. You might  also need to introduce a pubsub-common module for interfaces and classes shared between Pub/Sub and Pub/Sub Lite.

Furthermore, I'm not sure that implementing PubSubTemplate for Pub/Sub Lite is the best approach that our customers need. Maybe, they would be better off with the Kafka shim and an implementation of Spring Kafka for Pub/Sub Lite.

All of this requires quite a bit more investigation. So, I would suggest putting this PR on hold for now.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsublite/GcpPubSubLiteAutoConfiguration.java
dpcollins-google: There is no synchronous pull for Pub/Sub Lite, only the streaming pull model is supported. This is the one method that can actually be supported: I've split out this operation to a sub-interface, PubSubStreamingSubscriberOperations, but I believe this should not be a breaking change as written.

dpcollins-google:
Responses inline.

> You are trying to re-use some of the interfaces of Pub/Sub for Pub/Sub Lite. This might be OK, but there is a slight problem in that only subscribing to messages is available in Pub/Sub Lite, and no explicit pulling. Did I get that right?

Yes, there is no synchronous pull model for Pub/Sub Lite, only streaming pull. We try to actively discourage users from the synchronous pull API in Cloud Pub/Sub because it has lower performance (higher latency) and is more error prone to use.

> You are trying to combine Pub/Sub with Pub/Sub Lite autoconfiguration. I see this as potentially problematic because Pub/Sub has its own set of configuration properties which would be ignored. Also, it would make it hard to use Pub/Sub and Pub/Sub Lite in the same application.

I don't know what the correct way to do this is, but would appreciate advice. Essentially, I'm trying to implement the same interfaces, and allow users to choose their implementation. What would be the best way to structure this?

> Furthermore, I'm not sure that implementing PubSubTemplate for Pub/Sub Lite is the best approach that our customers need. Maybe, they would be better off with the Kafka shim and an implementation of Spring Kafka for Pub/Sub Lite.

I'd be hesitant to do this, as spring kafka is not owned by us. There are some incompatibilities with the Kafka API, specifically around the lack of transaction support and only supporting a single subscription with the consumer API, and I think the risk of a change introducing a breakage in our integration would be higher.

Issue #389, opened 2021-03-16
Linkage checker failing after recent deluge of Dependabot updates


ttomsu:
Builds 47-50 as of this writing, but check her for the latest: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions?query=branch%3Amain+workflow%3A%22Linkage+Check%22+is%3Afailure

suztomo:
Linkage Checker 1.5.7 fixed a bug where Linkage Checker was not working for spring-cloud-gcp. Now it work as designed but it reports too many. I created an issue on me to investigate these errors.

https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1983#issuecomment-800303299

elefeint:
Obsolete; we decided not to use linkage checker on this project.

PR #428, opened 2021-04-01
Linkage Checker: adding exclusion file for autoconfigure package

suztomo:
Towards https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/389

@meltsufin  As discussed, here is an example filtering XML file. The output of running Linkage Checker (https://gist.github.com/suztomo/7f3a02b246cb0e18e280a69c461fb466) does not show autoconfigure or GcsMessageHandler any more.

After this is merged, would you try adding more rules and share feedback?


meltsufin (approved):
Thanks, Tomo!

PR #436, opened 2021-04-11
Add implementation to GoogleStorageResource.java getFile() method

santfirax:
We would like to add an implementation to getFile() method in GoogleStorageResource.java in order to use it with FlatFileItemWriter and close https://github.com/spring-projects/spring-batch/issues/3854

TODO: I'm not confident about creating new File(path) inside the getFile() method and I'm open to suggestions about how to improve it.

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/com/google/cloud/spring/storage/GoogleStorageResource.java
+++ b/spring-cloud-gcp-storage/src/main/java/com/google/cloud/spring/storage/GoogleStorageResource.java
@@ -33,6 +33,7 @@ import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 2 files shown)


santfirax (commented):
any chance that someone can take a look about how to return a file?

meltsufin:
@santfirax The challenge here is that `java.util.File` is not really the right abstraction for GCS resources. Notice that `GoogleStorageResource` does implement `WritableResource`, which in principle, should be sufficient for something like `FlatFileItemWriter`. In fact, `FlatFileItemWriter` has `setResource()` that should work instead of the `File`. Have you tried passing `GoogleStorageResource` in that way?

santfirax:
Hello @meltsufin I agree that new File is not the best solution. Can you take a look at https://github.com/spring-projects/spring-batch/issues/3854. it is reported setting the resource with GoogleStorageResource

meltsufin:
@santfirax Thanks for the clarification. It looks like we can't get away from the `java.util.File` with `FlatFileItemWriter`. Also see a very detailed comment on this here: https://github.com/spring-projects/spring-batch/issues/3818#issuecomment-818857894. However, I'm not so sure we should implement `getFile()`.
One alternative could be to just use the `FlatFileItemWriter` with a local file and then copy it to GCS in a separate step. Another option would be to just extend `GoogleStorageResource` with your proposed change, but keep it in your own application.
Finally, the option that might be best is to try to implement a `WritableResourceItemWriter` or something like that in the Spring Batch project.

santfirax:
Thank you for your feedback. I'm going to close the PR then.

PR #446, opened 2021-04-21
Add class condition on ManagedChannel

herder:
This fixes the edge case where the project has the `spring.cloud.gcp.pubsub.emulator-host` set but not actually imports the `spring-cloud-gcp-starter-pubsub` starter.

Before this change, this would cause a `NoClassDefFoundError` when the auto config encountered the `GcpPubSubEmulatorAutoConfiguration`.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubEmulatorAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubEmulatorAutoConfiguration.java
@@ -21,6 +21,7 @@ import javax.annotation.PreDestroy;
 import com.google.api.gax.grpc.GrpcTransportChannel;
 import com.google.api.gax.rpc.FixedTransportChannelProvider;
 import com.google.api.gax.rpc.TransportChannelProvider;
+import com.google.cloud.spring.pubsub.core.PubSubTemplate;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 
@@ -40,7 +41,7 @@ import org.springframework.context.annotation.Configuration;
  * @author Mike Eltsufin
  */
 @Configuration(proxyBeanMethods = false)
-@ConditionalOnClass(ManagedChannel.class)
+@ConditionalOnClass({ManagedChannel.class, PubSubTemplate.class})
 @ConditionalOnProperty(prefix = "spring.cloud.gcp.pubsub", name = "emulator-host")
 @AutoConfigureBefore(GcpPubSubAutoConfiguration.class)
 @EnableConfigurationProperties(GcpPubSubProperties.class)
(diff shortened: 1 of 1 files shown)


herder:
@googlebot I signed it!

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubEmulatorAutoConfiguration.java
```diff
@@ -39,6 +40,7 @@
  * @author Mike Eltsufin
  */
 @Configuration(proxyBeanMethods = false)
+@ConditionalOnClass(ManagedChannel.class)
```
meltsufin, line 43: There is a potential problem with using `ManagedChannel` because it's not unique to Pub/Sub and could be brought in by another GCP dependency. I think this class should actually be conditional on the same as `GcpPubSubAutoConfiguration`.

Would this work?
```suggestion
@ConditionalOnProperty(value = "spring.cloud.gcp.pubsub.enabled", matchIfMissing = true)
@ConditionalOnClass(PubSubTemplate.class)
```

meltsufin (changes requested):
Thank you for your contribution. See my comment.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubEmulatorAutoConfiguration.java
herder: Yes, that makes sense. Though theoretically I guess we could have a `PubsubTemplate` on the classpath and not a `ManagedChannel`, which would cause this to break again. 
Maybe we should add `PubSubTemplate` to the list instead of removing  `@ConditionalOnClass(ManagedChannel.class)`? 
```
@ConditionalOnClass({ManagedChannel.class, PubSubTemplate.class})
```

meltsufin: Sure, that works too. Although, gRPC is a core dependency of Pub/Sub, and that's where the `ManagedChannel` is coming from.

herder: Yeah, something would be really off if `ManagedChannel` is not on the classpath when `PubSubTemplate` is, but I guess it could happen.

Added it in ba7464ecae57cd897519b0878d7ef74c8a371e86

meltsufin: I think we also might want `@ConditionalOnProperty(value = "spring.cloud.gcp.pubsub.enabled", matchIfMissing = true)` because there is no reason for this autoconfig to run when Pub/Sub autoconfig is disabled. Right?

herder: Totally agree - added in b195f4d4 :+1:

meltsufin (approved):

Issue #455, opened 2021-05-04
Emulator Enablement & No Credentials does not work


nursultanturdaliev:
**Description**

Trying to use google cloud spanner without credentials for the emulator is not working.

**Libraries In Use**

    implementation platform("com.google.cloud:spring-cloud-gcp-dependencies:2.0.3-SNAPSHOT")
    implementation 'com.google.cloud:spring-cloud-gcp-starter-data-spanner'

**Spring Boot Spanner Starter Config in the application.yaml**

```
spring:
  cloud:
    gcp:
      spanner:
        database: MyDatabase
        project-id: wf-gcp-us-db-dev
        instance-id: MyInstance
        emulator-host: ${SPANNER_EMULATOR_HOST:spanner-emulator:9010}
        emulator:
          enabled: true```

**Error Thrown**

```Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.```


According to this auto config it should have disabled credentials requirement

- https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/spanner/GcpSpannerEmulatorAutoConfiguration.java

**Comment by meltsufin** [comment]
Can you provide a more complete stack trace please?

**Comment by nursultanturdaliev** [comment]
Hi @meltsufin  Here is the full stack trace.


```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'com.google.cloud.spring.autoconfigure.spanner.GcpSpannerAutoConfiguration$CoreSpannerAutoConfiguration': Bean instantiation via constructor failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.google.cloud.spring.autoconfigure.spanner.GcpSpannerAutoConfiguration$CoreSpannerAutoConfiguration]: Constructor threw exception; nested exception is java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
[...]

PR #457, opened 2021-05-04
Fix Spanner Emulator triggering default credentials creation

meltsufin:
Fixes: #455.


elefeint (approved):
This will work. I think we may have been overriding client library's default behavior by setting credentials in the first place -- client library has a clause that will automatically set to `NoCredentials` if emulator is set -- https://github.com/googleapis/java-spanner/blob/7564c11dcf4e06a85d288c88754fe0f0acf47e26/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L1019

meltsufin:
> This will work. I think we may have been overriding client library's default behavior by setting credentials in the first place -- client library has a clause that will automatically set to `NoCredentials` if emulator is set -- https://github.com/googleapis/java-spanner/blob/7564c11dcf4e06a85d288c88754fe0f0acf47e26/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L1019

Either way, if we don't create the bean, our core starter will initialize the default credentials provider which will cause this problem.

PR #466, opened 2021-05-17
Add jdbc-socket-factory-core version

herder:
This is used by some services we use, and it would be convenient to be able to not have to keep track of the version manually.


meltsufin (approved):
Thank you!

PR #471, opened 2021-05-25
Revert "Bump spring-boot-starter-parent from 2.4.5 to 2.5.0"

meltsufin:
Reverts GoogleCloudPlatform/spring-cloud-gcp#468

Integration tests are failing and these are not just flakes.
For example `SecretManagerSampleIntegrationTests` failures can be reproduced locally.
Getting
` No converter found capable of converting from type [com.google.protobuf.ByteString$LiteralByteString] to type [java.lang.String]` which seems to indicate the converter registration in bootstrap autoconfig no longer works in Spring Boot 2.5.


elefeint (approved):
Yeah, that's a major version change. Will be work...

PR #482, opened 2021-05-28
Pub/Sub publish callbacks with original Message awareness

### LINKED ISSUE #467: PubSub Message Handler Async At-least-once Delivery
**Problem**

There is no way to ship the data into GCP PubSub via using com.google.cloud.spring.pubsub.integration.outbound.PubSubMessageHandler & with meeting as follows requirements:
* do it asynchronously (with batching is plus)
* be able programatically achieve finally at-least-once delivery guarantee

To be more descriptive here:
* configure PubSubMessageHandler in async mode
* in this case there is only "publish callback" to be provided for operation status notification
* "publish callback" provides only GCP PubSub message ID as input parameter
* so in "publish callback" original input message is missed + headers
* as result there is no way to correlate original input message (submitted into PubSubMessageHandler) with GCP PubSub response
* as result no way to mark original input (source) message as delivered or redeliver it if something happened

        PubSubMessageHandler adapter = new PubSubMessageHandler(pubsubTemplate, topic);

        adapter.setSync(true);

        adapter.setPublishCallback(new ListenableFutureCallback<String>() {
            @Override
            public void onFailure(Throwable ex) {
                log.debug("There was an error sending the message", ex);
            }

            @Override
            public void onSuccess(String result) {
                log.debug("Message was sent successfully - result: {}", result);
            }
        });

And would like to emphasise that trying to achieve at-least-once delivery guarantee. So "sigkill" should be handled by my app without data loss.

Consequently, at the moment custom PubSubMessageHandler implementation is required to achieve what was described above. Or please correct me if I've missed something. 
 
**Versions**
* spring-cloud.version --> 2020.0.2
* spring-cloud-gcp.version --> 2.0.2

Do similar as in org.springframework.integration.http.outbound.HttpRequestExecutingMessageHandler:
* add possibility to specify "OutputChannel"
* in a message delegated into "OutputChannel" preserve message headers of original input message

Having this will be a possibility to correlate PubSub response with input source message. And mark/rollback it if needed.

elefeint:
Two callbacks provided: `setSuccessCallback()` and `setFailureCallback()`. 
Second parameter to both is the original `Message` for consistency, and to enable forwarding to downstream channels.

Fixes #467.

### DIFF
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -262,8 +262,10 @@ The provided `PubSubTemplate` contains all the necessary configuration to publis
 A publish timeout can be configured for synchronous publishing.
 If none is provided, the adapter waits indefinitely for a response.
 
-It is possible to set user-defined callbacks for the `publish()` call in `PubSubMessageHandler` through the `setPublishFutureCallback()` method.
-These are useful to process the message ID, in case of success, or the error if any was thrown.
+It is possible to set user-defined callbacks for the `publish()` call in `PubSubMessageHandler` through the `setSuccessCallback()` and `setFailureCallback()` methods (either one or both may be set).
+These give access to the Pub/Sub publish message ID in case of success, or the root cause exception in case of error.
+Both callbacks include the original message as the second argument.
+The old `setPublishFutureCallback()` methods that only gave access to message ID or root cause exception is deprecated and will be removed in a future release.
 
 [source,java,indent=0]
 ----
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -265,7 +265,7 @@ If none is provided, the adapter waits indefinitely for a response.
 It is possible to set user-defined callbacks for the `publish()` call in `PubSubMessageHandler` through the `setSuccessCallback()` and `setFailureCallback()` methods (either one or both may be set).
 These give access to the Pub/Sub publish message ID in case of success, or the root cause exception in case of error.
 Both callbacks include the original message as the second argument.
-The old `setPublishFutureCallback()` methods that only gave access to message ID or root cause exception is deprecated and will be removed in a future release.
+The old `setPublishFutureCallback()` method that only gave access to message ID or root cause exception is deprecated and will be removed in a future release.
 
 [source,java,indent=0]
 ----
--- a/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/outbound/PubSubMessageHandler.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/outbound/PubSubMessageHandler.java
@@ -23,6 +23,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import com.google.cloud.spring.pubsub.core.publisher.PubSubPublisherOperations;
+import com.google.cloud.spring.pubsub.core.publisher.PubSubPublisherTemplate;
 import com.google.cloud.spring.pubsub.integration.PubSubHeaderMapper;
 import com.google.cloud.spring.pubsub.support.GcpPubSubHeaders;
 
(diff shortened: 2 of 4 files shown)


spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -130,11 +134,31 @@ public void setPublishTimeout(long timeoutMillis) {
 	/**
 	 * Set the callback to be activated when the publish call resolves.
 	 * @param publishCallback callback for the publish future
+	 * @deprecated Use {@code setSuccessCallback()} and {@code setFailureCallback()} instead.
```
meltsufin, line 137: You can use `{@link` here.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -130,11 +134,31 @@ public void setPublishTimeout(long timeoutMillis) {
...
 	 */
+	@Deprecated
 	public void setPublishCallback(ListenableFutureCallback<String> publishCallback) {
 		this.publishCallback = publishCallback;
 	}
 
+	/**
+	 * Set callback (can be a lambda) for processing the published message ID and the original {@code Message} after the
+	 * message was successfully published.
+	 * @param cb callback accepting a {@code String} message ID and the original {@code Message}.
```
meltsufin, line 147: `cb` seems to be different from the long-form style. `successCallback`?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -130,11 +134,31 @@ public void setPublishTimeout(long timeoutMillis) {
...
+	public void setSuccessCallback(SuccessCallback cb) {
+		this.successCallback = cb;
+	}
+
+	/**
+	 * Set callback (can be a lambda) for processing the root cause exception and the original {@code Message} in case
+	 * of failure.
+	 * @param cb callback accepting a {@code Throwable} and a {@code Message}.
+	 */
+	public void setFailureCallback(FailureCallback cb) {
```
meltsufin, line 158: `failureCallback`?

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -262,8 +262,10 @@ The provided `PubSubTemplate` contains all the necessary configuration to publis
 A publish timeout can be configured for synchronous publishing.
 If none is provided, the adapter waits indefinitely for a response.
 
-It is possible to set user-defined callbacks for the `publish()` call in `PubSubMessageHandler` through the `setPublishFutureCallback()` method.
-These are useful to process the message ID, in case of success, or the error if any was thrown.
+It is possible to set user-defined callbacks for the `publish()` call in `PubSubMessageHandler` through the `setSuccessCallback()` and `setFailureCallback()` methods (either one or both may be set).
+These give access to the Pub/Sub publish message ID in case of success, or the root cause exception in case of error.
```
meltsufin, line 266: "message ID" always seems confusing to me. What we're really talking about is the "message ack ID", right?

meltsufin (approved):
Looks good, just some minor naming comments.

docs/src/main/asciidoc/spring-integration-pubsub.adoc
elefeint: The confusion goes down to proto level. Publish operation returns [server-assigned message ID](https://github.com/googleapis/googleapis/blob/6afe7923cbf18ac9de4163fe1af316c74540b46b/google/pubsub/v1/pubsub.proto#L277). But when we acknowledge, it's called ["ack ID"](https://github.com/googleapis/googleapis/blob/6afe7923cbf18ac9de4163fe1af316c74540b46b/google/pubsub/v1/pubsub.proto#L1056).

meltsufin: When you pull a message, you get an ack ID with it, which you can use to acknowledge it.
I don't think you get anything back when you acknowledge a message.

elefeint: Right, but we are not pulling a message. We are _publishing_ a message, so we are getting that message's ID back.

Issue #474, opened 2021-05-25
Support for IAM database authentication


breun:
Cloud SQL supports IAM logins: https://cloud.google.com/sql/docs/postgres/iam-logins

Support for this has recently been added to the Cloud SQL Java socket factory library: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/350

Could Spring Cloud GCP expose support for this?

meltsufin:
This is blocked on the release of the Cloud SQL Java socket factory library. Once that's done, we'll take a look at what's involved, but otherwise we also welcome external contributions.

meltsufin:
The release has arrived: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/releases/tag/v1.3.0

meltsufin:
I started looking into this, but having some issues getting it to work.
See: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/515

mzeijen:
Thanks Mike!

PR #488, opened 2021-06-03
Add support for IAM database authentication

meltsufin:
Introduces `spring.cloud.gcp.sql.enableIamAuth=true` for enabling IAM authentication,
which currently only works on PostgreSQL.

Fixes: #474.

### DIFF
--- a/docs/src/main/asciidoc/sql.adoc
+++ b/docs/src/main/asciidoc/sql.adoc
@@ -71,6 +71,7 @@ Used to authenticate and authorize new connections to a Google Cloud SQL instanc
 | `spring.cloud.gcp.sql.credentials.encoded-key` | Base64-encoded contents of OAuth2 account private key in JSON format.
 Used to authenticate and authorize new connections to a Google Cloud SQL instance. | No
 | Default credentials provided by the Spring GCP Boot starter
+| `spring.cloud.gcp.sql.enableIamAuth` | Specifies whether to enable IAM database authentication (PostgreSQL only). | No | `False`
 | `spring.datasource.username` | Database username | No | MySQL: `root`; PostgreSQL: `postgres`
 | `spring.datasource.password` | Database password | No | `null`
 | `spring.datasource.driver-class-name` | JDBC driver to use. | No | MySQL: `com.mysql.cj.jdbc.Driver`; PostgreSQL: `org.postgresql.Driver`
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/sql.adoc
```diff
@@ -90,6 +91,17 @@ You can select the type of connection pool (e.g., Tomcat, HikariCP, etc.) by htt
...
+==== Cloud SQL IAM database authentication
+
+Currently, Cloud SQL only supports https://cloud.google.com/sql/docs/postgres/authentication:[IAM database authentication for PostgreSQL].
+It allows you to connect to the database using an IAM account, rather than a predefined database username and password.
+You will need to do the following to enable it:
+
+. In your database instance settings, turn on the `cloudsql.iam_authentication` flag.
+. Add the IAM user or service account to the list of database users.
+. In the application settings, set `spring.cloud.gcp.sql.enableIamAuth` to `true`.
+Note that this will also set the `sslmode` to `disabled`, as it's required for IAM authentication to work.
```
elefeint, line 103: Is there Cloud docs we can point to? Hell of a thing to turn off SSL to make things more secure.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/GcpCloudSqlProperties.java
```diff
@@ -40,6 +40,9 @@
 	/** Overrides the GCP OAuth2 credentials specified in the Core module. */
 	private Credentials credentials = new Credentials();
 
+	/** Specifies whether to enable IAM database authentication (PostgreSQL only). */
+	private boolean enableIamAuth;
```
elefeint, line 44: maybe just `iamAuth`? `isIamAuth()` reads easier.

elefeint (approved):
Does the feature exist on MySQL? If not, might want to clarify that we don't implement it due to it not being available.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/GcpCloudSqlProperties.java
meltsufin: I'm trying to keep the property name consistent with the socket factory.

docs/src/main/asciidoc/sql.adoc
meltsufin: > Does the feature exist on MySQL? If not, might want to clarify that we don't implement it due to it not being available.

The code change actually will also pass the property through for MySQL connections, but until MySQL gets IAM auth, it will do nothing. I believe it's coming to MySQL.

meltsufin: I believe the connection is already encrypted, but just in case, I asked the question here: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/490#issuecomment-854152505

PR #489, opened 2021-06-04
Add customizer for SpannerOptions object

### LINKED ISSUE #480: Facilitate Spanner client timeout configuration

I'd like to be able to configure timeout and retry settings for the Spanner client, per [Google's guidance to use `SpannerOptions`](https://cloud.google.com/spanner/docs/custom-timeout-and-retry#java). At the same time, I'd like to continue leveraging the auto-configuration of the core `SpannerOptions` properties (projectId, credentials, numRpcChannels, SessionPoolOptions, etc.) that's done in `spring-cloud-gcp-autoconfigure`, in [`GcpSpannerAutoConfiguration`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/366b3d9530efda9b9b1e7dcd92d0284a38034c25/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/spanner/GcpSpannerAutoConfiguration.java#L124-L170). There might be some Spring configuration trick to do this, but I'm not aware of it.

I'm using version `1.2.5.RELEASE` (`org.springframework.cloud:spring-cloud-gcp-starter-data-spanner:1.2.5.RELEASE`), but it looks like https://github.com/spring-cloud/spring-cloud-gcp (for v1) isn't taking feature requests anymore, and this still seems to be relevant here in the v2 releases.

I'd like to be able to configure timeout settings (particularly for `ExecuteSqlRequest`s), which to my knowledge is done by having access to a `SpannerOptions` builder; but I'd also like the `SpannerOptions` that gets used for my queries to have the rest of its configuration provided by the auto-configuration that's done in the `spring-cloud-gcp` project (specifically, in the `GcpSpannerAutoConfiguration` class).

I don't know what the best approach would be, but maybe it could involve exposing a `SpannerOptions.Builder` that's had its initial configuration done through the existing Spring auto-configuration, but that I could then configure further.

I've considered autowiring my own `SpannerOptions` bean in my app, to completely supersede the auto-configured one:

```java
	@Bean
	public SpannerOptions spannerOptions(SessionPoolOptions sessionPoolOptions) {
		// build a SpannerOptions from scratch and return it
	}
```

But, at least with my limited Spring experience, this would require somehow reproducing the logic and configuration that's done in `GcpSpannerAutoConfiguration` to set the projectId, credentials, etc. And even if I found a reasonable way to do that, this bean might now be at risk of diverging from any changes to the auto-configuration provided by future versions of this project.

If there's already a way to do this cleanly, I'd love to know.

**Additional context**

Probably not relevant, but I'm actually using Kotlin, not Java.

elefeint:
I looked at providing individual option overrides, but there are too many options that are not very relevant to Spring fundamentals. So we might as well introduce a customizer that works for any need. I put it in autoconfiguration project because it's not really useful for any other scenario -- without Spring Boot, users would be creating their own `SpannerOptions` in the first place.

Fixes #480.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
@@ -52,6 +52,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
 import org.springframework.data.rest.webmvc.spi.BackendIdConverter;
 
 /**
@@ -123,9 +124,11 @@ public class GcpSpannerAutoConfiguration {
 		}
 
 		@Bean
+		@Scope("prototype")
 		@ConditionalOnMissingBean
-		public SpannerOptions spannerOptions(SessionPoolOptions sessionPoolOptions,
-											Optional<SpannerOptionsCustomizer> customizer) {
+		public SpannerOptions.Builder spannerOptionsBuilder(
+				SessionPoolOptions sessionPoolOptions,
+				Optional<SpannerOptionsCustomizer> customizer) {
 			Builder builder = SpannerOptions.newBuilder()
 					.setProjectId(this.projectId)
 					.setHeaderProvider(new UserAgentHeaderProvider(this.getClass()))
@@ -137,8 +140,15 @@ public class GcpSpannerAutoConfiguration {
 				builder.setPrefetchChunks(this.prefetchChunks);
 			}
 			builder.setSessionPoolOption(sessionPoolOptions);
+
 			customizer.ifPresent(c -> c.apply(builder));
 
+			return builder;
+		}
+
+		@Bean
+		@ConditionalOnMissingBean
+		public SpannerOptions spannerOptions(SpannerOptions.Builder builder) {
 			return builder.build();
 		}
 
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
@@ -17,6 +17,7 @@
 package com.google.cloud.spring.autoconfigure.spanner;
 
 import java.io.IOException;
+import java.util.Optional;
 import java.util.function.Supplier;
 
 import com.google.api.gax.core.CredentialsProvider;
@@ -123,7 +124,8 @@ public class GcpSpannerAutoConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
-		public SpannerOptions spannerOptions(SessionPoolOptions sessionPoolOptions) {
+		public SpannerOptions spannerOptions(SessionPoolOptions sessionPoolOptions,
+											Optional<SpannerOptionsCustomizer> customizer) {
 			Builder builder = SpannerOptions.newBuilder()
 					.setProjectId(this.projectId)
 					.setHeaderProvider(new UserAgentHeaderProvider(this.getClass()))
@@ -135,6 +137,8 @@ public class GcpSpannerAutoConfiguration {
 				builder.setPrefetchChunks(this.prefetchChunks);
 			}
 			builder.setSessionPoolOption(sessionPoolOptions);
+			customizer.ifPresent(c -> c.apply(builder));
+
 			return builder.build();
 		}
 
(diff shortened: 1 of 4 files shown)


meltsufin (commented):
Have you considered just making `SpannerOptions.Builder` a bean that can be overwritten? Would that be sufficient?

elefeint:
Maybe. `SpannerOptions` already overridable. If we make `SpannerOptions.Builder` overridable, too, then our customizations will be on top of the user's customizations (because we'll take the user-provided builder bean). With a customizer, the user's changes go on top of our base image. 
I can't decide which one is more intuitive. Thoughts?

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/spanner/GcpSpannerAutoConfigurationTests.java
```diff
@@ -107,6 +110,24 @@ public void testIdConverterNotCreated() {
...
+	public void testSpannerCustomizerProvided() {
+		Duration duration = Duration.ofSeconds(42);
+		this.contextRunner.withBean(SpannerOptionsCustomizer.class, () -> {
+			return builder -> {
+				builder.getSpannerStubSettingsBuilder()
+						.executeSqlSettings()
+						.setRetrySettings(RetrySettings.newBuilder().setMaxRetryDelay(duration).build());
+			};
+		}).run(context -> {
+			SpannerOptions spannerOptions = context.getBean(SpannerOptions.class);
```
joshrai, line 123: Ignore this if it seems like overkill, but do you think it would be worth adding some assertion that the "core" configuration of `SpannerOptions` provided by `GcpSpannerAutoConfiguration` has been preserved? For example (though maybe getting too far into implementation details), if there were a way to set or mock `CoreSpannerAutoConfiguration.numRpcChannels`, and then assert that `spannerOptions.getNumChannels()` has the same value?

elefeint: I am always in favor of more testing.

meltsufin:
> Maybe. `SpannerOptions` already overridable. If we make `SpannerOptions.Builder` overridable, too, then our customizations will be on top of the user's customizations (because we'll take the user-provided builder bean). With a customizer, the user's changes go on top of our base image.
> I can't decide which one is more intuitive. Thoughts?

I'm not a huge fan of this partial overwriting of a bean whether it's user first or framework first. I would either expose all of the options that our customers need, or let the user just overwrite the whole bean.

elefeint:
@meltsufin I wrote the coarse `RetrySettings` overriding solution, and, well, I learned [something new](https://github.com/googleapis/java-spanner/blob/318e1074df7b9edd38db027a9f0504b44e5bdc05/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L810) -- Spanner disallows setting the coarse setting in the first place, since there are several disparate RPC stubs involved.

So, back to the drawing board on having to implement the granular, per-stub settings: it's down to prototype-scoped builder bean vs pre-Options-constructions customizer. I think the customizer will incur fewer unexpected side effects for the users, since it's pleasantly immutable, at a cost of an extra interface (in the future, per product).

meltsufin:
> @meltsufin I wrote the coarse `RetrySettings` overriding solution, and, well, I learned [something new](https://github.com/googleapis/java-spanner/blob/318e1074df7b9edd38db027a9f0504b44e5bdc05/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L810) -- Spanner disallows setting the coarse setting in the first place, since there are several disparate RPC stubs involved.
> 
> So, back to the drawing board on having to implement the granular, per-stub settings: it's down to prototype-scoped builder bean vs pre-Options-constructions customizer. I think the customizer will incur fewer unexpected side effects for the users, since it's pleasantly immutable, at a cost of an extra interface (in the future, per product).

Interesting. Sure, go ahead with the customizer.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -123,7 +124,8 @@
 
 		@Bean
 		@ConditionalOnMissingBean
-		public SpannerOptions spannerOptions(SessionPoolOptions sessionPoolOptions) {
+		public SpannerOptions spannerOptions(SessionPoolOptions sessionPoolOptions,
+											Optional<SpannerOptionsCustomizer> customizer) {
 			Builder builder = SpannerOptions.newBuilder()
```
meltsufin, line 129: Do you want to make this builder a bean?

meltsufin (approved):

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
elefeint: Sure. With all the configured properties, right? And then the `SpannerOptions` bean will only apply customizer and build.

meltsufin: Why not apply the customizer in the builder bean? What does Spring do in such cases?

PR #492, opened 2021-06-07
[Work in Progress] Tentative fix for secretmanager

dzou:
Tentative fix for secret manager to upgrade to spring 2.5.0 but doesn't work for spring version 2.4.x


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
```diff
@@ -62,16 +63,16 @@ public GcpSecretManagerBootstrapConfiguration(
 				? properties::getProjectId
 				: new DefaultGcpProjectIdProvider();
 
-		// Registers {@link ByteString} type converters to convert to String and byte[].
-		configurableEnvironment.getConversionService().addConverter(
```
meltsufin, line 66: Can we just register in both places to make it work for 2.4 and 2.5?

dzou: Yeah that would work. Following the thread I wonder if the Spring may know of a better solution.

Issue #491, opened 2021-06-07
No environment placeholder in configuration properties


ElectroLutz:
Hey,

I guess that after #131 (edit: I guess it's since 2.0.2) it's not possible to have a configuration like the following anymore:
```
spring.cloud.gcp:
  sql:
    enabled: ${CLOUD_SQL_ENABLED:false}
    database-name: ${CLOUD_SQL_DATABASE:}
    instance-connection-name: ${CLOUD_SQL_INSTANCE:}
    credentials:
      encoded-key: ${CLOUD_SQL_KEY:}
```
The placeholders are not replaced anymore and the values are passed as they are in the application.yml.

How can I archive that they are replaced again? I don't want to use any config management, because until now the environment variables been doing their job.

Cheers
Marko

meltsufin:
I think this is more likely caused by #273.
We disabled resolving placeholders to avoid issues with Secrets Manager placeholders resolver, but maybe we should just register a smarter resolver, or create an option to enable placeholder resolution.
cc/ @dzou.

ElectroLutz:
Thank you for the fast fix! Didn’t expect that.

PR #495, opened 2021-06-08
Allow placeholders in Cloud SQL config

meltsufin:
Secret Manager placehoders are still skipped as before.

Fixes: #491.


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -67,8 +69,20 @@ public void postProcessEnvironment(ConfigurableEnvironment environment, SpringAp
 				LOGGER.info("post-processing Cloud SQL properties for + " + databaseType.name());
 			}
 
-			// Bind properties without resolving placeholders
-			Binder binder = new Binder(ConfigurationPropertySources.get(environment), null, null, null, null);
+			// Bind properties without resolving Secret Manager placeholders
+			Binder binder = new Binder(ConfigurationPropertySources.get(environment), new PlaceholdersResolver() {
```
elefeint, line 73: Want to refactor this into a nested class? It's not so trivial, plus there are additional parameters after the anonymous class...

elefeint (approved):

dzou (approved):

Issue #504, opened 2021-06-15
Replace __name__ with FieldPath.documentId() in Firestore querying code


elefeint:
Name projections are using `__name__`; the constant literal is an implementation detail, and it would be better to use `FieldPath.documentId()` instead.

elefeint:
cc/ @mpeddada1 -- this is relevant to #424.

elefeint:
Fixed in #506.

PR #506, opened 2021-06-18
Use special property to query by document Id in Firestore

### LINKED ISSUE #424: Firestore `findByIdAndArrayContains` / `existsByIdAndArrayContains` doesn't work
Hi!

I'd like to use GCP Firestore integration. Most things seem to work, except some specific queries:

Methods doing `findById` / `existsById` combined by AND with a contains on a list do not work.

**Sample**
```java
@Document(collectionName = "examples")
public class Example {
    @DocumentId
    private String id;
    private List<String> array;
}
```
```java
public interface ExampleRepository extends FirestoreReactiveRepository<Example> {
   
    Mono<Boolean> existsByIdAndArrayContains(String id, String someString); // does not work
    Mono<Boolean> existsByIdAndArrayContaining(String id, String someString); // does not work
    Mono<Boolean> existsByIdAndArrayIsContaining(String id, String someString); // does not work
    Mono<Boolean> existsByArrayContains(String someString); // works
    Mono<Boolean> existsById(String id); // works

    Mono<Example> findByIdAndArrayContains(String id, String someString); // does not work
    Mono<Example> findByIdAndArrayContaining(String id, String someString); // does not work
    Mono<Example> findByIdAndArrayIsContaining(String id, String someString); // does not work
    Mono<Example> findByArrayContains(String someString); // works
    Mono<Example> findById(String id); // works
```

`byId` and `ArrayContains` work alone but not together.

Did I miss something or is that a bug?

Thanks!

mpeddada1:
Fixes #424
As described in #504, this PR also replaces the use of the constant literal `__name__` with `FieldPath.documentId()`.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
@@ -24,6 +24,8 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 
 import com.google.cloud.Timestamp;
+import com.google.cloud.firestore.DocumentReference;
+import com.google.cloud.firestore.Firestore;
 import com.google.cloud.firestore.Internal;
 import com.google.cloud.spring.data.firestore.mapping.FirestoreClassMapper;
 import com.google.cloud.spring.data.firestore.mapping.FirestoreMappingContext;
@@ -76,7 +78,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 
 	private static final String NOT_FOUND_DOCUMENT_MESSAGE = "NOT_FOUND: Document";
 
-	private final FirestoreStub firestore;
+	private final FirestoreStub firestoreStub;
 
 	private final String parent;
 
@@ -84,25 +86,28 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 
 	private final FirestoreMappingContext mappingContext;
 
+	private final Firestore firestore;
+
 	private Duration writeBufferTimeout = Duration.ofMillis(500);
 
 	private int writeBufferSize = FIRESTORE_WRITE_MAX_SIZE;
 
 	/**
 	 * Constructor for FirestoreTemplate.
-	 * @param firestore Firestore gRPC stub
+	 * @param firestoreStub Firestore gRPC stub
 	 * @param parent the parent resource. For example:
 	 *     projects/{project_id}/databases/{database_id}/documents
 	 * @param classMapper a {@link FirestoreClassMapper} used for conversion
 	 * @param mappingContext Mapping Context
 	 */
-	public FirestoreTemplate(FirestoreStub firestore, String parent, FirestoreClassMapper classMapper,
-			FirestoreMappingContext mappingContext) {
-		this.firestore = firestore;
+	public FirestoreTemplate(FirestoreStub firestoreStub, String parent, FirestoreClassMapper classMapper,
+							 FirestoreMappingContext mappingContext, Firestore firestore) {
+		this.firestoreStub = firestoreStub;
 		this.parent = parent;
 		this.databasePath = Util.extractDatabasePath(parent);
 		this.classMapper = classMapper;
 		this.mappingContext = mappingContext;
+		this.firestore = firestore;
 	}
 
 	/**
@@ -175,7 +180,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 			Optional<TransactionContext> transactionContext = ctx.getOrEmpty(TransactionContext.class);
 			if (transactionContext.isPresent()) {
 				ReactiveFirestoreResourceHolder holder = (ReactiveFirestoreResourceHolder) transactionContext.get()
-						.getResources().get(this.firestore);
+						.getResources().get(this.firestoreStub);
 				//In a transaction, all write operations should be sent in the commit request, so we just collect them
 				return Flux.from(instances).doOnNext(t -> {
 					holder.getWrites().add(createUpdateWrite(t));
@@ -260,7 +265,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 
 	private FirestoreReactiveOperations withParent(String resourceName) {
 		FirestoreTemplate firestoreTemplate =
-				new FirestoreTemplate(this.firestore, resourceName, this.classMapper, this.mappingContext);
+				new FirestoreTemplate(this.firestoreStub, resourceName, this.classMapper, this.mappingContext, this.firestore);
 		firestoreTemplate.setWriteBufferSize(this.writeBufferSize);
 		firestoreTemplate.setWriteBufferTimeout(this.writeBufferTimeout);
 
@@ -276,7 +281,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 			Optional<TransactionContext> transactionContext = ctx.getOrEmpty(TransactionContext.class);
 			if (transactionContext.isPresent()) {
 				ReactiveFirestoreResourceHolder holder = (ReactiveFirestoreResourceHolder) transactionContext.get()
-						.getResources().get(this.firestore);
+						.getResources().get(this.firestoreStub);
 				List<Write> writes = holder.getWrites();
 				//In a transaction, all write operations should be sent in the commit request, so we just collect them
 				return Flux.from(documentNames).doOnNext(t -> writes.add(createDeleteWrite(t)));
@@ -295,7 +300,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 					batch.forEach(e -> builder.addWrites(converterToWrite.apply(e)));
 
 					return ObservableReactiveUtil
-							.<CommitResponse>unaryCall(obs -> this.firestore.commit(builder.build(), obs))
+							.<CommitResponse>unaryCall(obs -> this.firestoreStub.commit(builder.build(), obs))
 							.flatMapMany(
 									response -> {
 										if (setUpdateTime) {
@@ -335,7 +340,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 			doIfTransaction(ctx, resourceHolder -> requestBuilder.setTransaction(resourceHolder.getTransactionId()));
 
 			return ObservableReactiveUtil
-					.<RunQueryResponse>streamingCall(obs -> this.firestore.runQuery(requestBuilder.build(), obs))
+					.<RunQueryResponse>streamingCall(obs -> this.firestoreStub.runQuery(requestBuilder.build(), obs))
 					.filter(RunQueryResponse::hasDocument)
 					.map(RunQueryResponse::getDocument);
 		});
@@ -353,7 +358,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 				builder.setMask(documentMask);
 			}
 
-			return ObservableReactiveUtil.<Document>unaryCall(obs -> this.firestore.getDocument(builder.build(), obs))
+			return ObservableReactiveUtil.<Document>unaryCall(obs -> this.firestoreStub.getDocument(builder.build(), obs))
 					.onErrorResume(throwable -> throwable.getMessage().startsWith(NOT_FOUND_DOCUMENT_MESSAGE),
 							throwable -> Mono.empty());
 		});
@@ -363,7 +368,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 		Optional<TransactionContext> transactionContext = ctx.getOrEmpty(TransactionContext.class);
 		transactionContext.ifPresent(transactionCtx -> {
 			ReactiveFirestoreResourceHolder holder = (ReactiveFirestoreResourceHolder) transactionCtx.getResources()
-					.get(this.firestore);
+					.get(this.firestoreStub);
 			if (!holder.getWrites().isEmpty()) {
 				throw new FirestoreDataException(
 						"Read operations are only allowed before write operations in a transaction");
@@ -442,5 +447,9 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 
 	public FirestoreClassMapper getClassMapper() {
 		return this.classMapper;
+	}
+
+	public DocumentReference getDocumentReference(String collection, String documentId){
+		return this.firestore.collection(collection).document(documentId);
 	}
 }
--- a/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
@@ -66,7 +66,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 
 	private static final int FIRESTORE_WRITE_MAX_SIZE = 500;
 
-	private static final String NAME_FIELD = "__name__";
+	public static final String NAME_FIELD = "__name__";
 
 	private static final StructuredQuery.Projection ID_PROJECTION = StructuredQuery.Projection.newBuilder()
 			.addFields(StructuredQuery.FieldReference.newBuilder().setFieldPath(NAME_FIELD).build())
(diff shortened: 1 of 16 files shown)


spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
```diff
@@ -62,61 +65,75 @@
...
 
 	private static final int FIRESTORE_WRITE_MAX_SIZE = 500;
 
-	private static final String NAME_FIELD = "__name__";
+	private static final String NOT_FOUND_DOCUMENT_MESSAGE = "NOT_FOUND: Document";
 
-	private static final StructuredQuery.Projection ID_PROJECTION = StructuredQuery.Projection.newBuilder()
-			.addFields(StructuredQuery.FieldReference.newBuilder().setFieldPath(NAME_FIELD).build())
-			.build();
+	private final String nameField = FieldPath.documentId().toString();
```
meltsufin, line 73: Why not `static`?

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
```diff
@@ -62,61 +65,75 @@
...
-	private final FirestoreStub firestore;
+	private final FirestoreStub firestoreStub;
 
 	private final String parent;
 
 	private final String databasePath;
 
 	private final FirestoreMappingContext mappingContext;
 
+	private final Firestore firestore;
```
meltsufin, line 89: Why do we need both `Firestore` and `FirestoreStub`?

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
```diff
@@ -62,61 +65,75 @@
...
 	/**
 	 * Constructor for FirestoreTemplate.
-	 * @param firestore Firestore gRPC stub
+	 *
+	 * @param firestoreStub Firestore gRPC stub
 	 * @param parent the parent resource. For example:
 	 *     projects/{project_id}/databases/{database_id}/documents
 	 * @param classMapper a {@link FirestoreClassMapper} used for conversion
 	 * @param mappingContext Mapping Context
+	 * @param firestore Firestore service
```
meltsufin, line 105: Note that this is a breaking change on a public API.

meltsufin (commented):
You started with a complex module right off the bat!
Are there any documentation updates needed?

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
elefeint: This was one of the experiments we did to determine feasibility. I think we can do without the extra client library object (`Firestore`) now.
`FirestoreStub` is used because the template implementation historically bypasses the client library and instead relies directly on gRPC layer. After our experiences with Spanner, I'd much rather rely on client library -- something to consider for the future, but far out of scope for this PR.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
```diff
@@ -443,4 +460,8 @@ private Object getIdValue(Object entity) {
 	public FirestoreClassMapper getClassMapper() {
 		return this.classMapper;
 	}
+
+	public DocumentReference getDocumentReference(String collection, String documentId) {
+		return this.firestore.collection(collection).document(documentId);
+	}
```
elefeint, line 466: I believe we can do without this method now. Once you remove it, the `Firestore` client library object can go away.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -216,9 +220,19 @@ public Object execute(Object[] parameters) {
 							"Too few parameters are provided for query method: " + getQueryMethod().getName());
 				}
 				Object value = it.next();
-				filter.getFieldFilterBuilder().setField(fieldReference)
-						.setOp(getOperator(part, value))
-						.setValue(this.classMapper.toFirestoreValue(value));
+				if (fieldReference.getFieldPath().equals(FieldPath.documentId().toString())) {
+					FirestoreTemplate firestoreTemplate = (FirestoreTemplate) this.reactiveOperations;
+					DocumentReference documentReference = firestoreTemplate.getDocumentReference(persistentEntity.collectionName(), (String) value);
```
elefeint, line 225: This field will be the same as the part we are working with in this loop iteration. So you can use `value` instead of `documentReference` when looking up the operation a couple of lines down.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -216,9 +220,19 @@ public Object execute(Object[] parameters) {
...
+					filter.getFieldFilterBuilder().setField(fieldReference)
+							.setOp(getOperator(part, documentReference))
+							.setValue(Value.newBuilder().setReferenceValue(
+									firestoreTemplate.buildResourceName(persistentEntity, documentReference.getId())));
```
elefeint, line 229: Here again, we know what the *value* of the `@DocumentId` field is without having to look it up from document reference. It's the `value` local variable!
Now, the local variable is an `Object`, while you need a `String`. But per our [documentation](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/docs/src/main/asciidoc/firestore.adoc#object-mapping), `@DocumentId` can only be applied to a `String` field since that's what Firestore uses for document IDs.

elefeint (changes requested):
It's awesome that this works! The next challenge is getting `Firestore` object out of `FirestoreTemplate`, like @meltsufin noted. You should be able to do it, but it will require a change in `createBuilderWithFilter()` implementation. Once you do that work, there might be an opportunity for a small refactoring to reduce emerging duplication of code in if/else clause. I love seeing opportunities like these pop up once the large refactorings are completed.

I think the renaming of `firestore` to `firestoreStub` is a good one, since it makes the actual object's nature much clearer. However, it's a personal style preference whether to include it in the same PR as functionality changes -- it's up to you whether you want to keep it in this PR or remove it to make the current PR even smaller. Some people like to clean up small things like these as they go, and some prefer tight, targeted pull requests that only achieve the task at hand.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/repository/query/PartTreeFirestoreQuery.java
mpeddada1: That is a great point! I cleaned this up a bit to use `value` directly.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
mpeddada1: Thanks for catching this. Sorry, I had left some experimental code in. After applying the changes @elefeint described [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/506#pullrequestreview-687777943), we no longer need to use the `Firestore` object.

mpeddada1:
> I think the renaming of `firestore` to `firestoreStub` is a good one, since it makes the actual object's nature much clearer.
Agreed! I kept this change in this PR since it is a minor refactoring. However, as you mentioned in your comment, if the refactoring draws the reader's attention away from the purpose of the PR then having a follow-up PR is definitely a great idea.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
mpeddada1: Done, thank you

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
mpeddada1: Ah `static` makes sense. Also reverted some of the unnecessary changes I made to this file.

mpeddada1:
> Are there any documentation updates needed?

That's a good question. We may want to update the CHANGELOG for the next release?

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
```diff
@@ -427,7 +434,7 @@ private void doIfTransaction(Context ctx, Consumer<ReactiveFirestoreResourceHold
 		return buildResourceName(persistentEntity, idVal.toString());
 	}
 
-	private String buildResourceName(FirestorePersistentEntity<?> persistentEntity, String s) {
+	public String buildResourceName(FirestorePersistentEntity<?> persistentEntity, String s) {
```
meltsufin, line 437: If we make this public, it should have a javadoc.

meltsufin (approved):
Just one comment on the public method javadoc.
Adding this to changelog should be sufficient I guess.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/mapping/FirestorePersistentPropertyImpl.java
```diff
@@ -61,6 +61,9 @@ public boolean isIdProperty() {
 	}
 
 	public String getFieldName() {
+		if (isIdProperty()) {
+			return FieldPath.documentId().toString();
```
elefeint, line 65: I'd rather make `NAME_FIELD` constant public and reuse it here (yep, it will need its own Javadoc then). No reason do do this dynamically for every field.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -216,9 +219,18 @@ public Object execute(Object[] parameters) {
 							"Too few parameters are provided for query method: " + getQueryMethod().getName());
 				}
 				Object value = it.next();
-				filter.getFieldFilterBuilder().setField(fieldReference)
-						.setOp(getOperator(part, value))
-						.setValue(this.classMapper.toFirestoreValue(value));
+				StructuredQuery.FieldFilter.Builder queryFilterBuilder = filter.getFieldFilterBuilder()
+						.setField(fieldReference).setOp(getOperator(part, value));
+				if (fieldReference.getFieldPath().equals(FieldPath.documentId().toString())) {
```
elefeint, line 224: Same here, reuse `NAME_FIELD`

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -216,9 +219,18 @@ public Object execute(Object[] parameters) {
...
+					FirestoreTemplate firestoreTemplate = (FirestoreTemplate) this.reactiveOperations;
```
elefeint, line 225: In real life, `reactiveOperations` will almost always be of type `FirestoreTemplate`. However, that's not guaranteed!
Also, for testing, if we want to mock `reactiveOperations`, this line will fail. 

I would instead suggest adding `buildResourceName()` as a valid interface method on `FirestoreReactiveOperations` (btw, then the javadoc will go on the interface method, and not on implementation). This will make it a proper part of the API, which we can then mock for testing or be required to implement for an alternate implementation.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -216,9 +219,18 @@ public Object execute(Object[] parameters) {
...
+					queryFilterBuilder
+							.setValue(Value.newBuilder().setReferenceValue(
+									firestoreTemplate.buildResourceName(persistentEntity, (String) value)));
```
elefeint, line 228: For future developers, I'd add a comment for why we can assume value is always a `String`.

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/entities/UserRepository.java
```diff
@@ -56,6 +58,8 @@
 
 	Flux<User> findByAgeAndPetsContains(Integer age, List<String> pets);
 
+	Flux<User> findByNameAndPetsContains(String name, List<String> pets);
```
elefeint, line 61: Great addition to testing, since that's what the original issue cited!

elefeint (changes requested):
This is looking great. Still, we can future-proof and help ourselves with future testing a bit more.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/FirestoreTemplate.java
mpeddada1: Good point, done.

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/repository/query/PartTreeFirestoreQuery.java
mpeddada1: > Also, for testing, if we want to mock reactiveOperations, this line will fail.

Ah I did not think of this, thank you!

spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/repository/query/PartTreeFirestoreQuery.java
mpeddada1: Done.

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/repository/query/FirestoreRepositoryTests.java
```diff
@@ -70,7 +71,7 @@ public void testSortQuery() {
 				Order.newBuilder()
 						.setDirection(Direction.ASCENDING)
 						.setField(
-								FieldReference.newBuilder().setFieldPath("name"))
+								FieldReference.newBuilder().setFieldPath(FieldPath.documentId().toString()))
```
elefeint, line 74: This is an interesting change. The former line used "name"; the current uses "__name__", so at first glance, it looks like a behavior change. But! Notice, this is a unit test (well, it's a unit test for a fairly large "unit" extending from "spring data parses repository method name" to "firestore template execute called"). So while this test "used to work", this same test would fail in an integration test setup. 

So this got me to wonder... Did our sort by ID property works at all? And when I wonder whether something works, I write an integration test :

_FirestoreRepositoryIntegrationTests_
```
	@Test
	public void testOrderByDocumentId() {
		User alice = new User("Alice", 99);
		User bob = new User("Bob", 99);
		User zelda = new User("Zelda", 99);
		this.userRepository.save(alice)
				.then(this.userRepository.save(bob))
				.then(this.userRepository.save(zelda))
				.block();

		StepVerifier.create(
				this.userRepository.findByAge(99, Sort.by(Order.desc("name"))).map(User::getName)
		).expectNext("Zelda", "Bob", "Alice")
				.verifyComplete();
	}
```

So, get this: the test fails on `main` branch, and passes on your pull request. Congratulations on fixing a bug we never knew we had!

elefeint (approved):
Excellent! By the way, you fixed *two* bugs, one of which we never knew about.

PR #516, opened 2021-07-01
Support OrderBy in method name when querying data in Firestore

### LINKED ISSUE #511: Was Firestore meant to support OrderBy clause in method name?
Step 1. Add `Flux<User> findByAge(Integer age, Sort sort);` to `UserRepository`, and the following test to `FirestoreRepositoryIntegrationTests` (it will pass after #506 is merged).

```
	@Test
	public void testOrderByDocumentId() {
		User alice = new User("Alice", 99);
		User bob = new User("Bob", 99);
		User zelda = new User("Zelda", 99);
		this.userRepository.save(alice)
				.then(this.userRepository.save(bob))
				.then(this.userRepository.save(zelda))
				.block();

		StepVerifier.create(
				this.userRepository.findByAge(99, Sort.by(Order.desc("name"))).map(User::getName)
		).expectNext("Zelda", "Bob", "Alice")
				.verifyComplete();
	}
```

Step 2. Add `Flux<User> findByAgeOrderByNameDesc(Integer age);` to `UserRepository`, and add a slightly different test:

```
	@Test
	public void testOrderByDocumentIdFromMethodName() {
		User alice = new User("Alice", 99);
		User bob = new User("Bob", 99);
		User zelda = new User("Zelda", 99);
		this.userRepository.save(alice)
				.then(this.userRepository.save(bob))
				.then(this.userRepository.save(zelda))
				.block();

		StepVerifier.create(
				this.userRepository.findByAgeOrderByNameDesc(99).map(User::getName)
		).expectNext("Zelda", "Bob", "Alice")
				.verifyComplete();
	}
```

That one fails. 

Part-tree query[ gets its sort information](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/a141159b394e4a7ca5614497903eba492864f0e1/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/spring/data/firestore/repository/query/PartTreeFirestoreQuery.java#L145-L146) from `ParameterAcessor`, which works if the `Sort` parameter is passed directly. The sort clause is present in the parsed parts, actually, so most likely, we need to implement "and if not present in explicit sort parameter, check if it was part of the method name" logic. Our Spanner integration does support methods with `orderBy` in them, so that might be an interesting place to start comparing. Or debug Spring Data Commons a bit and see where it's setting the sort options!

mpeddada1:
Fixes #511

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -9,6 +9,7 @@ For a detailed view of what has changed, refer to the https://github.com/GoogleC
 
 === Firestore
 * Fixed: Unable to query by document ID. (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/506[#506])
+* Added support for `OrderBy` clause in method name. (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/516[#516])
 
 == 2.0.3 (2021-06-08)
 
--- a/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/entities/UserRepository.java
+++ b/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/entities/UserRepository.java
@@ -38,11 +38,11 @@ public interface UserRepository extends FirestoreReactiveRepository<User> {
 
 	Flux<User> findByAge(Integer age, Sort sort);
 
+	Flux<User> findByAgeOrderByNameDesc(Integer age);
+
 	Flux<User> findByAgeNot(Integer age);
 
 	Flux<User> findByNameAndAge(String name, Integer age);
-
-	Flux<User> findByAgeOrderByNameDesc(Integer age);
 
 	Flux<User> findByHomeAddressCountry(String country);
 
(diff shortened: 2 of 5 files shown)


spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/entities/UserRepository.java
```diff
@@ -36,6 +36,10 @@
 
 	Flux<User> findByAge(Integer age);
 
+	Flux<User> findByAge(Integer age, Sort sort);
+
+	Flux<User> findByAgeOrderByNameDesc(Integer age);
```
meltsufin, line 41: Since `name` or document id is a somewhat special field in Firestore, can we also add a test that sorts by a regular field like `age`? Is something like `findAllOrderByAge()` possible?

CHANGELOG.adoc
```diff
@@ -9,6 +9,7 @@ For a detailed view of what has changed, refer to the https://github.com/GoogleC
 
 === Firestore
 * Fixed: Unable to query by document ID. (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/506[#506])
+* Added support for `OrderBy` clause in method name. (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/516[#516])
```
meltsufin, line 12: Love the changelog update as part of the PR!

elefeint (approved):
+1 to adding `findAllOrderByAge` test.

meltsufin (approved):
LGTM!

elefeint (approved):
Looks great!

Issue #520, opened 2021-07-04
SpannerTemplate should add query execution time in ApplicationEvents


pravsingh:
**Is your feature request related to a problem? Please describe.**
I'm capturing the ApplicationEvent. It has everything I need however missing the actual query execution times. Having this valuable information will help us monitor the queries better in our system.

**Describe the solution you'd like**
We are already capturing the timing in lower level methods e.g. executeRead. We could just pass it to the ApplicationEvent classes.

**Describe alternatives you've considered**
We tried abstracting out SpannerTemplate and measure the timings at this level however its another indirection which we would like to avoid and instead use the SpannerTemplate.

**Additional context**
N/A

**_In case you need developers, I can contribute as well._**

mpeddada1:
Hi @pravsingh, thanks for filing this feature request! To clarify, are you aiming to store the execution time in an event (presumably an "after-{action}" event)?

pravsingh:
yes.

mpeddada1:
We would be happy to review your PR if you want to make a contribution!

elefeint:
Closing the issue since nobody else expressed interest in the timing feature being added to Spanner events, and measuring timing before/after specific `PubSubTemplate` calls will result in essentially the same timing. It might make more sense to expose built-in Spanner query statistics (#598), but nobody seemed interested in that either.

PR #522, opened 2021-07-08
Emit query execution time in AfterExecution application events

pravsingh:
fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/520

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/SpannerTemplate.java
@@ -143,10 +143,11 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 	public long executeDmlStatement(Statement statement) {
 		Assert.notNull(statement, "A non-null statement is required.");
 		maybeEmitEvent(new BeforeExecuteDmlEvent(statement));
+		long queryStartTime = System.currentTimeMillis();
 		long rowsAffected = doWithOrWithoutTransactionContext(x -> x.executeUpdate(statement),
 				() -> this.databaseClientProvider.get().readWriteTransaction()
 						.run(transactionContext -> transactionContext.executeUpdate(statement)));
-		maybeEmitEvent(new AfterExecuteDmlEvent(statement, rowsAffected));
+		maybeEmitEvent(new AfterExecuteDmlEvent(statement, rowsAffected, queryStartTime));
 		return rowsAffected;
 	}
 
@@ -154,10 +155,11 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 	public long executePartitionedDmlStatement(Statement statement) {
 		Assert.notNull(statement, "A non-null statement is required.");
 		maybeEmitEvent(new BeforeExecuteDmlEvent(statement));
+		long queryStartTime = System.currentTimeMillis();
 		long rowsAffected = doWithOrWithoutTransactionContext(x -> {
 			throw new SpannerDataException("Cannot execute partitioned DML in a transaction.");
 		}, () -> this.databaseClientProvider.get().executePartitionedUpdate(statement));
-		maybeEmitEvent(new AfterExecuteDmlEvent(statement, rowsAffected));
+		maybeEmitEvent(new AfterExecuteDmlEvent(statement, rowsAffected, queryStartTime));
 		return rowsAffected;
 	}
 
@@ -174,10 +176,10 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 				this.mappingContext.getPersistentEntityOrFail(entityClass);
 
 		KeySet keys = KeySet.singleKey(key);
-
+		long queryStartTime = System.currentTimeMillis();
 		try (ResultSet resultSet = executeRead(persistentEntity.tableName(), keys,
 				Collections.singleton(persistentEntity.getPrimaryKeyColumnName()), null)) {
-			maybeEmitEvent(new AfterReadEvent(Collections.emptyList(), keys, null));
+			maybeEmitEvent(new AfterReadEvent(Collections.emptyList(), keys, null, queryStartTime));
 			return resultSet.next();
 		}
 	}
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/event/AfterDeleteEvent.java
```diff
@@ -39,7 +41,16 @@
 	 * @param targetType the target entity type that needs to be deleted. This may be
 	 *     {@code null} depending on the
 	 */
-	public AfterDeleteEvent(List<Mutation> source, Iterable targetEntities, KeySet targetKeys, Class targetType) {
+	public AfterDeleteEvent(List<Mutation> source, Iterable targetEntities, KeySet targetKeys, Class targetType, long queryStartTime) {
```
meltsufin, line 44: Note that this is a breaking change. Maybe add a second constructor instead?

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/SpannerTemplate.java
```diff
@@ -143,21 +143,23 @@ public SpannerEntityProcessor getSpannerEntityProcessor() {
 	public long executeDmlStatement(Statement statement) {
 		Assert.notNull(statement, "A non-null statement is required.");
 		maybeEmitEvent(new BeforeExecuteDmlEvent(statement));
+		long queryStartTime = System.currentTimeMillis();
```
elefeint, line 146: Because `System.currentTimeMillis()` has a runtime cost, it would be good to gate it on `eventPublisher` being present, so users not using eventing are not affected.

I wonder if it would make sense to have all the "After" events extend a new `SpannerAfterEvent` class that would allow setting/getting "before" timestamp. Then the individual "after" events' constructors don't need to change at all. `maybeEmitEvent` could both return and accept a timestamp, gating on the presence of `eventPublisher`, as it does now. It's conveniently a private method, so no breaking changes.

@meltsufin what do you think -- is it overengineering? I do want to make sure we don't grab system timestamp on every query for everyone, since Spanner usecases tend to be high volume.

elefeint (changes requested):

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/SpannerTemplate.java
meltsufin: Yeah that seems like a good idea!

pravsingh: Its good idea to have a common class providing the query execution time. However I see below class hierarchy which makes it bit hard to add SpannerAfterEvent:

```java
(Before/After)SaveEvent <- SaveEvent
After(Read/Query)Event <- LoadEvent
(Before/After)ExecuteDmlEvent <- ExecuteDmlEvent
(Before/After)DeleteEvent <- DeleteEvent
```

All the after event classes need timestamp just before the actual query execution begins and then pass it to`maybeEmitEvent`. If we fetch timestamp inside `maybeEmitEvent`, then after event's `getTimestamp()` and current system time will be same. One way to have it reliably done is to pass BeforeEvent object reference to AfterEvent. This way we always have related events tied together. It would be easy to get execution time as `getTimestamp() - before.Event.getTimestamp()` from inside AfterEvent object.

Let me know what you think.

pravsingh: pushed new commit to optimize on fetching current timestamp.

elefeint: You are right that the existing class hierarchy is getting in the way.
What if we reimagine "before timestamp" to mean "previous event's timestamp" instead? Then we are free to introduce a new class `SpannerTimestamp` at the very root of this project's event hierarchy. `SpannerTimestamp`'s constructor would pass the event payload (generic `Object`) up to its parent `ApplicationEvent`. It would contain a field for `previousTimestamp`, along with the corresponding getters and setters. "Before"-type events would also have this inherited field, but they don't need to use it -- the field can stay unset or be set to a null/zero/invalid. None of the leaf event constructors would need to change then -- `maybeEmitEvent` would set the previous timestamp and return the current one, just the way you are doing right now.

By the way, I think it's safe to use the primitive `long` type for all timestamps, and to define integer `0` as the absent value. In any other usecase, I would say that the wrapper type `Long` is more appropriate, but events are a special case -- we are only dealing with application events that happen in the future. There is no instance in which timestamp 0 (sometime in 1969) will be emitted.

elefeint: I'm sorry to waste your time writing several versions of the code. It's just that solving the problem in the fewest lines of code (especially duplicate code) possible makes the maintenance burden on our team much more manageable.
It would have been much better if we had avoided inheritance in favor of composition in the eventing code in the first place, but, well, hindsight is 20/20...

pravsingh: totally makes sense.
Its part of code evolution to streamline the code to adopt better design. Hopefully in the next major version upgrade, this can be made as a breaking change.

pravsingh: currently we are enhancing event to carry previousTimestamp however in future we may need to add extra metadata. Do you think we should rename `SpannerTimestamp` to something more generic e.g. `SpannerQueryData` which will have `previousTimestamp` for now and can be extended in future?
I'll make corresponding change and will adjust after the feedbacks.

elefeint: That's a great idea. `SpannerQueryData` works (or it could be even more generically `SpannerContext`). It could be a map of String to Object?
This will make it easier in the future to migrate away from the event hierarchy.

meltsufin:
@pravsingh Are you planning to continue working on the suggested refactoring?

elefeint:
Closing the PR since nobody else expressed interest in the timing feature being added to Spanner events, and measuring timing before/after specific `PubSubTemplate` calls will result in essentially the same timing. It might make more sense to expose built-in Spanner query statistics (#598), but nobody seemed interested in that either.

Sorry for the back-and-forth with the PR review. We've all been on the receiving end of unreasonable-seeming review comments, and it can be very frustrating.

PR #525, opened 2021-07-09
Revert to published cfenv version

elefeint:
`java-cfenv:2.4.0` is not released yet; it most likely made its way to spring release repo [as an experiment](https://github.com/pivotal-cf/java-cfenv/issues/155#issuecomment-877498452), and we got lucky and picked it up.


mpeddada1 (approved):
My bad, I didn't realize this was an experiment. thank you for catching this!

elefeint:
There is no way you could possibly have known that!

PR #548, opened 2021-07-23
Pub/Sub Trace Instrumentation

meltsufin:
Currently, traces are not automatically propagated through Pub/Sub. This PR implements the propagation. Parts of the implementation should ultimately go into https://github.com/openzipkin/brave, https://github.com/openzipkin/zipkin-gcp, and some parts should remain in this repo.

For now, the idea is to keep the feature off by default and enable it using `spring.cloud.gcp.trace.pubsub.enabled=true`.

Towards: https://github.com/openzipkin/zipkin-gcp/issues/190
Fixes: https://github.com/spring-cloud/spring-cloud-gcp/issues/2671


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubTracing.java
```diff
@@ -0,0 +1,316 @@
...
+	/** When an upstream context was not present, lookup keys are unlikely added */
+	static void addTags(PubsubMessage.Builder message, SpanCustomizer result) {
+
+		// TODO: anything to do here?
+	}
+
+
+	PullResponse tracePullResponse(PullResponse delegate, String subscriptionName) {
+
+		if (delegate.getReceivedMessagesCount() == 0 || tracing.isNoop()) {
```
elefeint, line 197: Could "no messages received" be a useful trace signal, too?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubTracing.java
```diff
@@ -0,0 +1,316 @@
...
+			return delegate;
+		}
+
+		Span[] batchSpan = null;
+		if (singleRootSpanOnReceiveBatch) {
+			batchSpan = new Span[1];
+		}
```
elefeint, line 204: Leave a comment that there is no NPE because `singleRootSpanOnReceiveBatch` is checked everywhere? Or even use `batchSpan`'s non-nullity instead of `singleRootSpanOnReceiveBatch` as a signal thatsspans should be grouped.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubTracing.java
```diff
@@ -0,0 +1,316 @@
...
+
+	final TraceContextOrSamplingFlags emptyExtraction;
+
+	final SamplerFunction<MessagingRequest> producerSampler;
+
+	final SamplerFunction<MessagingRequest> consumerSampler;
+
+	final String remoteServiceName;
+
+	final boolean singleRootSpanOnReceiveBatch;
```
elefeint, line 79: How about `groupReceivedMessageSpans` or `groupPulledMessageSpans` or something? I was confused by the purpose at first.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubTracing.java
```diff
@@ -0,0 +1,316 @@
...
+		else { // we extracted request-scoped data, so cannot share a consumer span.
+			Span span = nextMessagingSpan(consumerSampler, request, extracted);
+			if (!span.isNoop()) {
+				setConsumerSpan(span, subscriptionName);
+
+				// incur timestamp overhead only once
+				if (timestamp == 0L) {
+					timestamp = tracing.clock(span.context()).currentTimeMicroseconds();
+				}
+				span.start(timestamp);
```
elefeint, line 256: this part can probably become a helper method; it's similar to above.

On the other hand, if we are putting these classes back to brave, it might help for maintenance there if the code looks similar to the rest of brave..

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingMessageReceiver.java
```diff
@@ -0,0 +1,88 @@
...
+		PubsubMessage.Builder messageBuilder = pubsubMessage.toBuilder();
+		PubSubConsumerRequest request = new PubSubConsumerRequest(messageBuilder, subscriptionName);
+		TraceContextOrSamplingFlags extracted =
+				pubSubTracing.extractAndClearTraceIdHeaders(pubSubTracing.consumerExtractor, request, messageBuilder);
+
+		Span consumerSpan = pubSubTracing.nextMessagingSpan(pubSubTracing.consumerSampler, request, extracted);
+		Span listenerSpan = pubSubTracing.tracer.newChild(consumerSpan.context());
+
+		if (!consumerSpan.isNoop()) {
+			consumerSpan.name("next-message").kind(CONSUMER); // TODO: do we need this consumer span at all?
```
elefeint, line 53: When writing documentation, can we include screenshots of Cloud Trace to show what users can expect to see? This will help with the "group received spans vs not group spans" documentation a lot.

Also will help us visualize what we want the outcome to be.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingMessageReceiver.java
```diff
@@ -0,0 +1,88 @@
...
+			consumerSpan.tag("pubsub.subscription", subscriptionName); // TODO: shouldn't have to tag manually since it's in PubSubConsumerRequest
```
elefeint, line 54: Does it show up in trace visuals if we skip this tag?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingMessageReceiver.java
```diff
@@ -0,0 +1,88 @@
...
+			if (pubSubTracing.remoteServiceName != null) {
+				consumerSpan.remoteServiceName(pubSubTracing.remoteServiceName);
+			}
+
+			// incur timestamp overhead only once
+			long timestamp = pubSubTracing.tracing.clock(consumerSpan.context()).currentTimeMicroseconds();
+			consumerSpan.start(timestamp);
+			long consumerFinish = timestamp + 1L; // save a clock reading
```
elefeint, line 62: I am confused by why start and finish are basically the same. Should the start of `consumerSpan` be before we run the user callback `delegate.receiveMessage(...)` and the end after?

This seems to be what `listenerSpan` does, but then why do we need `consumerSpan` at all?

Update: haha, played myself by answering your TODO question. I still want to know what the visual looks like with and without `consumerSpan`, but IMHO we don't need the fake-ish `consumerSpan`.

Update2: found an example of what this looks like in the sample (great sample, btw). Yeah, "next-message" is completely redundant.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingMessageReceiver.java
```diff
@@ -0,0 +1,88 @@
...
+
+			// not using scoped span as we want to start with a pre-configured time
+			listenerSpan.name("on-message").start(consumerFinish);
+		}
+
+		Tracer.SpanInScope ws = pubSubTracing.tracer.withSpanInScope(listenerSpan);
+		Throwable error = null;
+		try {
+			// pass instrumented message the actual receiver
+			delegate.receiveMessage(messageBuilder.build(), ackReplyConsumer);
```
elefeint, line 73: However, instrumenting `ackReplyConsumer` may theoretically be useful. Meh, opportunity for future enhancement.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingPublisher.java
```diff
@@ -0,0 +1,85 @@
...
+			span = pubSubTracing.nextMessagingSpan(pubSubTracing.producerSampler, request, extracted);
+		}
+		else { // If we have a span in scope assume headers were cleared before
+			span = pubSubTracing.tracer.newChild(maybeParent);
+		}
+
+		if (!span.isNoop()) {
+			span.kind(PRODUCER).name("publish");
+			if (topic != null) {
+				span.tag("pubsub.topic", topic); // TODO: shouldn't have to do this manually because topic is in PubSubProducerRequest
```
elefeint, line 71: Does this empirically work without this tag?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingPublisher.java
```diff
@@ -0,0 +1,85 @@
...
+			}
+			if (pubSubTracing.remoteServiceName != null) {
+				span.remoteServiceName(pubSubTracing.remoteServiceName);
+			}
+			// incur timestamp overhead only once
+			long timestamp = pubSubTracing.tracing.clock(span.context()).currentTimeMicroseconds();
+			// the span is just an instant, since we can't track how long it takes to publish and carry that forward
```
elefeint, line 78: What if we add a listener on the publish future?

spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/src/main/java/com/example/Application.java
```diff
@@ -63,4 +82,43 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
...
+	}
+
+	@Bean
+	public PubSubInboundChannelAdapter messageChannelAdapter(
+			@Qualifier("pubsubInputChannel") MessageChannel inputChannel,
+			PubSubTemplate pubSubTemplate) {
+		PubSubInboundChannelAdapter adapter =
+				new PubSubInboundChannelAdapter(pubSubTemplate, "traceSubscription");
+		adapter.setOutputChannel(inputChannel);
+		adapter.setAckMode(AckMode.MANUAL);
```
elefeint, line 113: Might as well make it auto-ack; not testing that here.

elefeint (changes requested):
Tons of work; looks amazing. Great sample.

I only have one substantial request, about publisher caching.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/TracingPublisherFactory.java
```diff
@@ -0,0 +1,38 @@
...
+	private final PublisherFactory publisherFactory;
+
+	TracingPublisherFactory(PubSubTracing pubSubTracing, PublisherFactory publisherFactory) {
+		this.pubSubTracing = pubSubTracing;
+		this.publisherFactory = publisherFactory;
+	}
+
+	@Override
+	public PublisherInterface createPublisher(String topic) {
+		return pubSubTracing.publisher(publisherFactory.createPublisher(topic),  topic);
```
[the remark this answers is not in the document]
meltsufin: Great point.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubProducerRequest.java
```diff
@@ -0,0 +1,104 @@
...
+				}
+
+				@Override
+				public String get(PubSubProducerRequest request, String name) {
+					return request.delegate.getAttributesOrDefault(name, null);
+				}
+
+				@Override
+				public String toString() {
+					return "PubsubMessage::getAttribute";
```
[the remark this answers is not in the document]
meltsufin: `PubSubMessage` is used in both send and receive directions. So, I guess this is accruate.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubProducerRequest.java
```diff
@@ -0,0 +1,104 @@
...
+				}
+
+				@Override
+				public void put(PubSubProducerRequest request, String name, String value) {
+					request.delegate.putAttributes(name, value);
+				}
+
+				@Override
+				public String toString() {
+					return "PubsubMessage.Builder::putAttributes";
```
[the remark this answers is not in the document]
meltsufin: `PubSubMessage` is used in both send and receive directions. So, I guess this is accurate.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubTracing.java
meltsufin: It's a good question. I don't know. Maybe it would add too much noise. I haven't seen other implementation do anything with an empty response. So, I was just following that pattern established in `KafkaTracing` and others.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubTracing.java
meltsufin: `batchSpan` is a local variable that we init to `null` above. So, I'm not sure how to avoid using `singleRootSpanOnReceiveBatch`.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubTracing.java
meltsufin: I was just trying to stay consistent with `KafkaTracing` and others in Brave.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubTracing.java
meltsufin: Yep, for now trying to stay as close as possible to Brave code.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingMessageReceiver.java
meltsufin: Documentation is definitely a big TODO here.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingMessageReceiver.java
meltsufin: It doesn't. I'll just review the comment.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingMessageReceiver.java
meltsufin: Again this is following the Brave examples. So, I'm a bit worried about breaking consistency with those instrumentations.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingPublisher.java
meltsufin: Doesn't work without it. So, I'll just remove the comment.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/TracingPublisher.java
meltsufin: That's an interesting idea. It should be possible. Adding a TODO.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/brave/PubSubTracing.java
elefeint: Right, I mean in places to which batchSpan is passed. But this is fine, too -- consistent with the other places where we follow Brave style.

elefeint (approved):
LGTM.

Integration test fails because the topic does not exist -- reuse `exampleTopic` or create one before test. After fixing this, merge away.

`NOT_FOUND: Resource not found (resource=traceTopic).`

meltsufin:
@elefeint Good catch on the topic/subscription missing! I added instructions in the sample readme to create the topic and subscription. I will test on the `test` branch before merging.

Issue #452, opened 2021-04-29
Using Stream as result of a query method failed due to expecting a singular result


nbali:
Despite what https://www.baeldung.com/spring-data-java-8 says, when I created a `Stream<Foo>` on a `org.springframework.cloud.gcp.data.datastore.repository.DatastoreRepository` implementation, I received this:
```
org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataException: The query method returns a singular object but the query returned more than one result.
  at org.springframework.cloud.gcp.data.datastore.repository.query.GqlDatastoreQuery.convertSingularResult (GqlDatastoreQuery.java:227)
  at org.springframework.cloud.gcp.data.datastore.repository.query.GqlDatastoreQuery.execute (GqlDatastoreQuery.java:156)
  at org.springframework.data.repository.core.support.RepositoryMethodInvoker.doInvoke (RepositoryMethodInvoker.java:137)
  at org.springframework.data.repository.core.support.RepositoryMethodInvoker.invoke (RepositoryMethodInvoker.java:121)
  at org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor.doInvoke (QueryExecutorMethodInterceptor.java:152)
  at org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor.invoke (QueryExecutorMethodInterceptor.java:131)
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed (ReflectiveMethodInvocation.java:186)
  at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke (ExposeInvocationInterceptor.java:97)
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed (ReflectiveMethodInvocation.java:186)
  at org.springframework.aop.framework.JdkDynamicAopProxy.invoke (JdkDynamicAopProxy.java:215)
  at com.sun.proxy.$Proxy104.getValidCarrierDirections (Unknown Source)
```

meltsufin:
Does `List<Foo>` work?

nbali:
Both `List<Foo>` and `Iterable<Foo>` seems to work.

PR #551, opened 2021-07-27
Support Stream as return type to datastore queries

zhumin8:
This is fix for #452. 
Fixed code in `GqlDatastoreQuery` and `PartTreeDatastoreQuery` to support for Stream as a return type and added corresponding unit tests and integration tests.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
@@ -444,7 +444,13 @@ public class PartTreeDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 
 			structuredQueryBuilder.setKind(PartTreeDatastoreQuery.this.datastorePersistentEntity.kindName());
 
-			singularResult = (!isCountingQuery && collectionType == null && !isStreamQuery) && !PartTreeDatastoreQuery.this.tree.isDelete();
+			if (isCountingQuery || collectionType != null || isStreamQuery
+					|| PartTreeDatastoreQuery.this.tree.isDelete()) {
+				singularResult = false;
+			}
+			else {
+				singularResult = true;
+			}
 		}
 
 		boolean isReturnedTypeIsNumber() {
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
@@ -186,13 +186,18 @@ public class PartTreeDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 	}
 
 	private Object runQuery(Object[] parameters, Class returnedElementType, Class<?> collectionType, boolean requiresCount) {
-		ExecutionOptions options = new ExecutionOptions(returnedElementType, collectionType, requiresCount);
+		ExecutionOptions options = new ExecutionOptions(returnedElementType, collectionType, requiresCount,
+				getQueryMethod().isStreamQuery());
 
 		DatastoreResultsIterable rawResults = getDatastoreOperations()
 				.queryKeysOrEntities(
 						applyQueryBody(parameters, options.getQueryBuilder(),
 								requiresCount, options.isSingularResult(), null),
 						this.entityType);
+
+		if (getQueryMethod().isStreamQuery()) {
+			return StreamSupport.stream(rawResults.spliterator(), false);
+		}
 
 		Object result = StreamSupport.stream(rawResults.spliterator(), false)
 				.map(options.isReturnedTypeIsNumber() ? Function.identity() : this::processRawObjectForProjection)
@@ -422,7 +427,7 @@ public class PartTreeDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 
 		private boolean singularResult;
 
-		ExecutionOptions(Class returnedElementType, Class<?> collectionType, boolean requiresCount) {
+		ExecutionOptions(Class returnedElementType, Class<?> collectionType, boolean requiresCount, boolean isStreamQuery) {
 
 			returnedTypeIsNumber = Number.class.isAssignableFrom(returnedElementType)
 					|| returnedElementType == int.class || returnedElementType == long.class;
@@ -439,7 +444,7 @@ public class PartTreeDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 
 			structuredQueryBuilder.setKind(PartTreeDatastoreQuery.this.datastorePersistentEntity.kindName());
 
-			singularResult = (!isCountingQuery && collectionType == null) && !PartTreeDatastoreQuery.this.tree.isDelete();
+			singularResult = (!isCountingQuery && collectionType == null && !isStreamQuery) && !PartTreeDatastoreQuery.this.tree.isDelete();
 		}
 
 		boolean isReturnedTypeIsNumber() {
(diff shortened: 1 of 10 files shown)


spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/TestEntityRepository.java
```diff
@@ -134,6 +135,11 @@
 
 	Optional<TestEntity> findFirstByColor(String color);
 
+	Stream<TestEntity> findStreamByColor(String color);
```
meltsufin, line 138: I didn't that `findStream...` works?
Would this also work:
```
Stream<TestEntity> findByColor(String color);
```

meltsufin (changes requested):
Looks good to me. I would just encourage to also update Datastore documentation and one of the samples.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/TestEntityRepository.java
zhumin8: So according to [this post](https://stackoverflow.com/questions/37253571/spring-data-jpa-difference-between-findby-findallby) anything between `find` and `By` is ignored by Spring Data, which is neat when we have two same query with same parameters.
But in this case, `Stream<TestEntity> findByColor(String color);` would also work since the other `findByColor` in line 134 has different signature.

zhumin8:
Good idea. I'll look into updating Datastore documentation and sample.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/TestEntityRepository.java
meltsufin: Very cool! I didn't know that.

spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -439,7 +444,7 @@ public Cursor getCursor() {
 
 			structuredQueryBuilder.setKind(PartTreeDatastoreQuery.this.datastorePersistentEntity.kindName());
 
-			singularResult = (!isCountingQuery && collectionType == null) && !PartTreeDatastoreQuery.this.tree.isDelete();
+			singularResult = (!isCountingQuery && collectionType == null && !isStreamQuery) && !PartTreeDatastoreQuery.this.tree.isDelete();
```
elefeint, line 447: This boolean expression is getting too complicated. Since we are touching this line, could you refactor it into an if/else statement, with a view for reducing the negations ("if it's a counting query or a stream query, then singlarResult = false") is easier on the mind than "if it's not a counting query and it's not a stream query, singlarResult = true".

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/TestEntityRepository.java
```diff
@@ -134,6 +135,11 @@
...
+
+	@Query("select * from  test_entities_ci where color = @color")
+	Stream<TestEntity> findByColorStream(@Param("color") String color);
```
elefeint, line 141: The difference between `findStreamByColor` and `findByColorStream` is very subtle. Could you rename `findByColorStream` to something like `findGqlStreamByColor`? 

And possibly rename `findStreamByColor` to `findPartTreeStreamByColor`, so the targeted functionality is very clear.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/repository/query/GqlDatastoreQueryTests.java
```diff
@@ -447,6 +448,39 @@ public void pageableTestPageCursor() {
...
+	public void streamResultTest() {
+		Mockito.<Class>when(this.queryMethod.getReturnedObjectType())
+				.thenReturn(Trade.class);
+		Parameters parameters = mock(Parameters.class);
+
+		when(this.queryMethod.getParameters())
+				.thenReturn(parameters);
+		when(parameters.getNumberOfParameters())
+				.thenReturn(0);
+		when(this.queryMethod.isStreamQuery()).thenReturn(true);
```
elefeint, line 461: nit: some of these have `.thenReturn` on a new line, some on the same line. I don't have a personal preference, as long as they all do the same thing within at least a method (well, unless a line is too long and one line does not fit, but in this case all statements are short).

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/repository/query/GqlDatastoreQueryTests.java
```diff
@@ -447,6 +448,39 @@ public void pageableTestPageCursor() {
...
+
+		Trade tradeA = new Trade();
+		tradeA.id = "a";
+		Trade tradeB = new Trade();
+		tradeB.id = "b";
+		doAnswer(invocation -> {
+			GqlQuery statement = invocation.getArgument(0);
+			assertThat(statement.getQueryString()).isEqualTo("unusedGqlString");
```
elefeint, line 469: I vaguely remember something about asserts inside of `when()`/`doAnswer()` not propagating up to the test level. Could you see if replacing "unusedGqlString" with "something to trigger assert exception" makes the test fail? If it does not fail, then we might as well remove this assert line, since it will never work.

(but my recollection might be wrong! So if changing the string makes the test fail properly, then definitely leave it alone).

elefeint (changes requested):
The code looks great! I only have minor feedback.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/repository/query/GqlDatastoreQueryTests.java
zhumin8: Got it. Making them all in one line since they are short in this case.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/repository/query/GqlDatastoreQueryTests.java
zhumin8: Tested. It actually did fail properly!

spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQuery.java
zhumin8: Good point, I'll take the chance to make it easier to read.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/TestEntityRepository.java
zhumin8: hmm, do you think I should also make a comment here say these method are to test Stream as return type for GQL query and PartTree query respectively? Or is it too much?

meltsufin (approved):
Nice work!

elefeint (approved):

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/TestEntityRepository.java
elefeint: no, I think the descriptive naming you have right now is great.

PR #553, opened 2021-07-29
fix broken sample integration test in datastore.

zhumin8:
To fix `DatastoreSampleApplicationIntegrationTests` broke by #551.


elefeint (approved):
Another way to resolve the issue would be to disambiguate the URL on Spring Data Rest level:
```
	@RestResource(path="singersStream")
	Stream<Singer> findSingersByLastName(String name);
```

But I like your solution better -- it's more descriptive, and does not pollute the repository with Spring Data Rest annotations.

PR #561, opened 2021-08-04
Move class that defines publisher and subscriber properties to pubsub module

mpeddada1:
Part 1 of implementation for subscription-specific configurations. 
This PR makes Pub/Sub properties accessible to the pubsub module.


spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
```diff
@@ -0,0 +1,458 @@
...
+import com.google.api.gax.batching.FlowController.LimitExceededBehavior;
+
+/**
+ * Properties for Publisher or Subscriber specific configurations.
+ *
+ * @author João André Martins
+ * @author Mike Eltsufin
+ * @author Chengyuan Zhao
+ */
+public class PubSubEventSpecificProperties {
```
meltsufin, line 28: What does "event specific" mean? Does it make sense to just call it `PubSubConfiguration`?

meltsufin (commented):
This is exactly what I had in mind to resolve that circular dependency issue.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
elefeint: +1 for simpler name.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
```diff
@@ -0,0 +1,458 @@
...
+
+	/**
+	 * Contains settings specific to the subscriber factory.
+	 */
+	private final Subscriber subscriber = new Subscriber();
+
+	/**
+	 * Contains settings specific to the publisher factory.
+	 */
+	private final Publisher publisher = new Publisher();
```
elefeint, line 38: I wonder if while we are doing this refactoring, it would make sense to split subscriber properties and publisher properties up. Our actual application classes always either need publisher properties or subscriber properties, never both.
With the split, we'd still keep these fields in autoconfiguration's `GcpPubSubProperties`, but autoconfiguration's `Subscriber` would instead be `Subscriber extends SubscriberProperties`, with `SubscriberProperties` coming from the pubsub module. 
As a side effect, this would help with:
* the name confusion, so instead of `PubSubEventSpecificProperties` vs `GcpPubSubProperties`, you'd have `GcpPubSubProperties` containing two specific sets of configuration from pubsub module.
* backwards compatibility -- if our users injected our configuration and had code like `GcpPubSubProperties.Publisher props = /* injected object */`, keeping a shell class in autoconfiguration might help. On the other hand, I am not keen on replicating each subclass (like flow control) in both modules...

@meltsufin WDYT?

elefeint (commented):
I'd be okay with this PR as is, but there might be an opportunity to separate subscriber and publisher properties here. I want a second/third opinion on this :)

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
mpeddada1: I really like the idea of having a separate class each for Subscriber properties and Publisher properties. I was initially thinking of splitting them for the purpose of making `PubSubEventSpecificProperties` shorter but your suggestion with using `SubscriberProperties` and `PublisherProperties` in `GcpPubSubProperties` is more effective. This way, we can just pass in subscriber-specific or publisher-specific properties to the SubscriberFactory and PublisherFactory as you mentioned in [#562(comment)](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/562#discussion_r685291844). I can try it out and see how it looks.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
mpeddada1: That's a much better name, thank you.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
mpeddada1: Unfortunately, I just realized that it isn't possible to use `SubscriberProperties` and `PublisherProperties` directly in the factory classes. We created `PubSubEventSpecificProperties` (now `PubSubConfiguration`) for the purpose of holding a map of <String, Subscriber>. This map is later used in `DefaultSubscriberFactory` to get the appropriate Subscriber() associated with the subscription name which we only have access to in the `createSubscriber(String subscriptionName,..)` method. Getting the Subscriber from the map also involves some extra logic - If the subscription name is present in the map then we return the Subscriber with the subscription-specific properties set and if the subscription name is not found then we return either global properties (if set) or default properties.

mpeddada1: Update: Making SubscriberProperties contain Map<String, Properties> and initializing SubscriberProperties sub = new SubscriberProperties() in GcpPubSubAutoconfiguration also doesn't work as we expected. There seem to be two problems:
1. The map properties don't get recognized when we do spring.cloud.gcp.pubsub.properties.[sub-name].subscriber.executor-threads or spring.cloud.gcp.pubsub.sub.properties.[sub-name].subscriber.executor-threads. Spring is able to recognize map properties directly but not when we include it within an intermediate class for some reason.
2. Relying on the globally set configurations when the subscription-specific ones are not present becomes difficult. In order to ensure that the current configuration setup (of doing spring.cloud.gcp.pubsub.subscriber.executor-threads) still works, we need to introduce aSubscriber subscriber variable in GcpPubSubAutoConfiguration. However, this value is not accessible to SubscriberProperties#getSubscriber(String name) due to the circular dependency problem with the auto-config and the pubsub module.

elefeint: Okay, this is too deep of a rabbit hole. I withdraw my suggestion! Sometimes making the code perfectly map to a mental model is not worth the work.

elefeint (approved):
Let's merge it!

PR #562, opened 2021-08-04
Enable subscription-specific configuration for the `executor-threads` property

mpeddada1:
Part 2. On top of #561

TODO: Still need to add a lot more tests. 

This PR enables setting subscription-specific configuration for `executor-threads`. 
Configuration before the change:
```
spring.cloud.gcp.pubsub.subscriber.executor-threads= 2 # Sets this configuration for all subscribers
```
Configuration after the change:
```
spring.cloud.gcp.pubsub.subscriber.executor-threads= 2 # Sets this configuration for all subscribers
spring.cloud.gcp.pubsub.properties.[sub-name].subscriber.executor-threads # Sets this configuration for subscriber associated with subscription named [sub-name]. This takes precedence over the global configuration.
```

**Note** that with this change, we will be deprecating the `pubsubSubscriberThreadPool` and `subscriberExecutorProvider` beans. We will be setting these properties directly through `PubSubEventSpecificProperties` which is passed in as a parameter to `DefaultSubscriberFactory's` constructor.


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -131,6 +131,7 @@ public ExecutorProvider publisherExecutorProvider(
 
 	@Bean
 	@ConditionalOnMissingBean(name = "pubsubSubscriberThreadPool")
+	@Deprecated
```
meltsufin, line 134: Typically we include a javadoc comment pointing to what is replacing it.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -96,6 +95,8 @@
 
 	private Executor asyncPullExecutor = Runnable::run;
 
+	private Map<String, SubscriberStub> collectionOfSubscriberStub = new HashMap<>();
```
meltsufin, line 98: naming suggestion: `subscriptionToSubscriberStubMap`.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -190,13 +191,13 @@ public Subscriber subscribe(String subscription,
 	 * Pulls messages synchronously, on demand, using the pull request in argument.
 	 *
 	 * @param pullRequest pull request containing the subscription name
+	 * @param subscriberStub subscriber stub
 	 * @return the list of {@link AcknowledgeablePubsubMessage} containing the ack ID, subscription
 	 * and acknowledger
 	 */
-	private List<AcknowledgeablePubsubMessage> pull(PullRequest pullRequest) {
+	private List<AcknowledgeablePubsubMessage> pull(PullRequest pullRequest, SubscriberStub subscriberStub) {
```
meltsufin, line 198: Why do you need to pass `subscriberStub` into the method? Doesn't `pullRequest` contain the subscription name which you can use to lookup the stub in the map?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -250,13 +251,18 @@ public void onSuccess(PullResponse pullResponse) {
 	@Override
 	public List<AcknowledgeablePubsubMessage> pull(
 			String subscription, Integer maxMessages, Boolean returnImmediately) {
+		SubscriberStub subscriberStub = this.subscriberFactory.createSubscriberStub(subscription);
```
meltsufin, line 254: You probably only want to create one if it's not already in the map.
I would just create a helper method like `getSubscriberStub(subscription)`.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
```diff
@@ -0,0 +1,490 @@
...
+/**
+ * Properties for Publisher or Subscriber specific configurations.
+ *
+ * @author João André Martins
+ * @author Mike Eltsufin
+ * @author Chengyuan Zhao
+ */
+public class PubSubEventSpecificProperties {
+
+	private final Map<String, Properties> properties = new HashMap<>();
```
meltsufin, line 33: I think in Spring Cloud Stream they use the term "bindings" rather than "properties", but I'm not sure if has the same meaning.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
```diff
@@ -0,0 +1,490 @@
...
+		if (!this.properties.containsKey(name)) {
+			Properties property = new Properties();
+			property.setSubscriber(this.subscriber);
+			this.properties.put(name, property);
+		}
+
+		return this.properties.get(name).getSubscriber();
+	}
+
+	public static class Properties {
```
meltsufin, line 67: I think calling this class "Properties" could be too ambiguous. "PubSubProperties" or "PubSubConfiguration"? Also, I think it could be a standalone class.

meltsufin (commented):
Just some preliminary comments...
I know it's a draft. So, apologies in advance for comments that you were planning to address as you continue to refine the PR. :)

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
```diff
@@ -302,6 +302,18 @@ public void testSubscribeAndConvert_AndManualNack()
 		assertThat(testListenableFutureCallback.getThrowable()).isNull();
 	}
 
+	@Test
+	public void destroyingBeanClosesSubscriberStub() {
```
elefeint, line 306: Great test -- now try verifying that the executor created in `getExecutorProvider.getExecutorProvider()` is shut down. I'd run an integration test with a breakpoint in whatever concrete subclass `ScheduledExecutorService.shutdown()` is returned from `FixedExecutorProvider.create(scheduler.getScheduledExecutor())`, just to see what happens in real life, before trying to make a test.

Note: this might not actually require a unit test, but trying to make one will be an interesting learning experience. So this is the kind of thing to spend < 2 hours on. But verifying that the executor is shut down in real life is important because otherwise we'll create a memory leak.

elefeint (changes requested):
Direction is good!

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
mpeddada1: Thanks for the suggestions! I agree that the naming is a little vague, I will clean this up. 

> Do we actually need the intermediate class `Properties`? It is just a container for a single `Subscriber` property bag. 

That is a great question! Having the ui be `spring.cloud.gcp.pubsub.*name*.subscriber.executor-threads` instead of having the intermediate `Properties` class would certainly be ideal. However, this didn't play out very well with Spring autoinjection. When we had the setup in the following manner the global configurations were working fine but for the subscription-specific properties, we were relying on the default values instead of the ones being set. The map properties weren't getting recognized.
```
        private final Map<String, Subscriber> properties = new HashMap<>();

	/**
	 * Contains settings specific to the subscriber factory.
	 */
	private final Subscriber subscriber = new Subscriber();
``` 
Looking at this [spring reference doc](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.typesafe-configuration-properties.merging-complex-types), we may still need an intermediate class?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
mpeddada1: Hm I think keeping these wouldn't make a difference since we fall back to the default value in `PubSubEventSpecificProperties` if the subscription-specific settings are not set:
```
public Subscriber getSubscriber(String name) {
		if (!this.properties.containsKey(name)) {
			Properties property = new Properties();
			property.setSubscriber(this.subscriber);
			this.properties.put(name, property);
		}

		return this.properties.get(name).getSubscriber();
	}
```

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
elefeint: Yep, you are right. Let's keep the `.properties.`

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplate.java
mpeddada1: Thank you! Done.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplate.java
mpeddada1: Great catch. Sorry, passing in `subscriberStub` is unnecessary here.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
mpeddada1: Update: After having discussed this offline, we decided to keep the beans but deprecate them and log a message informing the user to use `application.properties` when the bean is set. The subscription-specific properties option will not be available to those who directly use the beans.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -282,4 +304,18 @@ public SubscriberStub createSubscriberStub() {
...
 
+	/**
+	 * Creates {@link ExecutorProvider} given subscriber properties.
+	 * @param subscriber subscriber properties
+	 * @return executor provider
+	 */
+	private ExecutorProvider getExecutorProvider(PubSubEventSpecificProperties.Subscriber subscriber) {
+		ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
+		scheduler.setPoolSize(subscriber.getExecutorThreads());
+		scheduler.setThreadNamePrefix("gcp-pubsub-subscriber");
```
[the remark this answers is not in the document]
mpeddada1: We could make the distinction by appending the subscription-name after `gcp-pubsub-subscriber`?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubEventSpecificProperties.java
```diff
@@ -0,0 +1,490 @@
...
+ * @author Chengyuan Zhao
+ */
+public class PubSubEventSpecificProperties {
+
+	private final Map<String, Properties> properties = new HashMap<>();
+
+	/**
+	 * Contains settings specific to the subscriber factory.
+	 */
+	private final Subscriber subscriber = new Subscriber();
```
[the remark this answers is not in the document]
mpeddada1: Ah yes that is a valid point. I modified this comment to mention that these are default subscriber properties.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -282,4 +304,18 @@ public SubscriberStub createSubscriberStub() {
 		}
 	}
 
+	/**
+	 * Creates {@link ExecutorProvider} given subscriber properties.
+	 * @param subscriber subscriber properties
+	 * @return executor provider
+	 */
+	private ExecutorProvider getExecutorProvider(PubSubEventSpecificProperties.Subscriber subscriber) {
```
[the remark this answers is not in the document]
mpeddada1: Thank you! This is important since we're moving away from using Spring-managed beans.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -90,6 +94,8 @@
 
 	private final HeaderProvider headerProvider = new UserAgentHeaderProvider(this.getClass());
 
+	private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
elefeint, line 97: We usually have a static logger with a hardcoded class name. This prevents multiple instances of the same logger for each object instance (which is not really a concern for an autoconfiguration object, which is a singleton, but it's good to have consistency across the codebase).

Sonar will have an opinion on this, too -- https://rules.sonarsource.com/java/tag/convention/RSPEC-1312

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -45,26 +50,34 @@ public Publisher getPublisher() {
...
 
+	public Map<String, PubSubProperties> getProperties() {
+		return this.properties;
+	}
+
+	public Subscriber getSubscriber(String name) {
+		PubSubProperties property = new PubSubProperties();
+		property.setSubscriber(this.subscriber);
+		this.properties.putIfAbsent(name, property);
+		return this.properties.get(name).getSubscriber();
```
elefeint, line 61: I would immediately return the object if it's found in the map first. Then construct `PubSubProperties` only if we know there was not one at that particular point in time. Otherwise we'll end up creating a bunch of `PubSubProperties` objects and leaving them to immediately be garbage collected.

`putIfAbsent` is still necessary to prevent a race condition, in which the properties were not in the map at the time of the initial check, but have been added by another thread since then.

The current order also makes it look like the method creates `PubSubProperties`, sets a subscriber field, and then returns the subscriber field... So it causes a question why not just return `this.subscriber` here. And *then* you notice the _putIfAbsent_, resolving that question.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -27,13 +30,15 @@
  */
 public class PubSubConfiguration {
 
+	private final Map<String, PubSubProperties> properties = new HashMap<>();
```
elefeint, line 33: This is me being dense, because I think you already explained it, but what was the reason we could not have a direct `Map<String, PubSubConfiguration.Subscriber>`?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubProperties.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spring.pubsub.core;
+
+/**
+ * Intermediate class to hold subscription-specific properties.
+ */
+class PubSubProperties {
```
elefeint, line 22: If it's only a wrapper for subscription properties, we should name it `PubSubSubscriptionProperties`. If we have to keep this intermediate class, there will be another one for the publisher properties, right?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -486,6 +488,11 @@ else if (numCompletedFutures.incrementAndGet() == numExpectedFutures) {
 		return settableListenableFuture;
 	}
 
+	private SubscriberStub getSubscriberStub(String subscription) {
+		subscriptionNameToStubMap.putIfAbsent(subscription, this.subscriberFactory.createSubscriberStub(subscription));
```
elefeint, line 492: Similar to the `putIfAbsent` comment in autoconfig -- don't create an object (` this.subscriberFactory.createSubscriberStub(subscription)`), unless you already know it does not exist. Otherwise we'd create objects only to throw them away. 

So, "if stub already in the map, just return that. If not, then return a new one from subscriber factory.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -183,13 +215,15 @@ public Subscriber createSubscriber(String subscriptionName, MessageReceiver rece
 		Subscriber.Builder subscriberBuilder = Subscriber.newBuilder(
 				PubSubSubscriptionUtils.toProjectSubscriptionName(subscriptionName, this.projectId), receiver);
 
+		PubSubConfiguration.Subscriber subscriberProperties = this.pubSubConfiguration
```
elefeint, line 218: What happens if this is null? (it will be if the deprecated `DefaultSubscriberFactory` constructor is called). `Subscriber.Builder` has its own default, which it's probably fine to fall back on. But you definitely want to gate `subscriberBuilder.setExecutorProvider(getExecutorProvider(subscriberProperties, subscriptionName));` on the presence of at least `subscriberProperties`.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
elefeint: I think the latest was that we can remove our autoconfigured bean `subscriberExecutorProvider`, and mark the place that uses it as accepting `Optional`.

If we don't remove the bean, it will always get injected into the subscriber factory and override the per-subscription settings, won't it?

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactoryTests.java
```diff
@@ -40,6 +42,9 @@
 	@Mock
 	private CredentialsProvider credentialsProvider;
 
+	@Mock
+	private PubSubConfiguration pubSubConfiguration;
```
elefeint, line 46: You'll want to add tests for all kinds of configuration state -- what kind of executor provider ends up provided for a default set of properties, for a subscription-specific set of properties.
This is also where the "oops, what happens if someone uses the deprecated constructor" tests go -- I sometimes use tests to figure out what _should_ happen.

elefeint (changes requested):
Sorry, this took forever. But now this is a proper review -- since you rebased it, it's much easier to reason about just this part.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
mpeddada1: That makes sense, done.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
mpeddada1: Ah I missed this, thank you

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
mpeddada1: When we were using `Map<String, PubSubConfiguration.Subscriber>` with `spring.cloud.gcp.pubsub.properties.[sub-name].subscriber.executor-threads` the subscription-specific properties weren't getting recognized and we were just falling back to the default Subscriber properties. So I followed this [example](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.typesafe-configuration-properties.merging-complex-types) to add an intermediate class. 
I haven't verified this design yet but `Map<String, PubSubConfiguration.Subscriber>` might work if our configuration setup looks like `spring.cloud.gcp.pubsub.properties.[sub-name].executor-threads` -- we would have to rename `properties` to something more descriptive such as `subscriber-properties`. What do you think?

elefeint: Give it a try to see if it works. If it does not, let's stick with the wrapper class. But if it does, I think `spring.cloud.gcp.pubsub.subscription.[sub-name].executor-threads` would be a bit more user friendly than `spring.cloud.gcp.pubsub.properties.[sub-name].subscriber.executor-threads`.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
mpeddada1: Done. Added a `null` check for `this.pubSubConfiguration` when we call `subscriberBuilder.setExecutorProvider(getExecutorProvider(subscriberProperties, subscriptionName))` to prevent a `NullPointerException` from occurring.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
mpeddada1: Ah you're absolutely right! Sorry, I keep mixing up the definition of deprecation vs removal. Yes, we need to remove our bean and mark the place where it is used as Optional.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
mpeddada1: It worked! The new setup looks much better.

elefeint: Oh, good. I was worried I was being a thorn in your side for no tangible benefit.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
```diff
@@ -96,6 +96,52 @@ public void maxInboundMessageSize_default() {
...
+				.withPropertyValues(
+						"spring.cloud.gcp.pubsub.subscriber.executor-threads=5",
+						"spring.cloud.gcp.pubsub.subscription.subscription-name.executor-threads=3")
+				.withUserConfiguration(TestConfig.class);
+
+		contextRunner.run(ctx -> {
+			GcpPubSubProperties gcpPubSubProperties = ctx
+					.getBean(GcpPubSubProperties.class);
+			assertThat(gcpPubSubProperties.getSubscriber("subscription-name").getExecutorThreads()).isEqualTo(3);
+			assertThat(gcpPubSubProperties.getSubscriber("other").getExecutorThreads()).isEqualTo(5);
```
elefeint, line 141: It would be good to add another test similar to this one, but with 1 customized and 2-3 non-customized subscriptions, then get the `DefaultSubscriberFactory` bean out of context and call `getExecutorProvider` on it (it's package private, so doable). Then make sure that the non-customized subscriptions got the same executor provider (so, only 2 unique objects should get created).
Or you could put this test in `DefaultSubscriberFactoryTests`; that might be even better.

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactoryTests.java
```diff
@@ -89,4 +126,49 @@ public void testCreatePullRequest_nonNullMaxMessages() {
...
 		assertThat(request.getMaxMessages()).isEqualTo(Integer.MAX_VALUE);
 	}
+
+	@Test
+	public void testGetExecutorProvider_customProvided() {
+		DefaultSubscriberFactory factory = new DefaultSubscriberFactory(() -> "project", null);
+		factory.setExecutorProvider(mockExecutorProvider);
+
+		assertThat(factory.getExecutorProvider("name"))
+				.isEqualTo(mockExecutorProvider);
```
elefeint, line 136: s/isEqualTo/isSameAs, which is stricter.

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactoryTests.java
```diff
@@ -89,4 +126,49 @@ public void testCreatePullRequest_nonNullMaxMessages() {
...
+	}
+
+	@Test
+	public void testGetExecutorProvider_customNotProvided() {
+		DefaultSubscriberFactory factory = new DefaultSubscriberFactory(() -> "project", mockPubSubConfiguration);
+		when(mockPubSubConfiguration.getSubscriber("name")).thenReturn(new PubSubConfiguration.Subscriber());
+
+		assertThat(factory.getExecutorProvider("name"))
```
elefeint, line 144: Aha! This would be a good place to check that only one executor provider gets created for different subscriptions (unless there are customized properties).

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
```diff
@@ -96,6 +96,52 @@ public void maxInboundMessageSize_default() {
...
+						"spring.cloud.gcp.pubsub.subscription.subscription-name.executor-threads=7")
+				.withUserConfiguration(TestConfig.class);
+
+		contextRunner.run(ctx -> {
+			GcpPubSubProperties gcpPubSubProperties = ctx
+					.getBean(GcpPubSubProperties.class);
+			assertThat(gcpPubSubProperties.getSubscriber("subscription-name").getExecutorThreads()).isEqualTo(7);
+		});
+	}
+
```
elefeint, line 127: Looking at Sonar coverage, we also need a test for the scenario of "user provided their own `ExecutorProvider` bean. This is easy to set up -- use `withBean()` clause when creating the runner.

Take a look at the other uncovered conditional clauses in Sonar -- I like using Sonar as a tool for finding more logic to test. Every uncovered conditional is a potential landmine, so I try to test every path. I've been surprised many times when testing something trivial looking.

elefeint (changes requested):
So, it's not over. But it's looking better and better. The logic is good; now we need to pay attention to memory -- making sure we don't breed unnecessary objects.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -282,4 +316,42 @@ public SubscriberStub createSubscriberStub() {
...
+	 * @return executor provider
+	 */
+	ExecutorProvider getExecutorProvider(String subscriptionName) {
+		if (this.executorProvider != null) {
+			return this.executorProvider;
+		}
+		if (this.pubSubConfiguration != null) {
+			PubSubConfiguration.Subscriber subscriber = this.pubSubConfiguration.getSubscriber(subscriptionName);
+			this.threadPoolTaskScheduler = createThreadPoolTaskScheduler(subscriber, subscriptionName);
+			this.threadPoolTaskScheduler.initialize();
```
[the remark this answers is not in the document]
mpeddada1: `ThreadPoolTaskScheduler#getSchedulerExecutor` returns an IllegalStateException with the message `ThreadPoolTaskScheduler not initialized` if the scheduler not initialized.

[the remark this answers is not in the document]
elefeint: Interesting. I wonder if Spring used to call `initialize` as part of bean creation.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -282,4 +316,42 @@ public SubscriberStub createSubscriberStub() {
...
+	 * @param subscriptionName subscription name
+	 * @return executor provider
+	 */
+	ExecutorProvider getExecutorProvider(String subscriptionName) {
+		if (this.executorProvider != null) {
+			return this.executorProvider;
+		}
+		if (this.pubSubConfiguration != null) {
+			PubSubConfiguration.Subscriber subscriber = this.pubSubConfiguration.getSubscriber(subscriptionName);
+			this.threadPoolTaskScheduler = createThreadPoolTaskScheduler(subscriber, subscriptionName);
```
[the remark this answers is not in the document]
elefeint: Stub shutdown *is* tested in `PubSubSubscriberTemplateTests.destroyingBeanClosesSubscriberStub`

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubConfigurationTests.java
```diff
@@ -112,13 +112,42 @@ public void testSubscriberRetrySettings() {
...
+		assertThat(pubSubConfiguration.getSubscription().get("subscription-name").getExecutorThreads()).isEqualTo(4);
+	}
+
+	@Test
+	public void testSubscriberMapProperties_returnCustom() {
+		PubSubConfiguration pubSubConfiguration = new PubSubConfiguration();
+		PubSubConfiguration.Subscriber subscriber = new PubSubConfiguration.Subscriber();
+		subscriber.setExecutorThreads(8);
+
+		pubSubConfiguration.getSubscription().put("subscription-name", subscriber);
```
elefeint, line 130: (non-actionable) So, this was handy for testing, but returning mutable collections from code is generally bad for this exact reason. Ideally, all maps/lists/etc that come from getters should be turned immutable with `ImmutableMap.of()`, `ImmutableList.of()` etc.

However, Spring Boot itself is full of this pattern ([example](https://github.com/spring-projects/spring-boot/blob/c406dda18160a29649eef6bbb73f3c93674f4028/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java#L558)) -- properties are numerous, and only loaded at context startup, so I guess it's a pretty safe use in real life.

elefeint (approved):
Looks amazing. Tests are very thorough.

Hold off on merging until Monday -- I'd still want to run a stress test to compare memory usage.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -282,4 +321,85 @@ public SubscriberStub createSubscriberStub() {
...
+	 * Creates {@link ThreadPoolTaskScheduler} given subscriber properties.
+	 * @param subscriptionName subscription name
+	 * @return thread pool scheduler
+	 */
+	ThreadPoolTaskScheduler createThreadPoolTaskScheduler(PubSubConfiguration.Subscriber subscriber,
+			String subscriptionName) {
+		if (this.threadPoolTaskSchedulerMap.containsKey(subscriptionName)) {
+			return threadPoolTaskSchedulerMap.get(subscriptionName);
+		}
+		ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
```
poznachowski, line 383: Hey @elefeint @meltsufin @mpeddada1 !
By direct instantiation of `ThreadPoolTaskScheduler` here I think we lose the ability of distributed tracing instrumentation via [ExecutorBeanPostProcessor](https://github.com/spring-cloud/spring-cloud-sleuth/blob/main/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/async/ExecutorBeanPostProcessor.java).
It should be exposed as a bean (as it was before the MR) or [explicitly wrapped in a dedicated class](https://docs.spring.io/spring-cloud-sleuth/docs/current/reference/html/integrations.html#sleuth-async-executor-service-integration)

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -282,4 +321,85 @@ public SubscriberStub createSubscriberStub() {
...
+		return this.defaultExecutorProvider;
+	}
+
+	ExecutorProvider createExecutorProvider(PubSubConfiguration.Subscriber subscriber, String subscriptionName) {
+		if (this.executorProviderMap.containsKey(subscriptionName)) {
+			return this.executorProviderMap.get(subscriptionName);
+		}
+		ThreadPoolTaskScheduler scheduler = createThreadPoolTaskScheduler(subscriber, subscriptionName);
+		scheduler.initialize();
+		ExecutorProvider executor = FixedExecutorProvider.create(scheduler.getScheduledExecutor());
```
poznachowski, line 369: @elefeint I'm wondering why `ThreadPoolTaskScheduler` is being used here, as only `ScheduledExecutor` created under the hood is passed?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/TracingSubscriberFactory.java
```diff
@@ -48,7 +48,7 @@ public PullRequest createPullRequest(String subscriptionName, Integer maxMessage
 	}
 
 	@Override
-	public SubscriberStub createSubscriberStub() {
-		return pubSubTracing.subscriberStub(delegate.createSubscriberStub());
+	public SubscriberStub createSubscriberStub(String subscriptionName) {
```
meltsufin, line 51: Wouldn't this be considered a breaking change? Can we keep both versions of the method, and deprecate one?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -129,23 +134,6 @@ public ExecutorProvider publisherExecutorProvider(
 		return FixedExecutorProvider.create(scheduler.getScheduledExecutor());
 	}
 
-	@Bean
-	@ConditionalOnMissingBean(name = "pubsubSubscriberThreadPool")
-	public ThreadPoolTaskScheduler pubsubSubscriberThreadPool() {
```
meltsufin, line 134: Why is it that we no longer need these beans? Couldn't they be used when there is no subscription-specific config?

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubAutoConfigurationIntegrationTests.java
```diff
@@ -0,0 +1,121 @@
...
+				pubSubAdmin.deleteSubscription(subscriptionName);
+			}
+
+			pubSubAdmin.createTopic(topicName);
+			pubSubAdmin.createSubscription(subscriptionName, topicName, 10);
+
+			PubSubTemplate pubSubTemplate = context.getBean(PubSubTemplate.class);
+
+			pubSubTemplate.publish(topicName, "message1");
+			pubSubTemplate.pull(subscriptionName, 4, false);
```
meltsufin, line 74: How are you validating that the pull worked?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -282,4 +321,85 @@ public SubscriberStub createSubscriberStub() {
 		}
 	}
 
+	/**
+	 * Creates {@link ExecutorProvider}. If a custom executor provider is set then the
+	 * subscriber properties configured through the application.properties file will be
```
meltsufin, line 326: Why did you decide to have the custom `ExecutorProvider` to take precedence over subscription-specific config? I would expect it to be the opposite.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -318,40 +323,83 @@ public SubscriberStub createSubscriberStub(String subscriptionName) {
...
+	ExecutorProvider getExecutorProviderFromConfigurations(PubSubConfiguration.Subscriber subscriber,
+			String subscriptionName) {
+		Boolean isCustom = subscriber.isCustom();
+		if (Boolean.TRUE.equals(isCustom)) {
+			return createExecutorProvider(subscriber, subscriptionName);
+		}
+
+		if (this.defaultExecutorProvider != null) {
+			return this.defaultExecutorProvider;
+		}
```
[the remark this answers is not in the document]
meltsufin: Would you expect the legacy users also specify subscription-specific properties?
I feel like subscription-specific properties should always take precedence over globals.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
meltsufin: I guess the problem here is that there would potentially be a separate instance of such a bean for each of the subscriptions. @mpeddada1 Can you look into how to keep these beans in the application context so that they can be exposed to `ExecutorBeanPostProcessor`?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/SubscriberFactory.java
```diff
@@ -66,8 +66,9 @@ PullRequest createPullRequest(String subscriptionName, Integer maxMessages,
 
 	/**
 	 * Create a {@link SubscriberStub} that is needed to execute {@link PullRequest}s.
+	 * @param subscriptionName the subscription name
 	 * @return the {@link SubscriberStub} used for executing {@link PullRequest}s
 	 */
-	SubscriberStub createSubscriberStub();
+	SubscriberStub createSubscriberStub(String subscriptionName);
```
meltsufin, line 72: Are we OK with this breaking change?

meltsufin (commented):
Post-merge pass.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
mpeddada1: Good question. We made this bean optional for the subscriber factory in order to avoid breaking the current functionality for users who are using custom beans instead of autoconfiguration to set properties. If a custom bean is present then we use that otherwise we use the autoconfigured values. We need to remove our bean in order to prevent it from always getting injected into our subscriber factory and as a result, overriding the subscription-specific/default properties. Additionally, when there are no subscription-specific properties, `PubSubConfiguration` already takes care of using global (if specified) or default configurations as a fallback.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
mpeddada1: Yes, that is a valid point! Previously, the auto-config subscriber properties would be taken into account only when the custom bean was missing (as indicated by the `ConditionalOnMissingBean` annotation on our bean). This new logic follows that previous situation where the custom bean takes first priority over the auto-config (global or subscription-specific) properties. 

We also considered the reverse scenario where the subscription-specific would take precedence over the custom bean but the implementation details turned out to be tricky. We would have to add extra logic in `PubSubConfiguration` to find some way to determine if the property was set and different from the default as well as the global setting. This would have to be done for not just one property but all the other properties to come. As a compromise, the resolution was to allow custom beans to take precedence and if one would like to use subscription-specific properties then they will need to have to avoid using the custom bean. 

This change also needs to be carefully documented.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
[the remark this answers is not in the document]
mpeddada1: I think there will be two scenarios possible here. One where our users will use both the custom bean and the subscription-specific/global properties and another where they won't use the custom bean but rely on the configuration settings. In the first scenario, the custom bean will take precedence (given the reasoning [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/562/files#r700562596)) and in the second scenario, the subscription-specific properties will take precedence over the global settings.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/TracingSubscriberFactory.java
mpeddada1: That is a good catch! This is indeed a breaking change and you're right that it will be safer to deprecate this and add a new one. I will make this change in a follow-up PR.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubAutoConfigurationIntegrationTests.java
mpeddada1: Good point. We may be under-asserting here. We could add an assert statement to check if the published messages have been received. Will also add this change in a follow-up PR.

meltsufin: Alternatively, maybe we don't even need to test pulling here? We test it in other places.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
meltsufin: Should we open an issue to track documentation/clarification of this?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/SubscriberFactory.java
meltsufin: Also here, maybe create an issue to remember to rollback the breaking changes from this PR.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
mpeddada1: Good point, done.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
mpeddada1: I see. Created an issue to track this, thank you.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
elefeint: We tend to prefer Spring abstractions over plain Java ones for convenience and consistency with other Spring modules. `ThreadPoolTaskScheduler` comes with easy customization of thread names and such.

poznachowski: Thanks, that makes perfect sense.

PR #569, opened 2021-08-10
Fix Datastore pageable return type compatibility

### LINKED ISSUE #564: Update DatastorePageable next/previous to return PageRequest
Return types for `previous()` and `next()` were restricted from generic `Pageable` to `PageRequest`.

Update `DatastorePageable` with the new return types. Slight complication that we pass some non-paged pageables into `DatastorePageable` static methods.

elefeint:
This PR allows Datastore to work with both Spring Data/Spring Boot 2.4 and 2.5. It also adds placeholder methods for the new `CrudRepository` method `deleteAllById()` to fix compilation issues with 2.5.

I've inlined `super.next()` creating a new `PageRequest` because there were inheritance issues with Spring Data 2.4.9 (at runtime, the overridden superclass method was not called, causing the infinite loop. But this only happens if the signature is `public PageRequest next() {`. If the signature is `public Pageable next() {`, java finds the overridden method, and all is well. Classic "this should never happen" type of Java situation, which I would normally troubleshoot to bitter end, but inlining the `PageRequest` creation also improves readability).

Fixes #564.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/DatastorePageable.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/query/DatastorePageable.java
@@ -24,7 +24,7 @@ import org.springframework.data.domain.Pageable;
 /**
  * A pageable implementation for Cloud Datastore that uses the cursor for efficient reads.
  *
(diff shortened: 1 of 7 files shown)


meltsufin (approved):
Note that you can use `{@link}` vs. `{@code}` in most instances when referencing a class. It makes it easier to navigate in javadocs.

PR #574, opened 2021-08-11
Fix Timestamp as datastore supported types

### LINKED ISSUE #552: DatastoreRepository: problem with Timestamp field as query parameter
Hi everyone,
we have an entity with a Timestamp field

```java
@Entity
public class WeatherData {
    @Id
    private String id;
    
    private Double latitude;
    private Double longitude;
    
    private Timestamp date;

    private Double temperature;
    private Double windSpeed;
    private Double windDirection;
    private Double directNormalIrradiance;
    private Double humidity;
    private Double pressure;
   
    // setters, getters
}
```
and its repository exposed as ReST resource
```java
@RepositoryRestResource(path = "weatherdata")
@Tag(name = "weather-data", description = "Weather Data API")
public interface WeatherDataRepository extends DatastoreRepository<WeatherData, String> {

@RestResource(path = "findByDates")
    @Operation(summary = "Find weather data for a range of dates", tags = { "renewable-production" })
    @Query("SELECT * FROM weatherData "
            + "WHERE "
            + "date <= @dateTo AND date >= @dateFrom")
    public Page<WeatherData> findByDates(
             @Param("dateFrom") Timestamp dateFrom
            , @Param("dateTo") Timestamp dateTo
            , Pageable pageable );

```
we have also defined a converter for date parameters
```java
@Configuration
    public class RenewableRestConfiguration implements RepositoryRestConfigurer {
        @Override
        public void configureConversionService(ConfigurableConversionService conversionService) {
            conversionService.addConverter(STRING_TO_GCP_TIMESTAMP_CONVERTER);
        }
    }

    static final Converter<String, Timestamp> STRING_TO_GCP_TIMESTAMP_CONVERTER = new Converter<String, Timestamp>() {
        @Override
        public Timestamp convert(String source) {
            return Timestamp.parseTimestamp(source);
        }

    };
```
when we call the ReST API like this
```bash
curl --request GET \
  --url 'http://localhost:8080/weatherdata/search/findByDates?dateFrom=2021-07-10T09%3A00%3A00&dateTo=2021-07-22T11%3A00%3A00'
```
first invocation works.
The next one crash with this exception
`com.google.cloud.spring.data.datastore.core.mapping.DatastoreDataException: An ID property was required but does not exist for the type: class com.google.cloud.Timestamp`

We debugged the `GqlDatastoreQuery` class and we found that the com.google.cloud.Timestamp are in sort of persistent entities cache maintained by `DatastoreMappingContext'.

What could be the problem?
Thanks.

spring-cloud: 2.0.3
spring-boot: 2.4.6

zhumin8:
Fixes #552.
Added Datastore native types on top of Spring natives as simple types in `DatastoreMappingContext` constructor.
Also added tests in `DatastoreMappingContextTests.java` and `DatastoreIntegrationTests.java` .

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/mapping/DatastoreMappingContext.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/mapping/DatastoreMappingContext.java
@@ -22,6 +22,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
+import com.google.cloud.spring.data.datastore.core.convert.DatastoreNativeTypes;
+
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.data.mapping.context.AbstractMappingContext;
@@ -54,6 +56,10 @@ public class DatastoreMappingContext extends
 	// Maps a given class to the set of other classes with which it shares the same Datastore
 	// Kind and that are subclasses of the given class.
 	private static final Map<Class, Set<Class>> discriminationFamilies = new ConcurrentHashMap<>();
+
+	public DatastoreMappingContext() {
+		this.setSimpleTypeHolder(DatastoreNativeTypes.HOLDER);
+	}
 
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) {
--- a/spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/core/mapping/DatastoreMappingContextTests.java
+++ b/spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/core/mapping/DatastoreMappingContextTests.java
@@ -72,7 +72,7 @@ public class DatastoreMappingContextTests {
 	}
 
 	@Test
-	public void testGetInvalidEntityTimeStamp() {
+	public void testGetInvalidEntityTimestamp() {
 		DatastorePersistentEntityImpl mockEntity = mock(
 				DatastorePersistentEntityImpl.class);
 		DatastoreMappingContext context = createDatastoreMappingContextWith(mockEntity);
--- a/spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/core/mapping/DatastoreMappingContextTests.java
+++ b/spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/core/mapping/DatastoreMappingContextTests.java
@@ -23,7 +23,6 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.data.util.ClassTypeInformation;
 import org.springframework.data.util.TypeInformation;
 
-import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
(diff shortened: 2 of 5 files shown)


spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/mapping/DatastoreMappingContext.java
```diff
@@ -55,6 +57,10 @@
 	// Kind and that are subclasses of the given class.
 	private static final Map<Class, Set<Class>> discriminationFamilies = new ConcurrentHashMap<>();
 
+	public DatastoreMappingContext() {
+		this.setSimpleTypeHolder(DatastoreNativeTypes.HOLDER);
```
meltsufin, line 61: It would nice to have a test validating that the `simpleTypeHolder` was set to `DatastoreNativeTypes.HOLDER`.

meltsufin (commented):
Seems like a simple fix, and it works!
Nice job navigating through the complexity of Datastore code!
I just left a couple of minor comments/questions.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/TestEntityRepository.java
```diff
@@ -140,6 +141,11 @@
 	@Query("select * from  test_entities_ci where color = @color")
 	Stream<TestEntity> findGqlStreamByColor(@Param("color") String color);
 
+	@Query("select * from  test_entities_ci where datetime >= @startTime AND datetime <= @endTime")
+	List<TestEntity> getAllBetweenDates(@Param("startTime") Timestamp startTime, @Param("endTime") Timestamp endTime);
+
+	List<TestEntity> findByDatetimeGreaterThan(Timestamp oldestTime);
```
elefeint, line 147: Nice integration test; hits both GQL and PartTree code paths!

elefeint (approved):
Excellent -- simple fix for a gnarly problem!

spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/mapping/DatastoreMappingContext.java
elefeint: It would be nice to validate that both, built-in types and datastore native types got added... But it might not be possible, since `simpleTypeHolder` is not really exposed from `AbstractMappingContext`.

A method `shouldCreatePersistentEntityFor()` might be helpful, but that's protected, so you would have to subclass it to test.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/core/mapping/DatastoreMappingContextTests.java
```diff
@@ -69,6 +70,14 @@ public void testGetInvalidEntity() {
 				.hasMessage("Unable to find a DatastorePersistentEntity for: class java.lang.Integer");
 	}
 
+	@Test
+	public void testGetSimpleTypeTimestamp() {
```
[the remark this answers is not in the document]
meltsufin: `testTimestampNotAnEntity`?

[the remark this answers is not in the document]
zhumin8: Since `simpleTypeHolder` is a private member, this was the closest I could get to test that `Timestamp` is treated as a simple type via `shouldCreatePersistentEntityFor` logic.
Let me change the test name and leave a short comment.

Issue #581, opened 2021-08-18
Publishing to and Subscribing from multiple Pub/Sub projects is extremely difficult


NickGrantX:
**Is your feature request related to a problem? Please describe.**
We publish and subscribe to/from topics and subscriptions in other projects, and the libraries do not support this at all well, especially with the Pub/Sub emulator complication.

Please have a look at how we did it, to get a sense of how difficult it was to meet the basic requirement of being able to specify a project for a topic to publish to, or a subscription to pull from: https://github.com/ONSdigital/ssdc-rm-caseprocessor/pull/50

**Describe the solution you'd like**
It would be ideal if there was simply an optional "project ID" parameter on the `PubSubInboundChannelAdapter` class constructor, in addition to the subscription name. For example:

```java
new PubSubInboundChannelAdapter(pubSubTemplate, subscriptionName, projectId);
```

Likewise, it would be ideal if there was an optional "project name" parameter `PubSubTemplate.publish()` function. For example:

```java
pubSubTemplate.publish(topic, message, projectId);
```

This would make it very trivial to be able to publish to a topic in a different project, or subscribe to a subscription in a different project.

**Describe alternatives you've considered**
We have written a lot of config to work around this, as you can see here: https://github.com/ONSdigital/ssdc-rm-caseprocessor/pull/50

In my opinion, you shouldn't have to write so much code just to specify the project.

**Additional context**
There's not a lot of documentation around this. I had to dig in the source code to arrive at the solution, which in fact is not complete in the PR I attached... we had to do extra stuff because the `GcpPubSubProperties` and `PubsubTransportConfig` behaved in a really unexpected way in production. We ended up bringing more and more of the `GcpPubSubAutoConfiguration` class into our code, to make it do what we need.

Our requirement is fairly simple: to be able to do integration tests with the pubsub emulator, to be able to run on our local machine with the pubsub emulator, and then to be able to deploy to GCP with very similar code... except we've ended up with a lot of code having to be in our "prod" codebase just for testing, which is really not ideal. It should be easier.

We appreciate the help with this. There's not a lot of stuff in the docs or in the wider community about how to achieve what we need. I found just 1 (one) single stackoverflow post about this: https://stackoverflow.com/questions/61495154/how-to-subscribe-to-multiple-google-pubsub-projects-in-spring-gcp

suztomo:
[...]

PR #584, opened 2021-08-19
Add PubSubTemplate ProjectTopicName/ProjectSubscriptionName parameters

NickGrantONS:
# Motivation and Context
The PubSubTemplate interface contract does not make it clear that it's possible to specify a fully qualified topic or subscription name, including the project. The documentation of this feature is present, but it's easily overlooked.

# What has changed
These convenience methods make it clear that a ProjectSubscriptionName or a ProjectTopicName can be supplied with a project, to make it easier for developers to understand this functionality/capability.

# How to test?
Zero regression - existing test all pass.

# Links
Original issue/feature request: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/581

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubTemplate.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.spring.pubsub.core;
 
+import com.google.pubsub.v1.ProjectSubscriptionName;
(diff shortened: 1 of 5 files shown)


meltsufin (commented):
Please see might comment in the issue (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/581#issuecomment-901934010). I think we should keep the API as is, but perhaps improve reference documentation.

NickGrantONS:
> Please see might comment in the issue ([#581 (comment)](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/581#issuecomment-901934010)). I think we should keep the API as is, but perhaps improve reference documentation.

Will close this PR then

Issue #472, opened 2021-05-25
Upgrade to Spring Boot 2.5


meltsufin:
Spring Boot 2.5 is a major version upgrade, and so cannot be done automatically.

## Changes to address for 2.5:

New method in `CrudRepository`: [`deleteAllById()`](https://github.com/spring-projects/spring-data-commons/pull/476). Since we support both Spring Boot 2.4 and 2.5, this means we'll have to omit `@Overrides` annotation to maintain ability to compile against 2.4.
- [ x] Implement in `SimpleDatastoreRepository`  (#565)
- [ x] Implement in `SimpleFirestoreReactiveRepository` (#566)
- [ x] Implement in `SimpleSpannerRepository` (#567)

Return types for `previous()` and `next()` were restricted from generic `Pageable` to `PageRequest`.
- [ x] Update `DatastorePageable` with the new return types. Slight complication that we pass some non-paged pageables into `DatastorePageable` static methods (#564).  

Ref: #471

elefeint:
Spring Cloud tracks Spring Boot 2.5 migration in [3.1.x branch](https://github.com/spring-cloud/spring-cloud-build/blob/3.1.x)

spencergibb:
@elefeint the 3.1.x branch will only be 2.5 until 2.6 starts. 2.5 support is on the 3.0 branch but as forwards compatible. So spring cloud 2020.0.3 supports both boot 2.4 and 2.5. Spring cloud 2021.0 will support 2.6

elefeint:
@spencergibb That's good to know, thank you -- I remember that the original Illford announcement was for both, 2.4 and 2.5 Boot support, but then I saw that Spring Cloud Build main is [still on 2.4.6](https://github.com/spring-cloud/spring-cloud-build/blob/6ff762f675d6c8302dcab435a4e984dd86a65605/pom.xml#L31).

Out of curiosity, how will Spring Cloud train test compatibility with both, 2.4 and 2.5 -- will you be running parallel builds with a different Spring Boot dependency? We would face the same issue if we support both versions simultaneously...

spencergibb:
We run all of our builds with boot 2.5. We actually change the version of boot in spring cloud build, build it locally, then run the project build

meltsufin:
Thanks for the clarification!

elefeint:
[...]

PR #590, opened 2021-08-23
Implement deleteAllById() in SimpleFirestoreReactiveRepository 

### LINKED ISSUE #566: Implement `deleteAllById()` in `SimpleFirestoreReactiveRepository`
Spring Data 2.5 introduced a new method in `CrudRepository`: [`deleteAllById()`](https://github.com/spring-projects/spring-data-commons/pull/476). Since we support both Spring Boot 2.4 and 2.5, this means we'll have to omit `@Overrides` annotation to maintain ability to compile against 2.4.

zhumin8:
Implement `deleteAllById()` in `SimpleFirestoreReactiveRepository` and add relevant tests.
Resolves #566.
This is in the checklist of #472.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
+++ b/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
@@ -215,6 +215,23 @@ public class FirestoreTemplateTests {
 		verify(this.firestoreStub, times(1)).commit(eq(builder.build()), any());
 	}
 
+	@Test
+	public void deleteByIdTest() {
+		mockCommitMethod();
+		Flux<String> idPublisher = Flux.just("e1", "e2");
+		StepVerifier.create(
+				this.firestoreTemplate
+						.deleteById(idPublisher, TestEntity.class))
+				.verifyComplete();
+
+		CommitRequest.Builder builder = CommitRequest.newBuilder()
+				.setDatabase("projects/my-project/databases/(default)");
+
+		builder.addWrites(Write.newBuilder().setDelete(parent + "/testEntities/e1").build());
+		builder.addWrites(Write.newBuilder().setDelete(parent + "/testEntities/e2").build());
+
+		verify(this.firestoreStub, times(1)).commit(eq(builder.build()), any());
+	}
 
 	private void mockCommitMethod() {
 		doAnswer(invocation -> {
--- a/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
+++ b/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
@@ -230,7 +230,7 @@ public class FirestoreTemplateTests {
 		builder.addWrites(Write.newBuilder().setDelete(parent + "/testEntities/e1").build());
 		builder.addWrites(Write.newBuilder().setDelete(parent + "/testEntities/e2").build());
 
-		verify(this.firestoreStub, times(1)).commit(eq(builder.build()), any());
+		verify(this.firestoreStub).commit(eq(builder.build()), any());
 	}
 
 	private void mockCommitMethod() {
--- a/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/SimpleFirestoreReactiveRepositoryTests.java
+++ b/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/SimpleFirestoreReactiveRepositoryTests.java
@@ -17,13 +17,15 @@
 package com.google.cloud.spring.data.firestore;
 
 import java.util.Arrays;
+import java.util.List;
 
 import org.junit.Test;
+import org.mockito.ArgumentCaptor;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.ArgumentMatchers.same;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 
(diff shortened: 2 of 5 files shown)


spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/it/FirestoreIntegrationTests.java
```diff
@@ -273,6 +273,14 @@ public void writeReadDeleteTest() {
 		assertThat(this.firestoreTemplate.deleteAll(User.class).block()).isEqualTo(2);
 		assertThat(this.firestoreTemplate.findAll(User.class).collectList().block()).isEmpty();
 
+		alice.setUpdateTime(null);
+		this.firestoreTemplate.save(alice).block();
+		bob.setUpdateTime(null);
+		this.firestoreTemplate.save(bob).block();
+
+		this.firestoreTemplate.deleteById(Flux.just("Bob", "Saitama", "Alice"), User.class).block();
+		assertThat(this.firestoreTemplate.count(User.class).block()).isZero();
```
elefeint, line 282: Instead of blocking, try using `StepVerifier` here -- it will both subscribe to the publisher stream, and allow you to assert the stream content with `.expectNext()`.

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/it/FirestoreIntegrationTests.java
```diff
@@ -273,6 +273,14 @@ public void writeReadDeleteTest() {
 		assertThat(this.firestoreTemplate.deleteAll(User.class).block()).isEqualTo(2);
 		assertThat(this.firestoreTemplate.findAll(User.class).collectList().block()).isEmpty();
 
+		alice.setUpdateTime(null);
```
elefeint, line 276: I think it would be easier for us to troubleshoot in case of failure if `deleteAllById()` had a dedicated test method. We do combine different scenarios for integration tests when it's hard to set up each test individually, but in this case the two new objects need to be set up anyway (since Alice object got deleted a few lines above).

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/it/FirestoreRepositoryIntegrationTests.java
```diff
@@ -114,6 +116,11 @@ public void writeReadDeleteTest() {
 		User aliceLoaded = this.userRepository.findById("Alice").block();
 		assertThat(aliceLoaded.getAddresses()).isEqualTo(addresses);
 		assertThat(aliceLoaded.getHomeAddress()).isEqualTo(homeAddress);
+
+		// cast to SimpleFirestoreReactiveRepository for method be reachable with Spring Boot 2.4
+		SimpleFirestoreReactiveRepository repository = AopTestUtils.getTargetObject(this.userRepository);
+		repository.deleteAllById(Arrays.asList("Alice", "Bob")).block();
+		assertThat(this.userRepository.count().block()).isEqualTo(0);
```
elefeint, line 123: Same, use `StepVerifier` for idiomatic publisher testing testing.

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/SimpleFirestoreReactiveRepositoryTests.java
```diff
@@ -16,21 +16,32 @@
...
+	public void deleteAllById() {
 		FirestoreTemplate mockTemplate = mock(FirestoreTemplate.class);
-		SimpleFirestoreReactiveRepository<String> repository =
-				new SimpleFirestoreReactiveRepository<>(mockTemplate, String.class);
-		assertThatThrownBy(() -> repository.deleteAllById(new ArrayList<>()))
-				.isInstanceOf(UnsupportedOperationException.class);
+		SimpleFirestoreReactiveRepository<String> repository = new SimpleFirestoreReactiveRepository<>(mockTemplate,
+				String.class);
+		Iterable<String> idList = Arrays.asList("1", "2");
+		repository.deleteAllById(idList);
```
elefeint, line 40: I'd leave a comment above this line saying that the test is only testing that the request was passed through to `FirestoreTemplate`, so the returned `Publisher` does not need to be subscribed to. It's a bit jarring to see a publisher output discarded -- it would lead to the functionality never being invoked in real life.

Alternatively, you could do what you've done in `FirestoreTemplateTests` -- set up a when/then mock structure for `mockTemplate` that *only* matches the correct publisher input, and then instead of doing argument captor below, a `StepVerifier` can be used to subscribe and validate that the response of `repository.deleteAllById` matches the output that mockito would only give on correct parameters.

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
```diff
@@ -215,6 +215,23 @@ public void deleteTest() {
...
+						.deleteById(idPublisher, TestEntity.class))
+				.verifyComplete();
+
+		CommitRequest.Builder builder = CommitRequest.newBuilder()
+				.setDatabase("projects/my-project/databases/(default)");
+
+		builder.addWrites(Write.newBuilder().setDelete(parent + "/testEntities/e1").build());
+		builder.addWrites(Write.newBuilder().setDelete(parent + "/testEntities/e2").build());
+
+		verify(this.firestoreStub, times(1)).commit(eq(builder.build()), any());
```
elefeint, line 233: `times(1)` is the default value, so it can be omitted -- https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html
(although I keep having to re-search for this documentation because I don't seem to trust it for some reason)

elefeint (changes requested):
Thank you for adding extra tests to the template! 
I am curious how you liked the reactive style now.
Sonar has a minor quibble.

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/it/FirestoreIntegrationTests.java
zhumin8: Ah, thanks for the suggestion. `StepVerifier` are quite powerful in testing reactive stream!

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/SimpleFirestoreReactiveRepositoryTests.java
zhumin8: Tried with this alternative route, but stuck on comparing two publisher input: `Flux` equals is inherited from Object, I cannot find easy ways to compare the input for mock. The argument captor looks more intuitive to me.
Going to just add the comment instead.

PR #601, opened 2021-09-01
Implement subscription-specific property configurations 

mpeddada1:
Implementing subscription-specific property configurations for:

- `retry settings`
- `flow control settings`
- Pull config: `max-ack-extension-period`, `parallel pull count` and `pull endpoint`

As done in #562, we will:
- Remove our beans for  subscriber `flow control settings` and `retry settings`.
- ~~Deprecate the setters for `flow control settings` and `retry settings` and remove setters for the `pull config` properties.~~ The setters will take precedence over auto config if used.
- Use the custom property bean if provided otherwise use the values derived from `PubSubConfiguration`.


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubAutoConfigurationIntegrationTests.java
```diff
@@ -37,11 +39,26 @@
 public class PubSubAutoConfigurationIntegrationTests {
 
 	private static final Log LOGGER = LogFactory.getLog(PubSubTemplateIntegrationTests.class);
+
 	private static GcpProjectIdProvider projectIdProvider;
+
 	private ApplicationContextRunner contextRunner = new ApplicationContextRunner()
-			.withPropertyValues("spring.cloud.gcp.pubsub.subscriber.max-ack-extension-period=0",
+			.withPropertyValues(
 					"spring.cloud.gcp.pubsub.subscription.test-sub-1.executor-threads=3",
```
dzou, line 47: Minor nit: For this test case, can you group all the `test-sub-1` properties together and the `test-sub-2` properties together?

dzou (commented):
Looks good on my end.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubAutoConfigurationIntegrationTests.java
mpeddada1: This definitely improves readability, thanks!

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactoryTests.java
```diff
@@ -144,6 +150,30 @@ public void testCreatePullRequest_nonNullMaxMessages() {
...
+
+		Subscriber expectedSubscriber = factory.createSubscriber("defaultSubscription", (message, consumer) -> {
+		});
+
+		assertThat(expectedSubscriber.getFlowControlSettings().getLimitExceededBehavior())
+				.isEqualTo(FlowController.LimitExceededBehavior.Ignore);
+		assertThat(expectedSubscriber).hasFieldOrPropertyWithValue("maxAckExtensionPeriod", Duration.ofSeconds(2L))
+				.hasFieldOrPropertyWithValue("numPullers", 2);
+
+		factory.getThreadPoolTaskSchedulerMap().get("defaultSubscription").shutdown();
```
mpeddada1, line 174: Added tests to make sure that these properties are set correctly. However, I'm wondering if asserting directly on the field name and value will make the test brittle. Any thoughts?

meltsufin (commented):
As discussed offline, I think we should consider a design where properties feed into beans and beans feed into the `DefaultSubscriberFactory`.
So, you might have beans in the application context that are specific to a subscription-name, or not.
`sub1RetrySettings`, `sub2RetrySettings`, etc.
Then in the `DefaultSubscriberFactory` we search the application context for presence of subscription-specific beans and use them, if they exist. Otherwise, we fall back to the default beans, i.e `retrySettings`, etc. Of course, this can be done in a follow-up PR, if it makes sense at all.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -169,35 +174,14 @@ public void setSystemExecutorProvider(ExecutorProvider systemExecutorProvider) {
...
+	@Deprecated
 	public void setFlowControlSettings(FlowControlSettings flowControlSettings) {
 		this.flowControlSettings = flowControlSettings;
 	}
 
-	/**
-	 * Set the maximum period the ack timeout is extended by.
-	 * @param maxAckExtensionPeriod the max ack extension period to set
-	 */
-	public void setMaxAckExtensionPeriod(Duration maxAckExtensionPeriod) {
```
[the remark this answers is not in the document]
mpeddada1: Ah that is a good point. Kept this and the other setters to allow users to set the global property values without using application.properties.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -210,7 +194,9 @@ public void setApiClock(ApiClock apiClock) {
 	 * Set the retry settings for the generated subscriber stubs.
 	 * @param subscriberStubRetrySettings parameters for retrying pull requests when they fail,
 	 * including jitter logic, timeout, and exponential backoff
+	 * @deprecated Directly use application.properties to configure Retry instead of Spring-managed beans
```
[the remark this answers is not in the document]
mpeddada1: You make a valid point. I removed the `@Deprecated` annotation for this and the other properties to allow users to use setters (to set global configurations) if they prefer not to use `application.properties`.

meltsufin (approved):
LGTM, assuming we do not release before addressing the follow-up issues you've filed a few days ago after the initial review.

mpeddada1:
> LGTM, assuming we do not release before addressing the follow-up issues you've filed a few days ago after the initial review.

Thanks for reviewing. Yup, we will not make a release until the follow-up issues have been addressed.

PR #611, opened 2021-09-08
Add implementation to choose global properties when subscription-specific properties are not specified

### LINKED ISSUE #605: Global setting gets ignored when we specify a subscription-specific setting for a different property
When we specify subscription-specific settings for `property1` of `subscription-one`, the global setting of `property2` (which is not subscription-specific) gets ignored. 

Found this issue when writing the following test:

```@Test
public void executorThreads_globalSpecifiedAndSelectiveNotSet() {
  ApplicationContextRunner contextRunner = new ApplicationContextRunner()
        .withConfiguration(AutoConfigurations.of(GcpPubSubAutoConfiguration.class))
        .withPropertyValues(
              "spring.cloud.gcp.pubsub.subscriber.executor-threads=5",
              "spring.cloud.gcp.pubsub.subscription.subscription-one.pull-endpoint=my-endpoint")
        .withUserConfiguration(TestConfig.class);

  contextRunner.run(ctx -> {
     GcpPubSubProperties gcpPubSubProperties = ctx
           .getBean(GcpPubSubProperties.class);
     GcpProjectIdProvider projectIdProvider = ctx.getBean(GcpProjectIdProvider.class);
     assertThat(gcpPubSubProperties.getSubscriber("subscription-one", projectIdProvider.getProjectId())
           .getExecutorThreads()).isEqualTo(5);
     assertThat(gcpPubSubProperties.getSubscriber("other", projectIdProvider.getProjectId())
           .getExecutorThreads()).isEqualTo(5);
  });
}
```
The test above fails because it expects the value of `gcpPubSubProperties.getSubscriber("subscription-one", projectIdProvider.getProjectId()).getExecutorThreads()` to be `4` (the default).

mpeddada1:
Fixes #605

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
@@ -129,7 +129,11 @@ public class GcpPubSubAutoConfigurationTests {
 		contextRunner.run(ctx -> {
 			GcpPubSubProperties gcpPubSubProperties = ctx
 					.getBean(GcpPubSubProperties.class);
+			GcpProjectIdProvider projectIdProvider = ctx.getBean(GcpProjectIdProvider.class);
 			assertThat(gcpPubSubProperties.getSubscriber().getExecutorThreads()).isEqualTo(7);
+			assertThat(
+					gcpPubSubProperties.computeSubscriberExecutorThreads("subscription-name", projectIdProvider.getProjectId()))
+							.isEqualTo(7);
 		});
 	}
 
@@ -145,8 +149,9 @@ public class GcpPubSubAutoConfigurationTests {
 			GcpPubSubProperties gcpPubSubProperties = ctx
 					.getBean(GcpPubSubProperties.class);
 			GcpProjectIdProvider projectIdProvider = ctx.getBean(GcpProjectIdProvider.class);
-			assertThat(gcpPubSubProperties.getSubscriber("subscription-name", projectIdProvider.getProjectId())
-					.getExecutorThreads()).isEqualTo(7);
+			assertThat(
+					gcpPubSubProperties.computeSubscriberExecutorThreads("subscription-name", projectIdProvider.getProjectId()))
+							.isEqualTo(7);
 		});
 	}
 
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -81,6 +83,23 @@ public Subscriber getSubscriber(String name, String projectId) {
...
+	 * @param subscriptionName subscription name
+	 * @param projectId project id
+	 * @return number of executor threads
+	 */
+	public int computeSubscriberExecutorThreads(String subscriptionName, String projectId) {
+		Integer numThreads = getSubscriber(subscriptionName, projectId).getExecutorThreads();
+		if (numThreads != null) {
+			return numThreads;
+		}
+		Integer globalExecutorThreads = subscriber.getExecutorThreads();
```
dzou, line 99: Minor nit: Would be nice if you could rename `subscriber` to `defaultSubscriber` to provide a clearer difference between `subscriber` and `getSubscriber(subscriptionName, projectId)`.

dzou (approved):
LGTM on my end.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
```diff
@@ -163,13 +168,57 @@ public void executorThreads_globalAndSelectiveConfigurationSet_selectiveTakesPre
...
 					.containsKey("projects/fake project/subscriptions/subscription-name");
 			assertThat(
-					gcpPubSubProperties.getSubscriber("other", projectIdProvider.getProjectId()).getExecutorThreads())
+					gcpPubSubProperties.computeSubscriberExecutorThreads("other", projectIdProvider.getProjectId()))
+							.isEqualTo(5);
+		});
+	}
+
+	@Test
+	public void executorThreads_globalAndDifferentSelectiveConfigurationSet_pickGlobal() {
```
dzou, line 183:
```suggestion
	public void executorThreads_inheritsGlobalConfig() {
```

nit: suggestion to rename test (maybe is more concise)

dzou: Nevermind, I think the way you had better matches the existing naming conventions.

dzou:
Seems like this PR only touches `executorThreads`; do you have to consider more fields in order to fully solve this problem?

mpeddada1:
> Seems like this PR only touches `executorThreads`; do you have to consider more fields in order to fully solve this problem?

Yes, that's right. Similar methods would need to be added for other properties as well.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
mpeddada1: That's a good idea! I assumed Spring autoconfig wouldn't be able to pick up the global config if the variable was renamed to something else but interestingly it works when the getter's name stays as `getSubscriber()`.

meltsufin (approved):

PR #612, opened 2021-09-09
Fix the GraalVM Github Actions step

dzou:
The old GHA step for GraalVM seems to have broken; this uses a different implementation.

### DIFF
--- a/.github/workflows/integrationTests.yaml
+++ b/.github/workflows/integrationTests.yaml
@@ -54,6 +54,8 @@ jobs:
           graalvm: '21.2.0'
           java: 'java11'
           arch: 'amd64'
+      - name: Install GraalVM Native Image
+        if: matrix.it == 'native'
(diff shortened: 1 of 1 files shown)


mpeddada1 (approved):

.github/workflows/integrationTests.yaml
```diff
@@ -48,12 +48,15 @@ jobs:
...
-          java-version: "java11"
-          native-image: "true"
+          graalvm: '21.2.0'
+          java: 'java11'
+          arch: 'amd64'
+      - name: Install GraalVM Native Image
+        if: matrix.it == 'native'
+        run: gu install native-image
       - uses: actions/cache@v2
         id: mvn-cache
```
suztomo, line 58: My 2 cents: if we cache the GraalVM native image installation too, this job may become a bit faster. (I don't know the bottleneck)

Issue #597, opened 2021-08-30
Find why TraceSampleApplicationIntegrationTests flakes consistently


elefeint:
After #548 was merged, `TraceSampleApplicationIntegrationTests` started flaking. The scheduled daily test run often fails on the weekend -- I usually re-run on Monday, and the same run succeeds.

Some diagnostic information is added in #587.

If the test succeeds, it succeeds quickly. Otherwise it keeps pinging the backend trying to find the 25th span. The span that is missing is one of the final "send" spans on publish. In related news, the successful runs have both subscriber threads outputting "Overriding default instance of MessageHandlerMethodFactory with provided one.", while the failed runs only have one of the threads printing this out.

Bad run:
```
13:00:52.234 [http-nio-auto-1-exec-1] INFO  com.example.WorkService - starting busy work
13:00:52.901 [http-nio-auto-1-exec-2] INFO  com.example.ExampleController - meeting took 658ms
13:00:53.614 [http-nio-auto-1-exec-3] INFO  com.example.ExampleController - meeting took 650ms
13:00:54.224 [http-nio-auto-1-exec-4] INFO  com.example.ExampleController - meeting took 589ms
13:00:54.229 [http-nio-auto-1-exec-1] INFO  com.example.WorkService - finished busy work
13:00:55.522 [gcp-pubsub-subscriber2] INFO  o.s.i.h.s.MessagingMethodInvokerHelper - Overriding default instance of MessageHandlerMethodFactory with provided one.
13:00:55.527 [gcp-pubsub-subscriber1] INFO  o.s.i.h.s.MessagingMethodInvokerHelper - Overriding default instance of MessageHandlerMethodFactory with provided one.
13:00:55.570 [gcp-pubsub-subscriber1] INFO  com.example.Application - Message arrived! Payload: All work is done via PubSubTemplate.
13:00:55.574 [gcp-pubsub-subscriber2] INFO  com.example.Application - Message arrived! Payload: All work is done via SI.
```

Good run:

```
13:09:18.827 [http-nio-auto-1-exec-1] INFO  com.example.WorkService - starting busy work
13:09:19.408 [http-nio-auto-1-exec-2] INFO  com.example.ExampleController - meeting took 571ms
13:09:19.998 [http-nio-auto-1-exec-3] INFO  com.example.ExampleController - meeting took 528ms
[...]

PR #618, opened 2021-09-14
Disable SI trace instrumentation with Pub/Sub tracing

meltsufin:
There is little value in Spring Integration Sleuth support when you enable Pub/Sub trace instrumentation.

So, the recommended settings are:
```
spring.cloud.gcp.trace.pubsub.enabled=true
spring.sleuth.integration.enabled=false
```

This avoids the flakyness we observe with SI instrumentation.

This PR updates Trace sample to disable Spring Integration instrumentation, when Pub/Sub trace instrumentation is used.
Fixes flaky `TraceSampleApplicationIntegrationTests`.

Fixes: #597.

### DIFF
--- a/docs/src/main/asciidoc/trace.adoc
+++ b/docs/src/main/asciidoc/trace.adoc
@@ -164,7 +164,7 @@ It's set to `false` by default, but when set to `true`, trace spans will be crea
 # Enable Pub/Sub tracing using this property
 spring.cloud.gcp.trace.pubsub.enabled=true
 
-# You should disable Spring Integration instrumentation by Sleuth as it's unnecessary
+# You should disable Spring Integration instrumentation by Sleuth as it's unnecessary when Pub/Sub tracing is enabled
 spring.sleuth.integration.enabled=false
 ----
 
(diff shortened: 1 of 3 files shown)


docs/src/main/asciidoc/trace.adoc
```diff
@@ -164,7 +164,7 @@ It's set to `false` by default, but when set to `true`, trace spans will be crea
 # Enable Pub/Sub tracing using this property
 spring.cloud.gcp.trace.pubsub.enabled=true
 
-# You can disable Spring Integration instrumentation by Sleuth as it's unnecessary
+# You should disable Spring Integration instrumentation by Sleuth as it's unnecessary
```
suztomo, line 167: You may want to explain the relationship with `spring.cloud.gcp.trace.pubsub.enabled`

suztomo (approved):

docs/src/main/asciidoc/trace.adoc
meltsufin: Done.

Issue #593, opened 2021-08-24
Add support for JSON Spanner data type in Spring Data


elefeint:
Client library [6.12.0 release](https://github.com/googleapis/java-spanner/pull/1364) will have JSON support, as googleapis/java-spanner#872 was merged.

Add any necessary converter changes and a JSON field in the sample.

elefeint:
Fixed by #635.

PR #635, opened 2021-09-28
Add support for JSON Spanner data type in Spring Data Spanner

zhumin8:
To add support for JSON field in spring data spanner module. 
With these changes, a custom POJO field can be mapped as JSON spanner data type with annotation `@Column(spannerType = TypeCode.JSON)`. Supporting:
- create DDL
- read
- write
- sql query edge case: where query return is Json-field or list of Json-field.
This PR does not include support for Array of Json.

Corresponding issue: #593.
Creating draft PR as I am still working on the unit test for [added logic in SqlSpannerQuery.java](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/9e5554d26a06846ea33c25bc587a7aeb4d8f61ca/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQuery.java#L241-L249)


spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -339,6 +341,16 @@ private static boolean attemptSetIterableValue(Iterable<Object> value,
 		return true;
 	}
 
+
+	private static Value jsonToValueConverter(Object value) {
+		if (value == null) {
+			return Value.json(null);
+		}
+		Gson gson = new Gson();
```
meltsufin, line 349: This might have been discussed already, but I believe Spring has built-in conversion for JSON where it uses Jackson. Have you looked into trying to use that? Any cons to that?

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
```diff
@@ -170,4 +174,13 @@ private Class getSingleItemTypeCode(Type colType) {
 				? SpannerTypeMapper.getArrayJavaClassFor(colType.getArrayElementType().getCode())
 				: SpannerTypeMapper.getSimpleJavaClassFor(code);
 	}
+
+	public <T> T getSingleJsonValue(String colName, Class<T> colType) {
+		if (this.struct.isNull(colName)) {
+			return null;
+		}
+		String jsonString = this.struct.getJson(colName);
+		Gson gson = new Gson();
```
meltsufin, line 183: I don't think you need to create a `Gson` object on every call. You can make it a member variable. It's supposed to be thread-safe.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/admin/SpannerSchemaUtils.java
```diff
@@ -186,8 +186,13 @@ String getColumnDdlString(SpannerPersistentProperty spannerPersistentProperty,
 					spannerPersistentProperty.isGenerateSchemaNotNull(),
 					spannerPersistentProperty.isCommitTimestamp());
 		}
-		spannerJavaType = spannerEntityProcessor
+		if (spannerColumnType != null && spannerColumnType.equals(Type.Code.JSON)) {
```
meltsufin, line 189: Can this logic go into `spannerEntityProcessor.getCorrespondingSpannerJavaType`?

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
```diff
@@ -125,6 +126,9 @@ Object getSingleValue(String colName) {
 	}
 
 	public Object getSingleValue(int colIndex) {
+		if (this.struct.isNull(colIndex)) {
```
meltsufin, line 129: In what situation would this be `true`? Can you contrive a test for it?

meltsufin (commented):
First pass...
Spring Data Spanner is really hard! Nice work!

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
zhumin8: Opting to Gson just because it's already brought in through client libraries, no need to add new dependency.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
zhumin8: This is kind of unrelated to json support. But when I was testing, I found an edge case when the query result is a simple field, and return is null, it fails. Like this one. 
`@Query("SELECT name from test where id = @id")`
`Optional<String> getNameById(@Param("id") String id);`

I'll add it to a test somewhere.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
zhumin8: Good point. I'll change it.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -393,6 +405,12 @@ private void writeProperty(MultipleValueBinder sink,
 				valueSet = attemptSetSingleItemValue(Value.COMMIT_TIMESTAMP, Timestamp.class, valueBinder,
 						Timestamp.class, this.writeConverter);
 			}
+			// annotated json column, bind directly
+			else if (property.getAnnotatedColumnItemType() != null &&
+					property.getAnnotatedColumnItemType().equals(Type.Code.JSON)) {
+				valueBinder.to(jsonToValueConverter(propertyValue));
```
dzou, line 411: I think it may be preferable to keep all the stuff about the ".to() method" in one of the `attemptSetSingleValue` or `attemptSetSingleItemValue` method.

It would make the code much more parallel I think since it's the responsibility of one of those methods to invoke the correct `.to()` method on the binder. You can input the check in there on the property type if it's JSON to select the correct binder.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructPropertyValueProvider.java
```diff
@@ -104,6 +105,11 @@
 	private <T> T readSingleWithConversion(
 			SpannerPersistentProperty spannerPersistentProperty) {
 		String colName = spannerPersistentProperty.getColumnName();
+		Type.Code spannerColumnType = spannerPersistentProperty.getAnnotatedColumnItemType();
+		if (spannerColumnType != null && spannerColumnType.equals(Type.Code.JSON)) {
```
dzou, line 109: Same thing here, can be simplified to `spannerColumnType == Type.Code.JSON`.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -180,6 +183,11 @@ else if (property.isEagerInterleaved()) {
 										+ " in " + getType().getSimpleName() + ".");
 					});
 		}
+
+		if (property.getAnnotatedColumnItemType() != null
+				&& property.getAnnotatedColumnItemType().equals(Type.Code.JSON)) {
```
dzou, line 188: Can also simplify this enum equality check.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -96,6 +97,8 @@
 
 	private final String where;
 
+	private final Map<Class<?>, String> jsonPropertiesClassToName = new HashMap<>();
```
dzou, line 100: What is the purpose of this map again? Couldn't quite figure out based on the name.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -232,15 +233,33 @@ private List executeReadSql(Pageable pageable, Sort sort, QueryTagValue queryTag
...
-						spannerQueryOptions)
-				: this.spannerTemplate.query(this.entityType,
-						statement,
+		if (getReturnedSimpleConvertableItemType() != null) {
+			return this.spannerTemplate.query(
+					struct -> new StructAccessor(struct).getSingleValue(0), statement,
+					spannerQueryOptions);
+		}
+		// check if is json field
+		String jsonFieldName = getJsonFieldNameIfExists(returnedType);
```
dzou, line 242: Could you explain what this part of the code does? Looks like there is a way to query single column?

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
zhumin8: It's a map holding the class to field-name of Json annotated fields in the persistent entity. This gets produced once during the initializing process of persistent entity, so that in the added logic in `SqlSpannerQuery`, this information is accessible and does not need to loop over each persistent property to get if it's a json field. 

The logic added in `SqlSpannerQuery` is needed when the query returns a list or a single item of the Json field, not the whole entity or a simple field. This is an edge case that cannot be handled by the original logic, the code checks if the `returnedType` is one of the json field/s in the entity, if so, gets the corresponding field name from `jsonPropertiesClassToName` and proceed with `getSingleJsonValue` structAccessor function. 

Here is an [integration test](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/41be5213ed7e5bee7695f247604a2f8ae61b35a9/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/repository/it/SpannerRepositoryIntegrationTests.java#L531-L533) added for this edge case.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQuery.java
zhumin8: The part of the code determines how the query result get processed. If returned type is for example String or List<String>, versus Entity or List<Entity>. Now with JSON fields, there is a new possibility that the returned type is a custom POJO class that's annotated with json. 

This is an edge case that cannot be handled by the original logic, the code checks if the `returnedType` is one of the json field/s in the entity, if so, gets the corresponding field name from `jsonPropertiesClassToName` and proceed with `getSingleJsonValue` structAccessor function. 

Here is an [integration test](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/41be5213ed7e5bee7695f247604a2f8ae61b35a9/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/repository/it/SpannerRepositoryIntegrationTests.java#L531-L533) added for this edge case.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
zhumin8: Good point. I am adding `annotatedColumnItemType` to the parameter of `attemptSetSingleItemValue` since this one is private, so that the Json logic goes inside.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -393,6 +405,12 @@ private void writeProperty(MultipleValueBinder sink,
 				valueSet = attemptSetSingleItemValue(Value.COMMIT_TIMESTAMP, Timestamp.class, valueBinder,
 						Timestamp.class, this.writeConverter);
 			}
+			// annotated json column, bind directly
+			else if (property.getAnnotatedColumnItemType() != null &&
+					property.getAnnotatedColumnItemType().equals(Type.Code.JSON)) {
```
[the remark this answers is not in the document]
zhumin8: Ah, thanks for pointing out. and this saves one null check!

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/admin/SpannerSchemaUtils.java
zhumin8: I see that moving logic inside would be more readable. but I don't see a direct way of doing it, since `spannerEntityProcessor` have no access to either `spannerPersistentProperty` or `spannerPersistentEntity`, thus has no info about the annotation. And `getCorrespondingSpannerJavaType` is a public method, I don't know if I want to change and add parameter to it?

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
zhumin8: Added a test [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/0e4aae3dfba729a10295f12604937953646dbc66/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/repository/it/SpannerRepositoryIntegrationTests.java#L115-L120).

zhumin8: filed issue: #637.

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQueryTests.java
```diff
@@ -496,18 +504,97 @@ public void sqlCountWithWhereTest() throws NoSuchMethodException {
 			assertThat(paramMap.get("id").getString()).isEqualTo(params[0]);
 			assertThat(paramMap.get("trader_id").getString()).isEqualTo(params[1]);
 
-			return null;
-		}).when(this.spannerTemplate).executeQuery(any(), any());
+			return Arrays.asList(2L);
+		}).when(this.spannerTemplate).query((Function<Struct, Object>) any(), any(), any());
```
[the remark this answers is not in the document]
zhumin8: So this change is not because of the added json logic. I found when running this test, it is originally going through the wrong logic: in `executeReadSql`, because `sqlSpannerQuery.getReturnedSimpleConvertableItemType()` was not mocked to return `true`, it then takes the path to the query method returning the whole entity, which isn't the case in real sql count queries.  At the end of both routes, `this.spannerTemplate.executeQuery` get called with same parameters, so result as far as this test concerns is the same. But I wanted to make this change to reflect the actual code path that this query should take. 
That said, the changes here to verify on `query` instead of `executeQuery` is one layer up and not ideal. I made the compromise because the test breaks [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/206a2531a900ed4643483fc9341746a99ee43654/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/SpannerTemplate.java#L239-L243) without mocking `resultSet` correctly. 
In a newer commit, I am able to switch back to verify `executeQuery` with `resultSet` mocked.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -322,7 +326,12 @@ private static boolean attemptSetIterableValue(Iterable<Object> value,
 
 	@SuppressWarnings("unchecked")
 	private static <T> boolean attemptSetSingleItemValue(Object value, Class<?> sourceType,
-			ValueBinder<WriteBuilder> valueBinder, Class<T> targetType, SpannerCustomConverter writeConverter) {
+			ValueBinder<WriteBuilder> valueBinder, Class<T> targetType,
+			SpannerCustomConverter writeConverter, Type.Code annotatedColumnItemType) {
+		if (annotatedColumnItemType == Type.Code.JSON) {
```
dzou, line 331: Sorry, I think my previous advice about moving the JSON check logic here was not quite correct. I didn't realize you don't have access to the column type code in here, and then it doesn't seem quite right to have to pass `null` each time for the other usages.

I think we should go back to doing the `valueBinder.to(..)` back in the `writeProperty(..)` method, if you feel that way is simpler. Sorry to go back and forth.

elefeint:
@dzou Yes. However, we have R2DBC JSON support and Spring Boot 2.6 upgrades to do first.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
```diff
@@ -170,4 +176,12 @@ private Class getSingleItemTypeCode(Type colType) {
 				? SpannerTypeMapper.getArrayJavaClassFor(colType.getArrayElementType().getCode())
 				: SpannerTypeMapper.getSimpleJavaClassFor(code);
 	}
+
+	public <T> T getSingleJsonValue(String colName, Class<T> colType) {
+		if (this.struct.isNull(colName)) {
+			return null;
+		}
+		String jsonString = this.struct.getJson(colName);
```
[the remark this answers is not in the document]
elefeint: Discussed offline. Works intuitively for 90% of usecases.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -232,15 +233,33 @@ private List executeReadSql(Pageable pageable, Sort sort, QueryTagValue queryTag
...
+		if (getReturnedSimpleConvertableItemType() != null) {
+			return this.spannerTemplate.query(
+					struct -> new StructAccessor(struct).getSingleValue(0), statement,
+					spannerQueryOptions);
+		}
+		// check if returnedType is a field annotated as json
+		String jsonFieldName = getJsonFieldNameIfExists(returnedType);
+		if (jsonFieldName != null) {
+			return this.spannerTemplate.query(
+					struct -> new StructAccessor(struct).getSingleJsonValue(jsonFieldName, returnedType), statement,
```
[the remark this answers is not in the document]
zhumin8: >If the query return type is an object like Details, is there a requirement that the entity must use Details mapped to one of the fields?

hmm, good question. I am only concerned here for cases where the the returned object is mapped to one of the fields. Like in `SELECT detail from traders where id = @id` where `detail` is already a JSON field in `traders`. If user created some new JSON type field to return in the SQL, then we have no way of knowing what object to map it to.

>Why are we using StructAccessor? Wouldn't you need something like JsonAccessor?

`StructAccessor` is used to process any query result row returned from spanner. `Struct` here does not necessarily mean a STRUCT column in spanner.

>I would imagine that all you would need to determine whether to convert from JSON is to detect that the query returns a ResultSet from Spanner with a single column that is a Type.TYPE_JSON. What am I missing here?

Right, but unlike the case of [single value returns](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/eb88c823d6d436f7359f73be0469aa98c3498df8/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQuery.java#L236-L240) , I will also need to know the returnType to convert a Type.TYPE_JSON to. But you are right that I do not need `jsonFieldName` in this case. I am making changes to `SpannerPersistentEntityImpl.java` to only hold a set instead of a map for json properties. This will also resolve concerns for multiple JSON field of the same type.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -398,6 +405,11 @@ public boolean hasWhere() {
 		return Collections.unmodifiableSet(this.columnNames);
 	}
 
+	// Lookup whether a particular class is a JSON entity property
+	public String getJsonPropertyName(Class<?> type) {
```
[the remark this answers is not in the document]
zhumin8: Ah, thanks for pointing this out. On a second thought, keeping only a set of classes annotated with JSON is enough here. I am making the corresponding changes.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructPropertyValueProvider.java
```diff
@@ -104,6 +105,11 @@
 	private <T> T readSingleWithConversion(
 			SpannerPersistentProperty spannerPersistentProperty) {
 		String colName = spannerPersistentProperty.getColumnName();
+		Type.Code spannerColumnType = spannerPersistentProperty.getAnnotatedColumnItemType();
+		if (spannerColumnType == Type.Code.JSON) {
+			Object value = this.structAccessor.getSingleJsonValue(colName, spannerPersistentProperty.getType());
```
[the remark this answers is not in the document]
meltsufin: I think before JSON, we only had STRUCT type that could map to a user class. Maybe that's the reason? It's just a bit odd that JSON and STRUCT are conflated, but maybe STRUCT is more generic than I thought.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQuery.java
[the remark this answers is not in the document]
meltsufin: > hmm, good question. I am only concerned here for cases where the the returned object is mapped to one of the fields. Like in SELECT detail from traders where id = @id where detail is already a JSON field in traders. If user created some new JSON type field to return in the SQL, then we have no way of knowing what object to map it to.

Do you have information about the method signature, i.e. the return type of the method? That would be the type you need to map JSON too. I imagine the same problem would exist for STRUCT.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -232,15 +233,33 @@ private List executeReadSql(Pageable pageable, Sort sort, QueryTagValue queryTag
...
+		if (getReturnedSimpleConvertableItemType() != null) {
+			return this.spannerTemplate.query(
+					struct -> new StructAccessor(struct).getSingleValue(0), statement,
+					spannerQueryOptions);
+		}
+		// check if returnedType is a field annotated as json
+		boolean isJsonField = isJsonFieldType(returnedType);
+		if (isJsonField) {
+			return this.spannerTemplate.query(
+					struct -> new StructAccessor(struct).getSingleJsonValue(0, returnedType), statement,
```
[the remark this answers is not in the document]
meltsufin: Nice!

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQuery.java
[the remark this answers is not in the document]
elefeint: @meltsufin You are right that the type information is available in `struct.getType().getStructFields()`. But checking the struct metadata is not a pattern used in the code base right now -- we've used it extensively for R2DBC, but Spring Data integration is much older.

@zhumin8 If the type information from Spanner is inventoried and stored in `StructAccessor` in the same way that [column names currently are](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/b29c131b675ac479658ebdd9a56fb581be405a1d/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java#L159), would that be useful? On the other hand, I feel like the current way you are doing this might be more flexible, since even if the column is *stored* as JSON, the user might choose to *read* it as a `String` entity field anyway -- for example if the data needs to immediately be sent elsewhere.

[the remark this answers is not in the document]
meltsufin: > On the other hand, I feel like the current way you are doing this might be more flexible, since even if the column is stored as JSON, the user might choose to read it as a String entity field anyway -- for example if the data needs to immediately be sent elsewhere.

@elefeint Wouldn't the user express their intent to read as `String` by specifying the return type of the query method as `String`? We can use that to decide whether to parse JSON or not.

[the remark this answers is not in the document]
elefeint: That's what this PR does, no?

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/main/java/com/example/Trader.java
```diff
@@ -58,9 +58,6 @@ public class Trader {
 	@Column(spannerType = TypeCode.JSON)
 	private TraderDetails details;
```
Kusum2402: What data type to put for a dynamic JSON which has variable keys and how to handle it?

PR #638, opened 2021-10-01
Register ThreadPoolTaskSchedulers as beans 

### LINKED ISSUE #604: Register subscriber properties as beans
As a result of #562 the `ThreadPoolScheduler` is being directly instantiated. This requires us to manually take care of clean up and shutdown and also makes it difficult for it to be exposed to ExecutorBeanPostProcessor. 

As described below we need to look it if it is possible to have each of these subscription-specific settings be in the form of Spring-managed beans. 

I guess the problem here is that there would potentially be a separate instance of such a bean for each of the subscriptions. @mpeddada1 Can you look into how to keep these beans in the application context so that they can be exposed to `ExecutorBeanPostProcessor`?

_Originally posted by @meltsufin in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/562#discussion_r700519650_

mpeddada1:
Fixes #604

Uses `@PostConstruct` to create a global and subscription-specific threadPoolTaskSchedulers based on the executor-thread properties configured. We always create a global thread pool task scheduler bean (with the global config or default value if the configuration is not present) and only create a subscription-specific one if the subscription specific configuration is present.

After having registered threadPoolTaskSchedulers as beans, we no longer have to take care of manually shutting them down.


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/PubSubBeanProcessor.java
```diff
@@ -0,0 +1,144 @@
...
+	public PubSubBeanProcessor(ConfigurableEnvironment environment) {
+		this.environment = environment;
+	}
+
+	@Override
+	public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry beanDefinitionRegistry) throws BeansException {
+		Binder binder = new Binder(ConfigurationPropertySources.get(this.environment));
+		String cloudPropertiesPrefix = GcpProperties.class.getAnnotation(ConfigurationProperties.class)
+				.value();
+		GcpProperties gcpProperties = binder.bind(cloudPropertiesPrefix, GcpProperties.class)
```
mpeddada1, line 60: The unit test failure seems to be happening due to this because the GcpProjectIdProvider bean isn't accessible to this method.

mpeddada1: Fixed test my modifying test setup to specify projectId through application.properties. However, this seems to be a drawback of using `Binder` to determine the projectId in `BeanDefinitionRegistryPostProcessor`

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/PubSubBeanProcessor.java
```diff
@@ -0,0 +1,144 @@
...
+	 */
+	ThreadPoolTaskScheduler createThreadPoolTaskScheduler(Integer executorThreads, String threadName) {
+		ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
+		scheduler.setPoolSize(executorThreads);
+		scheduler.setThreadNamePrefix(threadName);
+		scheduler.setDaemon(true);
+		return scheduler;
+	}
+
+	public ConcurrentMap<String, ThreadPoolTaskScheduler> getThreadPoolSchedulerMap() {
```
elefeint, line 136: It's in the same package as the code that uses it, so it can definitely be package-private. Generally, returning collections is a tricky thing because Java collections are by definition mutable. We still do it, but if it can be done safer by restricting visibility (to package private in this case) or by only retrieving one keyed value at a time, it's generally worth it.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -187,15 +188,20 @@ private FlowControlSettings buildFlowControlSettings(
 	@Bean
 	@ConditionalOnMissingBean
 	public SubscriberFactory defaultSubscriberFactory(
+			@Qualifier("pubSubBeanProcessor") PubSubBeanProcessor pubSubBeanProcessor,
 			@Qualifier("subscriberExecutorProvider") Optional<ExecutorProvider> executorProvider,
 			@Qualifier("subscriberSystemExecutorProvider")
-			ObjectProvider<ExecutorProvider> systemExecutorProvider,
+					ObjectProvider<ExecutorProvider> systemExecutorProvider,
```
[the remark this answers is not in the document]
mpeddada1: Sorry about that, thanks!

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/PubSubBeanProcessor.java
```diff
@@ -0,0 +1,144 @@
...
+
+	private ThreadPoolTaskScheduler globalScheduler;
+
+	public PubSubBeanProcessor(ConfigurableEnvironment environment) {
+		this.environment = environment;
+	}
+
+	@Override
+	public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry beanDefinitionRegistry) throws BeansException {
+		Binder binder = new Binder(ConfigurationPropertySources.get(this.environment));
```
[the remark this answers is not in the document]
mpeddada1: Ah I wasn't able to find a way to get the binder object using the application context so I referred to using the [Binder API](https://spring.io/blog/2018/03/28/property-binding-in-spring-boot-2-0) directly instead. 

> There isn't a `Binder` object already available from the context? Alternatively, I was looking at Binder uses from Spring Boot, and it looks like [static singleton retrieval `.get()` method](https://github.com/spring-projects/spring-boot/blob/c6c139d980a50a1b39e564d44a1df6d46f8ca11b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/info/EnvironmentInfoContributor.java#L45) is common.

Thanks for the suggestion!

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/PubSubBeanProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+				: new DefaultGcpProjectIdProvider();
+		String cloudPubSubPropertiesPrefix = GcpPubSubProperties.class.getAnnotation(ConfigurationProperties.class)
+				.value();
+		GcpPubSubProperties pubSubProperties = binder.bind(cloudPubSubPropertiesPrefix, GcpPubSubProperties.class)
+				.orElse(new GcpPubSubProperties());
+		Integer globalExecutorThreads = pubSubProperties.getSubscriber().getExecutorThreads();
+		Integer numThreads = globalExecutorThreads != null ? globalExecutorThreads
+				: PubSubConfiguration.DEFAULT_EXECUTOR_THREADS;
+		this.globalScheduler = createAndRegisterSchedulerBean(numThreads, "global-gcp-pubsub-subscriber",
+				"globalThreadPoolScheduler", beanDefinitionRegistry);
```
elefeint, line 72: `globalThreadPoolScheduler` should mention `Pubsub`. Are we going to have a different one for publishing?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/PubSubBeanProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	@Override
+	public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry beanDefinitionRegistry) throws BeansException {
+		Binder binder = Binder.get(this.environment);
+		String cloudPropertiesPrefix = GcpProperties.class.getAnnotation(ConfigurationProperties.class)
+				.value();
+		GcpProperties gcpProperties = binder.bind(cloudPropertiesPrefix, GcpProperties.class)
+				.orElse(new GcpProperties());
+		GcpProjectIdProvider projectIdProvider = (gcpProperties.getProjectId() != null)
```
elefeint, line 61: Slight wrinkle: This should actually be coming from  `spring.cloud.gcp.pubsub.project-id`, *or* [the default provider](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/62389625dc205ce3e431967249877c9c0f1fa600/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/core/GcpContextAutoConfiguration.java#L73).

Take another look at [pubsub autoconfiguration](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/0def25d2cac17a5045c0f8d4c4dd9d15059622aa/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java#L102) -- it's confusing, but the order of precedence is:
* product-specific property `spring.cloud.gcp.pubsub.project-id`
* global property `spring.cloud.gcp.project-id`
* default from environment

I'd separate the logic for determining project ID into a helper method here, comment in javadoc why the duplication is needed, *and* possibly even add a few unit tests for autoconfiguration verifying that this project ID logic will get the same result as the regular autoconfiguration. The reason for the unit tests is that we'll need them to fail to signal that the two project ID sets of logic have diverged. Otherwise we'll never remember to keep them in sync because they Almost Never Change.

elefeint (changes requested):

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/PubSubBeanProcessor.java
meltsufin: Is there any way to have this post processor run after Spring has already processed `GcpPubSubAutoConfiguration`? Maybe you can create another auto-configuration class for the post-processor where you do `@AutoConfigureAfter(GcpPubSubAutoConfiguration.class)`? I'm not sure if it would work, but it would nice not have to duplicate this logic.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/PubSubBeanProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProcessor;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.bind.Binder;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+
+/**
+ * Class to register ThreadPoolTaskScheduler beans.
+ */
+public class PubSubBeanProcessor implements BeanDefinitionRegistryPostProcessor {
```
meltsufin, line 42: Have you tried using `BeanFactoryPostProcessor` instead? It might be more flexible because AFAIK it gives you access to the already-created beans.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/PubSubBeanProcessor.java
mpeddada1: > Slight wrinkle: This should actually be coming from `spring.cloud.gcp.pubsub.project-id`, _or_ [the default provider](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/62389625dc205ce3e431967249877c9c0f1fa600/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/core/GcpContextAutoConfiguration.java#L73).

Ah you're right. There is a lot more logic involved here which unfortunately may have to be duplicated.

> Is there any way to have this post processor run after Spring has already processed `GcpPubSubAutoConfiguration`

That's an interesting thought! Unfortunately, `@AutoConfigureAfter(GcpPubSubAutoConfiguration.class)` still results in the BeanDefinitionRegistryPostProcessor occurring before any of the configurations. The order looks like: BDRPP ->  GcpContextAutoConfiguration --> GcpPubSubAutoConfiguration --> "Configuration annotated with AfterConfigureAfter"

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/PubSubBeanProcessor.java
meltsufin: You're right. I take that back.
What about `@PostConstruct` method in `GcpPubSubAutoConfiguration`. Can you add beans to the context there?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -96,6 +104,13 @@
 
 	private final HeaderProvider headerProvider = new UserAgentHeaderProvider(this.getClass());
 
+	private final ConcurrentHashMap<String, ThreadPoolTaskScheduler> threadPoolTaskSchedulerMap = new ConcurrentHashMap<>();
+
+	@Autowired
```
elefeint, line 109: Let's do construction injection into `GcpPubSubAutoConfiguration()` constructor instead.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -96,6 +104,13 @@
...
+	ApplicationContext context;
```
elefeint, line 110: private?

mpeddada1:
> So half-answer my own question (would still like empirical evidence), this should work. [Spring documentation](https://docs.spring.io/spring-framework/docs/5.3.x/reference/html/core.html#beans-factory-extension-bpp) says that post-processor callbacks happen both before and after individual object lifecycle callbacks. And looking at [ExecutorBeanPostProcessor](https://github.com/spring-cloud/spring-cloud-sleuth/blob/3dc8387a6fe2e85d8f147e3018b4357ae357a5bc/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/async/ExecutorBeanPostProcessor.java#L57), the logic is implemented in the "after".
> 
> (also useful, although not directly relevant right now is [this section](https://docs.spring.io/spring-framework/docs/5.3.x/reference/html/core.html#beans-factory-lifecycle-combined-effects) on combining different lifecycle mechanisms for an object)

Thank you for verifying this, Elena! I created a BeanPostProcessor bean in GcpPubSubAutoConfiguration and double checked (with the help of `if bean instanceOf ThreadPoolTaskScheduler`) to make sure that the global and subscription-specific ThreadPoolTaskScheduler beans are accessible to the BeanPostProcessor#postProcessAfterInitialization.

Issue #671, opened 2021-10-25
Follow-up: Register rest of the subscriber settings as beans in GcpPubSubAutoconfiguration


mpeddada1:
Similar to what was done in #638, we should register the remaining subscriber settings as beans:

- [x] retry settings
- [x] flow control settings
- [x] executor provider

PR #690, opened 2021-11-10
Register subscriber flow control settings as beans

mpeddada1:
Part 1 of #671 

This PR registers global and subscription-specific flow control settings as beans in `GcpPubSubAutoConfiguration`. However, note that if a custom bean is provided then the global and subscription-specific beans won't be created.


elefeint (changes requested):
Looks good. 
Could you also add a test in autoconfiguration to make sure that if no user-provided settings exist (neither either global or subscription specific) that Retry settings default to client library's `Subscriber.getDefaultFlowControlSettings()`? I manually verified that, but we should have a test. We had caused a production issue last summer where we were not falling back correctly to client library values.

elefeint:
You'd have to actually `createSubscriber()` to validate this, since the logic is in client library building the stub, and not in this project.

mpeddada1:
> Could you also add a test in autoconfiguration to make sure that if no user-provided settings exist (neither either global or subscription specific) that Retry settings default to client library's `Subscriber.getDefaultFlowControlSettings()`?

Yes, that is a good idea!

elefeint (approved):
Thank you!

PR #691, opened 2021-11-11
GH-689: add support for Google Cloud Storage based distributed lock 

fireboy-will:
This PR provides an implementation (GcsLock) for https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/689

Signed-off-by: Williams Simon <openbites.org@gmail.com>


elefeint:
I am sorry; we are not ready to accept this contribution.
Spring Cloud GCP's goal is to make common user journeys seamless for Spring developers. Using GCS as a distributed lock is a niche requirement that is best left as an experimental or a third party library.

PR #696, opened 2021-11-12
Migrating one test from JUnit4 to JUnit5 as a sample.

ddixit14:
Migrated _SpannerDatabaseAdminTemplateTests_ from JUnit4 to JUnit5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,13 +146,6 @@
 				<artifactId>spring-aot</artifactId>
 				<version>${spring-native.version}</version>
 			</dependency>
-			<dependency>
-				<groupId>org.junit</groupId>
-				<artifactId>junit-bom</artifactId>
-				<version>5.8.1</version>
-				<type>pom</type>
-				<scope>import</scope>
-			</dependency>
 		</dependencies>
 	</dependencyManagement>
 
@@ -163,19 +156,16 @@
 			<version>4.1.1</version>
 			<scope>test</scope>
 		</dependency>
-
-		<dependency>
-			<groupId>org.junit.platform</groupId>
-			<artifactId>junit-platform-launcher</artifactId>
-			<scope>test</scope>
-		</dependency>
-
 		<dependency>
 			<groupId>org.junit.jupiter</groupId>
 			<artifactId>junit-jupiter-api</artifactId>
 			<scope>test</scope>
 		</dependency>
-
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+			<scope>test</scope>
+		</dependency>
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
@@ -185,11 +175,6 @@
 		<dependency>
 			<groupId>org.junit.vintage</groupId>
 			<artifactId>junit-vintage-engine</artifactId>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>org.junit.jupiter</groupId>
-			<artifactId>junit-jupiter-engine</artifactId>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
@@ -204,11 +189,6 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>org.junit.jupiter</groupId>
-			<artifactId>junit-jupiter-engine</artifactId>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -146,7 +146,13 @@
 				<artifactId>spring-aot</artifactId>
 				<version>${spring-native.version}</version>
 			</dependency>
-
+			<dependency>
+				<groupId>org.junit</groupId>
+				<artifactId>junit-bom</artifactId>
+				<version>5.8.1</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
 		</dependencies>
 	</dependencyManagement>
 
@@ -157,6 +163,19 @@
 			<version>4.1.1</version>
 			<scope>test</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>org.junit.platform</groupId>
+			<artifactId>junit-platform-launcher</artifactId>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-api</artifactId>
+			<scope>test</scope>
+		</dependency>
+
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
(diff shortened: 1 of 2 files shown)


suztomo:
Do you see the test ran in the output?

ddixit14:
Yes, the tests are running fine, none getting ignored.

ddixit14:
@elefeint Please let me know if this looks good so I that can proceed with other tests.

pom.xml
```diff
@@ -168,6 +187,11 @@
 			<artifactId>junit-vintage-engine</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+			<scope>test</scope>
+		</dependency>
```
elefeint, line 194: I'd probably move this block up, so all Jupiter dependencies are grouped together.

elefeint (changes requested):
Looks good, just minor comments.
While you are there, could you remove this comment in pom.xml -- ""? It has been untrue since [last October](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/64) when we separated our parent pom from spring-cloud-build.

pom.xml
```diff
@@ -157,6 +163,19 @@
 			<version>4.1.1</version>
 			<scope>test</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>org.junit.platform</groupId>
+			<artifactId>junit-platform-launcher</artifactId>
```
meltsufin, line 169: Do you really need this dependency? I don't think we run tests outside of Maven or the IDE.

pom.xml
```diff
@@ -146,7 +146,13 @@
 				<artifactId>spring-aot</artifactId>
 				<version>${spring-native.version}</version>
 			</dependency>
-
+			<dependency>
+				<groupId>org.junit</groupId>
+				<artifactId>junit-bom</artifactId>
```
meltsufin, line 151: I think this is already imported via:
`spring-cloud-build-dependencies` -> `spring-boot-dependencies`

pom.xml
```diff
@@ -146,7 +146,13 @@
...
+				<version>5.8.1</version>
```
meltsufin, line 152: We should try to keep all versions in the properties above.

pom.xml
```diff
@@ -168,6 +187,11 @@
 			<artifactId>junit-vintage-engine</artifactId>
```
meltsufin, line 187: Wouldn't you need to remove this dependencies to not rely on JUnit 4?

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/admin/SpannerDatabaseAdminTemplateTests.java
```diff
@@ -64,7 +64,7 @@
 
 	private List<String> ddlList;
 
-	@Before
+	@BeforeEach
 	public void setup() {
```
chanseokoh, line 68: FYI: AFAIK, all `@Test` (and perhaps this one too) can be `private` now.

ddixit14: Yeah I studied that in the official JUnit5 guide but I wasn't so sure about it because I couldn't find a good technical reason to change the methods visibility. 
They've suggested to keep the methods public by default because of two reasons, one being the possibility of a test class being extended by test class in another package and the other to simplify testing on the module path when using the Java Module System. 
Please let me know what you think about this, I'd love to have a new perspective about it.

elefeint: Package private is the default I've seen most commonly -- it's less visible (and less verbose!) than `public`, but it still lets other classes in the same package extend.

pom.xml
ddixit14: I think we still have a lot of tests running on JUnit4 so it wouldn't make sense to remove this. To ensure that the tests have been successfully migrated to JUnit5, for every test after migration, I remove this dependency from the code and run the test. 
Does this make sense or is there a better way to go about it?

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/admin/SpannerDatabaseAdminTemplateTests.java
ddixit14: I'll have a look if we have any extensions of classes/methods outside packages in the code. If we don't on a majority, it might make sense to go classes/methods being package-private, right ?

pom.xml
meltsufin: Sure, as  long as there is a way to tell that the migrated tests are running with JUnit 5, and not 4.

elefeint: That's a good validation! To make extra sure there is no JUnit4 left in the project, you can check with `mvn dependency:tree -Dincludes=junit:junit`. I think you'd have to remove both `junit:junit` and `org.junit.vintage:junit-vintage-engine`, since the latter has a transitive dependency on the former.

pom.xml
elefeint: That's right; this block can be removed.

pom.xml
elefeint: We don't need this one; it can be removed.

There is an inconsistency between JUnit's [list of required dependencies](https://junit.org/junit5/docs/current/user-guide/#running-tests-junit-platform-runner-setup), which has 4 required dependencies for running Junit4 and 5 in the same project, and [Surefire's](https://maven.apache.org/surefire/maven-surefire-plugin/examples/junit-platform.html), which only has two dependencies.

Surefire's seems to work fine from both command line and IDE. *shrug*

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/admin/SpannerDatabaseAdminTemplateTests.java
elefeint: Yes, let's stick to package-private default for everything where it does not cause a problem.

elefeint: @ddixit14 By the way, if you take a look at the code smells (Sonar check), you'll see that it advises to remove "public" modifier. I like reading Sonar's "why is this an issue" explanations -- most of the time I agree; some of the time I don't, but I almost always learn something.

elefeint: @chanseokoh for example, I just learned (had no idea!) that JUnit5 runner will [silently ignore private methods](https://sonarcloud.io/organizations/googlecloudplatform/rules?open=java%3AS5810&rule_key=java%3AS5810).

pom.xml
elefeint: Okay, to answer the original question -- `org.junit.vintage:junit-vintage-engine` uses `junit:junit` for actual implementation. From [docs](https://junit.org/junit5/docs/current/user-guide/#running-tests-build-maven-engines-configure): "Maven Surefire and Maven Failsafe can run JUnit 4 based tests alongside Jupiter tests as long as you configure test scoped dependencies on JUnit 4 and the JUnit Vintage TestEngine".

So we have to keep `junit:junit` dependency until we have migrated all Junit4 tests to Junit5.

elefeint (approved):

Issue #563, opened 2021-08-06
Expose StackdriverSender.serverResponseTimeoutMs via a config property


meltsufin:
Follow-up to https://github.com/spring-cloud/spring-cloud-gcp/issues/1617#issuecomment-894200998.

We need to add `serverResponseTimeoutMs` to `GcpTraceProperties` and update the `StackdriverTraceAutoConfiguration.stackdriverSender()` to use it.

mvpzone:
Proposed [PR #698](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/698) to address this issue.

elefeint:
Fixed in #698. Thanks, @mvpzone !

PR #698, opened 2021-11-14
Expose StackdriverSender.serverResponseTimeoutMs via a config property.

mvpzone:
Allows configuring the server response waiting timeout.

This PR addresses  https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/563
See also: https://github.com/spring-cloud/spring-cloud-gcp/issues/1617

### DIFF
--- a/docs/src/main/asciidoc/trace.adoc
+++ b/docs/src/main/asciidoc/trace.adoc
@@ -80,7 +80,7 @@ All configurations are optional:
 | `spring.cloud.gcp.trace.max-outbound-size` | Maximum size for outbound messages | No |
 | `spring.cloud.gcp.trace.wait-for-ready` | https://github.com/grpc/grpc/blob/main/doc/wait-for-ready.md[Waits for the channel to be ready] in case of a transient failure | No | `false`
 | `spring.cloud.gcp.trace.messageTimeout` | Timeout in seconds before pending spans will be sent in batches to GCP Cloud Trace. (previously `spring.zipkin.messageTimeout`) | No | 1
-| `spring.cloud.gcp.trace.serverResponseTimeoutMs` | Server response timeout in millis. | No | `5000`
+| `spring.cloud.gcp.trace.server-response-timeout-ms` | Server response timeout in millis. | No | `5000`
 | `spring.cloud.gcp.trace.pubsub.enabled` | (Experimental) Auto-configure Pub/Sub instrumentation for Trace. | No | `false`
 |===
 
--- a/docs/src/main/asciidoc/trace.adoc
+++ b/docs/src/main/asciidoc/trace.adoc
@@ -80,6 +80,7 @@ All configurations are optional:
 | `spring.cloud.gcp.trace.max-outbound-size` | Maximum size for outbound messages | No |
 | `spring.cloud.gcp.trace.wait-for-ready` | https://github.com/grpc/grpc/blob/main/doc/wait-for-ready.md[Waits for the channel to be ready] in case of a transient failure | No | `false`
 | `spring.cloud.gcp.trace.messageTimeout` | Timeout in seconds before pending spans will be sent in batches to GCP Cloud Trace. (previously `spring.zipkin.messageTimeout`) | No | 1
+| `spring.cloud.gcp.trace.serverResponseTimeoutMs` | Server response timeout in millis. | No | `5000`
 | `spring.cloud.gcp.trace.pubsub.enabled` | (Experimental) Auto-configure Pub/Sub instrumentation for Trace. | No | `false`
 |===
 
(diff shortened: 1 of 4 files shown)


elefeint (changes requested):
Could you add a test in `StackdriverTraceAutoConfigurationTests` and documentation in `trace.adoc`?
Otherwise looks good.

mvpzone:
> Could you add a test in `StackdriverTraceAutoConfigurationTests` and documentation in `trace.adoc`?
> Otherwise looks good.

Adjusted trace doc to include the new property. Assessing test change.

docs/src/main/asciidoc/trace.adoc
```diff
@@ -80,6 +80,7 @@ All configurations are optional:
 | `spring.cloud.gcp.trace.max-outbound-size` | Maximum size for outbound messages | No |
 | `spring.cloud.gcp.trace.wait-for-ready` | https://github.com/grpc/grpc/blob/main/doc/wait-for-ready.md[Waits for the channel to be ready] in case of a transient failure | No | `false`
 | `spring.cloud.gcp.trace.messageTimeout` | Timeout in seconds before pending spans will be sent in batches to GCP Cloud Trace. (previously `spring.zipkin.messageTimeout`) | No | 1
+| `spring.cloud.gcp.trace.serverResponseTimeoutMs` | Server response timeout in millis. | No | `5000`
```
meltsufin, line 83:
```suggestion
| `spring.cloud.gcp.trace.server-response-timeout-ms` | Server response timeout in millis. | No | `5000`
```

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/GcpTraceProperties.java
```diff
@@ -86,6 +86,11 @@
 	 */
 	private int messageTimeout = 1;
 
+	/**
+	 * Timeout in seconds for server response. Ref: https://github.com/spring-cloud/spring-cloud-gcp/issues/1617.
```
meltsufin, line 90: I'm not sure we should have any URLs in the property description.
```suggestion
	 * Timeout in seconds for server response when sending traces.
```

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/GcpTraceProperties.java
```diff
@@ -86,6 +86,11 @@
...
+	 */
+	private Long serverResponseTimeoutMs = 5000L;
```
meltsufin, line 92: Can we just leave it `null` by default so that we inherit the default from `Sender`?
It will be the same default.

meltsufin (changes requested):
Would it be possible to also add some tests. Maybe in `TracePubSubAutoConfigurationTest`?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/GcpTraceProperties.java
mvpzone: Changing to null breaks a load of tests (NPE due to the null). So leaving as 5000L for now. Assessing a sane default.

mvpzone:
> ...add a test in `StackdriverTraceAutoConfigurationTests`?

Done.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/GcpTraceProperties.java
mvpzone: From StackDriverSender the default is set to zipkin2.reporter.stackdriver.internal.UnaryClientCall.DEFAULT_SERVER_TIMEOUT_MS (5000L). So proposing to set the default to this constant to avoid NPE in setting up the default sender bean. A value of zero will also fail since builder validates for a timeout greater than zero.

mvpzone: @meltsufin This does require an import from zipkin2 package What do you think?

mvpzone:
> Would it be possible to also add some tests. Maybe in `TracePubSubAutoConfigurationTest`?

Added tests to StackdriverTraceAutoConfigurationTests instead.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/GcpTraceProperties.java
meltsufin: Can you just skip setting the value on `StackdriverSender.Builder` if the property is `null`?

meltsufin (commented):
Thanks for the tests! It's looking pretty good. Just one comment about leaving it `null`.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/GcpTraceProperties.java
mvpzone: +1 .Done

elefeint (changes requested):

meltsufin (approved):

elefeint (approved):

Issue #707, opened 2021-11-16
`PubSubMessageChannelBinderEmulatorIntegrationTests` ignoring `-Dit.pubsub-emulator` settings


zhumin8:
`PubSubMessageChannelBinderEmulatorIntegrationTests` ignoring `-Dit.pubsub-emulator` settings and failing this test.
This results in integration tests for _all modules_ running this test.

Suspect Cause: mix use of junit4 and 5.
the abstract test class `AbstractBinderTests` that it extends from spring-cloud-stream-binder uses junit 5 now.

Tasks:
- [x] fix junit4/5 mis-match and `-Dit.pubsub-emulator` settings is honored when running integration tests.
- [x] fix `PubSubMessageChannelBinderEmulatorIntegrationTests`  failures.

PR #708, opened 2021-11-16
Unblock integration tests: fix to respect `-Dit.pubsub-emulator` settings

zhumin8:
This is PR brings `PubSubMessageChannelBinderEmulatorIntegrationTests` to junit 5 to avoid mis-match with the abstract test class it inherit from. This test has been running with all modules's integration tests regardless of the `-Dit.pubsub-emulator` settings.
This is step 1 of #707 and should unblock and reveal any further issues on it test for other modules.


elefeint (approved):
I've also filed #710 to eventually replace the test rule with a testcontainer (a dependency running in an isolated container that's created and torn down for each test). It's not urgent, but if the PubSubEmulator rule gives you trouble, taking it out and replacing with a testcontainer might be easier.

PR #713, opened 2021-11-17
temp fix to get away from circular reference error

zhumin8:
This is a temporarily fix to let datastore basic sample work with spring boot 2.6.
circular reference is on class `org.springframework.shell.result.ResultHandlerConfig` ([link to error](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/runs/4239959410?check_suite_focus=true))

Suspect cause:
This sample uses `spring-shell` for cmd line interaction. [Spring shell](https://spring.io/projects/spring-shell#learn) seems to be a project relative slow in updating. We are already on the newest version `2.0.1.RELEASE` which has dependencies on `spring boot 1.5.8`.

For longer term, should we rewrite this sample to move away from spring shell?


elefeint (approved):
Yes; it would be good to not have one-off shell sample apps (Spanner template sample is also command-line based).
Having a simple web application allows more interactivity with the user; it's clear which functionality is getting invoked.

zhumin8:
oh, this datastore-basic-sample is actually interactive: it has a bunch of commands that user can interactively save and find books. That's why it has spring shell dependency.

zhumin8:
Root cause for this circular reference error is fixed here:
https://github.com/spring-projects/spring-shell/pull/327
[v3.0.0-M1](https://github.com/spring-projects/spring-shell/releases/tag/v3.0.0-M1) is out recently. We should pick up in next release.

PR #716, opened 2021-11-18
spring cloud gcp data firestore tests migration(JUnit4 to JUnit5)

ddixit14:
Migrated the following tests from JUnit4 to JUnit5 :

**module: spring-cloud-gcp-data-firestore**

1. FirestorePersistentEntityImplTests.java
2. PartTreeFirestoreQueryTests.java
3. FirestoreTemplateTests.java
4. SimpleFirestoreReactiveRepositoryTests.java
5. UtilTests.java

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
+++ b/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
@@ -38,8 +38,8 @@ import com.google.firestore.v1.StructuredQuery;
 import com.google.firestore.v1.Value;
 import com.google.firestore.v1.Write;
 import io.grpc.stub.StreamObserver;
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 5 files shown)


elefeint (changes requested):
Very useful PR descriptions in the last couple of PRs! The description for this PR explained the intent to me, so it was easier to understand the file diff.

This PR actually contains files from both, Firestore and Spanner because your local branch contains both of those, and `main` branch contains neither. You'll find that it's easier to work in separate new branches for each PR. It's a common approach when working with github; in fact, this project automatically deletes the remote version of the branch once the commits in it are merged into `main`. You will likely need to do an interactive rebase of your branch from the latest `main`; we can do it during the code party.

One more thing about branch naming -- it's simpler when there aren't special characters like spaces or parentheses. It's not a big deal (it's easy enough to add single quotes around the branch name when working from command line), but it's less prone to weird edge cases with tooling.

elefeint:
I looked closer, and realized that you _are_ using different branches. In this case, fixing this is easy:
`git rebase main`.
It will take the `main` branch as it is right now, and add your commits to the Firestore branch on top.

elefeint (approved):
Could you also update the title of the pull request, so it's easy to see which module's tests got migrated?

ddixit14:
> Could you also update the title of the pull request, so it's easy to see which module's tests got migrated?

Done!

PR #718, opened 2021-11-18
Fix secret manager it test setup to work with spring boot 2.6

zhumin8:
The test were failing at value injection to TestConfiguration due to missing converter. This failure only affects integration test as its setup does not rely on auto-configuration.
 
This change explicitly provides SecretManagerTestConfiguration as source when building context so that the custom converter gets picked up.


elefeint:
@dzou Could you take a look? This fixes the problem, but I don't understand _why this worked in the first place_.

elefeint (approved):

PR #723, opened 2021-11-19
spring cloud gcp autoconfigure tests migration(JUnit4 to JUnit5)

ddixit14:


elefeint (changes requested):
Take a look at the Sonar result -- it will look familiar.

ddixit14:
"StackdriverTraceAutoConfigurationTests" class is actually used in autoconfigure/trace/pubsub/TracePubSubAutoConfigurationTest.java which is outside the package. So removing the public access modifier and making it package private will not work here.

elefeint (changes requested):
That's what's called a "code smell". If  `MockConfiguration` is embedded in one tests in package A, tests from package B really should not be referring to it. It's the equivalent of me reaching into a shopping bag that a neighbor left on their porch and borrowing a cup of flour -- clearly physically possible (since they did temporarily leave it out on the porch!) but just as clearly breaking social norms and private property encapsulation boundaries.

I would recommend making MockConfiguration a standalone public class that's in its own file. Then both integration tests can refer to it without breaking their encapsulation.

elefeint (approved):

PR #728, opened 2021-11-19
Update documentation for per-subscription configuration support

mpeddada1:


elefeint (changes requested):
About "If the per-subscription configuration for this field is not provided then the global configuration is picked, otherwise, the default is picked": a few properties (like `parallel-pull-count` did not have this wording) -- does this indicate a difference in actual behavior?

Either way, I think it would be better to place this as a "NOTE:" before each of the two blocks of properties, rather than repeating it in each property.

mpeddada1:
> "If the per-subscription configuration for this field is not provided then the global configuration is picked, otherwise, the default is picked"

Thanks for catching this! This applies to all the settings so a "NOTE" will be better suited for this point.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -86,6 +86,23 @@ Example: `"us-east1-pubsub.googleapis.com:443"`.
 This is useful in conjunction with enabling message ordering because sending messages to the same region ensures they are received in order even when multiple publishers are used. | No | pubsub.googleapis.com:443
 |===
 
+===== Subscription-specific Configurations
+Subscriber settings can also be set on a per-subscription basis. A custom configuration (set through a setter in `DefaultSubscriberFactor` or a custom bean) will take precedence over auto-configuration. When using auto-configuration, if one wishes to use per-subscription configuration for a property then they need to avoid using a custom bean for that property. If both global and per-subscription configurations are provided then the per-subscription configuration will be used. However, if subscription-specific property is not specified then the global or default configuration will be used.
```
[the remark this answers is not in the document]
mpeddada1: Thank you! Applied the suggestions specified here.

elefeint (approved):

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -55,7 +55,9 @@ Pub/Sub credentials | No | https://www.googleapis.com/auth/pubsub
 
 ==== Publisher/Subscriber Configuration
 
-This section describes configuration options to customize the behavior of the application's Pub/Sub publishers and subscribers.
+This section describes configuration options to customize the behavior of the application's Pub/Sub publishers and subscribers. Subscriber settings can be either global or subscription-specific.
```
meltsufin, line 58: We used to follow the one-sentence-per-line convention. I don't know if we still want to continue with it after Spring conventions are no longer mandatory.

meltsufin (approved):

docs/src/main/asciidoc/pubsub.adoc
elefeint: One-sentence-per-line has grown on me over time: it does make it easier to review changes, since it does not lead to line reformatting whenever a sentence length changes (also, we don't have the internal tools automatically reformatting to 80 chars in open source world).

mpeddada1: Done, applied the one-sentence-per-line convention.

PR #730, opened 2021-11-20
Fixes for`PubSubMessageChannelBinderEmulatorIntegrationTests`

### LINKED ISSUE #707: `PubSubMessageChannelBinderEmulatorIntegrationTests` ignoring `-Dit.pubsub-emulator` settings

zhumin8:
This is step 2 and fixes #707 

There are 2 fixes in this PR: 
1. Let `PubSubEmulator` extend `ParameterResolver` to inject the emulator port number into the instance. This is left-over from task 1 upgrading to junit5 test, solves the `java.lang.NullPointerException: target` error caused by no port number provided.
2. After above change, tests fail due to a [recent change](https://github.com/spring-cloud/spring-cloud-stream/commit/2c47c6de8977345e0595ba122adf89c2709c41d6#diff-70aeaabb8c7de9d4d4b239ba643f594bf49f111d18569056dd93a835d5e9867cR132) in spring-cloud-stream logic. 
    - when checking if context already has an ObjectMapper bean, calls `getBeansOfType(ObjectMapper.class)`, which has an internal call to `assertBeanFactoryActive()`
    - Test setup fails with error: `“org.springframework.context.support.GenericApplicationContexthas not been refreshed yet”`.
    - Fix by `refresh` the context.

Both fixes are within the test setup.

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubEmulator.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubEmulator.java
@@ -102,19 +102,19 @@ public class PubSubEmulator implements BeforeAllCallback, AfterAllCallback, Para
 	}
 
 	/**
-	 * Set up ParameterResolver to support PubSubEmulator as parameter type.
+	 * Set up ParameterResolver to support String as parameter type.
 	 */
 	@Override
 	public boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
-		return parameterContext.getParameter().getType() == PubSubEmulator.class;
+		return parameterContext.getParameter().getType() == String.class;
 	}
 
 	/**
-	 * Set up ParameterResolver to return current instance of PubSubEmulator to test.
+	 * Set up ParameterResolver to return current emulatorHostPort to test.
 	 */
 	@Override
-	public PubSubEmulator resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
-		return this;
+	public String resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
+		return this.emulatorHostPort;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -53,7 +54,7 @@
  *
  * @since 1.1
  */
-public class PubSubEmulator implements BeforeAllCallback, AfterAllCallback {
+public class PubSubEmulator implements BeforeAllCallback, AfterAllCallback, TestInstancePostProcessor {
```
elefeint, line 57: I think it might be simpler to instead implement `ParameterResolver`, and return _itself_ as an instance of `PubSubEmulator`. Then the test can accept `PubSubEmulator` as a constructor argument and continue to function without any new functionality (emulator instance in  used to be static, but I don't think it needs to be anymore now that it's not a class rule).

elefeint (changes requested):
I have a suggestion that will make it unnecessary to mess with reflection (great in a pinch, but harder to read/maintain and not type safe).
If we stick with test post processing, then doing a check for the method existing before acquiring it would be necessary.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubEmulator.java
zhumin8: Hmm, good point. Let me try this.

elefeint (approved):

PR #729, opened 2021-11-19
Implement datastore Blob to byte[] conversion on read

elefeint:
Blob-to-byte[] and reverse are a special case in conversion -- it's the only exception to the general rule of "Datastore collection becomes Java collection; Datastore simple type becomes Java type". It seems to have been intended to work in original implementation, but never had a test, so either never worked right or broke over the years. The other direction, on write, works fine through `BYTE_ARRAY_TO_BLOB_CONVERTER` combined with `ValueUtil.isArrayOfItems()` exempting `byte[]` from array status. 

There were several ways of fixing this (each commit in this PR is a working solution), but at the end I went with making the standard conversion path worked. It required redefining the meaning of collection to exclude byte arrays, which is already done in the local utility `ValueUtil.isArrayOfItems()`. I replaced Spring Data's built-in `TypeDiscoverer.isCollectionLike()`, which (reasonably) assumes byte arrays are collection-like with the Spring Cloud Gcp Datastore implementation-specific  `ValueUtil.isArrayOfItems()`, which excludes `byte[]`. This exclusion is already documented in the refdoc, so no additional documentation is needed.

Fixes #661.

PR #284, opened 2021-02-10
Unwind nested ternary statements

ttomsu:
[Ternary statements shouldn't be nested](https://sonarcloud.io/organizations/googlecloudplatform/rules?open=java%3AS3358&rule_key=java%3AS3358) x 4 (major smells)

PR #733, opened 2021-11-22
Datastore: remove unused code handling array conversion

elefeint:
**I want a second opinion on this. TLDR: code handling array-to-list conversion broke a year ago and nobody noticed. We can either (1) remove this code entirely (and document it as unsupported) or (2) fix the array-to-list conversion to the way it was before #284 and add a test.** In an ideal world, I'd prefer a type-safe conversion interface that has separate methods to convert single values and collections, but we don't have bandwidth for refactoring what's not technically broken.

When working on #729, this block in `convertOnRead(Object, EmbeddedType, TypeInformation)` was bothering me -- it did not seem right to make a List-wrapped array.
```
if (val.getClass().isArray()) {
  elements = Collections.singletonList(val);
}
```
Looking into the history of the change, this line accidentally changed behavior (old behavior: convert array to an equivalent list; new behavior: convert array into as single-element list with the array wrapped as the lone element) as part of a [routine code cleanup](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/commit/76ddf6cc96c7cf5f4fe73fb77f41a248c123d521). I missed this behavior change in code review, but also it's a consequence of the non-type-safe nature of the original code.
Since this would be a breaking change, I suspected that this code was never used. The following is brute-force (brute tediousness?) proof that it is safe to remove this code.
	
					

TwoStepsConversions.convertOnRead(Object, EmbeddedType, Class, TypeInformation) is private and called from 2 places: convertOnRead(Object, Class, Class) and convertOnRead(Object, EmbeddedType, TypeInformation. These are both public because they want to get accessed from autoconfiguration, not because they are part of any documented API.
1. convertOnRead(Object, Class, Class) is called from 4 places: 
    - `DatastoreTemplate.fetchReferenced()`, which passes a List<T> returned from findAllById() as the first parameter
    - `DatastoreTemplate.resolveDescendantProperties()`, which passes an unparameterized List as the first parameter. 
    - `GqlDatastoreQuery.convertCollectionResult()`, which passes `DatastoreResultsIterable`, therefore not an array
    - `GqlDatastoreQuery.convertSingularResult()`. This was a bit hard to reason about, since theoretically the object passed is the first element in a collection returned from either `queryIterable()` or `queryKeysOrEntities()`. 
      - `queryIterable()` will operate on a `DatastoreResultsIterable` of objects returned from `GqlDatastoreQuery::getNonEntityObjectFromRow`, which can be either Key or a primitive Datastore type. Neither is an array.
      - `queryKeysOrEntities()` returns an interable of either Key objects or results from `DatastoreTemplate.convertEntitiesForRead()`, which is a typed List.
    So this use is safe against possibility of an array being passed.
2. convertOnRead(Object val, EmbeddedType embeddedType, TypeInformation targetTypeInformation) is called from 3 places:
    - `DefaultDatastoreEntityConverter.readAsMap()`, which always passes a `String`, so not an array.
    - `EntityPropertyValueProvider.persistentProperty()` which passes a key (extension of `IncompleteKey`).
    - `EntityPropertyValueProvider.getPropertyValue()` which passes a native Datastore type.
  
  I've further verified that this change is safe by adding an exception to the `if (val.getClass().isArray()) {` block and running all unit tests plus all datastore integration tests.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/convert/ReadWriteConversions.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/convert/ReadWriteConversions.java
@@ -36,7 +36,7 @@ public interface ReadWriteConversions {
 
 	/**
 	 * Converts a given object to an object of a target type.
-	 * @param val the value to convert
+	 * @param val the simple type or Iterable value to convert. Arrays are not supported.
 	 * @param targetCollectionType the type of the collection to be converted into.
 	 * {@code null} if the property is a singular object.
 	 * @param targetComponentType the type of the property to convert. For collection-like
@@ -50,7 +50,7 @@ public interface ReadWriteConversions {
 	/**
 	 * Converts a given object to an object of a target type that is possibly an embedded
 	 * entity.
-	 * @param val the value to convert.
+	 * @param val the simple type or Iterable value to convert. Arrays are not supported.
 	 * @param embeddedType contains the type of embedded entity conversion should produce.
 	 * @param targetTypeInformation type metadata information for the desired type.
 	 * @param <T> the type of the object that is produced by reading
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/convert/TwoStepsConversions.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/convert/TwoStepsConversions.java
@@ -17,7 +17,6 @@
 package com.google.cloud.spring.data.datastore.core.convert;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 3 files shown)


meltsufin:
Since the current behavior doesn't make any sense and was actually an accidental regression, I would think we should ultimately fix the regression. Better late than never, I guess. Also, I agree that the fix should come with tests. That being said, in the interim, this PR is fine. The only risk it carries is that someone came to rely on the broken behavior, which I find highly unlikely.

elefeint:
@meltsufin Plot twist: it was not a regression; this code never did the right thing. In fact, I am guessing that the change during refactoring was suggested by Sonar.

Here is why this never worked, in test form:
```
@Test
void convertingArraysTranslatesToList() {
	String[] arr = new String[] { "a", "b", "c"};
	Object arrWithoutType = arr;

	List<String> arrResult = Arrays.asList(arr);
	assertThat(arrResult).containsExactly("a", "b", "c");	// succeeds; result is List<String>

	List arrWithoutTypeResult = Arrays.asList(arrWithoutType);
	assertThat(arrWithoutTypeResult).containsExactly("a", "b", "c");	// fails because result is List<String[]>
}
```

meltsufin (approved):
That wasn't easy to see at a glance. :)
Kudos to Sonar, if it figured it out!
Let's get this in first, and then possibly add the other PR as an enhancement.

PR #734, opened 2021-11-22
update spring-cloud-config.version for pubsub-bus-config-sample

zhumin8:
Update spring-cloud-config.version for pubsub-bus-config-sample compatible with Spring Boot 2.6.0-RC1.


zhumin8:
Is it better to keep this version specification centralized in `spring-cloud-gcp-samples/pom.xml` instead?  We already have `spring-native.version` used for `spring-cloud-gcp-data-firestore-sample` there.

elefeint (approved):
I think if it's only useful for one sample, keeping the property localized in that sample is fine. It's the same amount of work for dependabot.

Another argument in favor of localizing is that each sample is supposed to work on its own, with only a single change updating the maven parent from samples parent to spring boot starter parent.

PR #742, opened 2021-11-23
Datastore array conversion: turn array into a list with same elements

elefeint:
Pick your poison -- this PR fixes the logic, and #733 rips it out.


meltsufin (approved):
Would it make sense to also add an integration test and update docs?

elefeint:
Is this something we *want* to support officially? None of our higher level code exercises this path, so integration test would have to be something fairly contrived.

meltsufin:
I was thinking along the lines of exercising the feature in on of our sample apps.

elefeint:
Closing in favor of #733.

PR #766, opened 2021-12-07
change to index.yaml to match integration test setup.

zhumin8:
`index.yaml` is only used for integration test setup once per GCP project. It affects 3 test in `DatastoreIntegrationTests.java`: `testSlicedEntityProjections`, `testSliceString`, `projectionTest` due to the query method used in tests.
Proposing change aligns with current CI project setting and tested with new GCP project.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/EmbeddedEntity.java
+++ b/spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/EmbeddedEntity.java
@@ -24,19 +24,19 @@ import com.google.cloud.spring.data.datastore.core.mapping.Entity;
  * @author Dmitry Solomakha
  */
 @Entity
-public class EmbeddedEntity {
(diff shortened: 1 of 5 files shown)


elefeint (changes requested):
I wonder if the disconnect is because ` com.google.cloud.spring.data.datastore.it.EmbeddedEntity` is used both from an integration test (`DatastoreIntegrationTests`)
and a unit test (`PartTreeDatastoreQueryTests`). This is not great; we'll eventually accidentally break one or the other.

Could you make EmbeddedEntity package-private (and duplicate it into the unit test package or class), and then remove unused fields?

elefeint (approved):

zhumin8:
It doesn't seem to be directly related, but you are right, we should definitely separate out:
- made `it.EmbeddedEntity` and `it.TestEntity` package private
- for unit test (`PartTreeDatastoreQueryTests`) added `EmbeddedEntity` as inner class (used in `compoundNameConventionTest` and `compoundNameConventionProjectionTest`)

PR #770, opened 2021-12-07
Set graal to 21.2

elefeint:
Java8 is [not supported](https://github.com/paketo-buildpacks/graalvm/commit/276cf8880ed8f81573db0f3577e5679bd71734b7) by the latest graal buildpack.

In any case, we should keep the version of graal in buildpack consistent with the rest of the dependencies.


mpeddada1 (approved):
Looks good!

Issue #205, opened 2021-01-14
Autoconfigure R2DBC support for Cloud SQL


breun:
When using the Spring Cloud GCP Cloud SQL PostgreSQL starter (`spring-cloud-gcp-starter-sql-postgresql`) you get a JDBC-based stack using `org.springframework.boot:spring-boot-starter-jdbc` and the `org.postgresql:postgresql` JDBC driver.

For Spring WebFlux applications I would like to be able to use Cloud SQL via [R2DBC](https://r2dbc.io) instead of JDBC.

I can use R2DBC with Spring Boot and a plain PostgreSQL database via `org.springframework.boot:spring-boot-starter-data-r2dbc` and the `io.r2dbc:r2dbc-postgresql` R2DBC driver for PostgreSQL.

I have found https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/master/r2dbc-postgres and an example project at https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/cloud-sql/r2dbc that shows what dependencies to add and how to create a `ConnectionFactory` bean, but it would be great if Spring Cloud GCP would have a starter and auto-configuration for this, so Spring Cloud GCP users don't need to manage and configure these dependencies.

ttomsu:
Sounds like a great idea. I don't think we have the bandwidth to support it at the moment but once https://github.com/GoogleCloudPlatform/cloud-spanner-r2dbc maybe @elefeint would want to take a stab at Cloud SQL :stuck_out_tongue_winking_eye:

olaf-2:
That would be great. Currently I use an authorized static outgoing IP address for my Cloud Run services as a workaround.

elefeint:
@olaf-2 Have you tried using https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory ? This issue is to simplify setting it up, but the mechanism is there already.

olaf-2:
Yes, did not work. Now I retried with this little :gcp: in the URL I had overlooked and it works. (⌒_⌒;)
Thanks.

chanseokoh:
@elefeint with https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory, I guess then the issue no longer stands?

elefeint:
It does in that we can make the correct `ConnectionFactory` creation automatic given a dependency on R2DBC and socket factory.
I'll update the issue title to reflect that.

PR #772, opened 2021-12-07
Introduce EnvironmentPostProcessor for R2DBC support in Cloud Sql 

mpeddada1:
Part 1 of #205
This PR adds two starters for R2DBC support in Cloud SQL (one for mySQL and another for postgreSQL). It also includes 2 samples, associated integrations tests and documentation. The starters bring in `spring-boot-starter-data-r2dbc` which takes care of the creation of the `ConnectionFactory` bean. 

Currently, authentication can only be done through the use of Application Default Credentials. The feature to pass credentials through `spring.cloud.gcp.credentials` or `spring.cloud.gcp.sql.credentials` will be in the next PR(#775).


spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+            <groupId>com.google.cloud</groupId>
+            <artifactId>spring-cloud-gcp-starter-sql-r2dbc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.r2dbc</groupId>
+            <artifactId>r2dbc-spi</artifactId>
```
meltsufin, line 40: Shouldn't this already be brought in by the starter?

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/test/java/com/example/SqlR2dbcMySqlSampleApplicationIntegrationTests.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.http.ResponseEntity;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Simple integration test to verify the SQL sample application with MySQL.
+ *
+ */
+@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT, classes = { SqlApplication.class }, properties = {
+		"spring.r2dbc.url=r2dbc:gcp:mysql://spring-cloud-gcp-ci:us-central1:testmysql/code_samples_r2dbc_db",
```
meltsufin, line 39: Can we put this configuration in under the `spring-cloud-gcp-ci-it` profile like we did for the non-R2DBC sample?

spring-cloud-gcp-samples/spring-cloud-gcp-sql-postgres-r2dbc-sample/README.adoc
```diff
@@ -0,0 +1,49 @@
+= Spring Cloud GCP MySQL Sample
```
meltsufin, line 1:
```suggestion
= Spring Cloud GCP PostgreSQL R2DBC Sample
```

spring-cloud-gcp-samples/spring-cloud-gcp-sql-postgres-r2dbc-sample/README.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+This code sample demonstrates how to connect to a PostgreSQL Google Cloud SQL instance using the link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql-r2dbc[Spring Cloud GCP R2DBC Starter].
+
+You will create an instance, a database within the instance, populate the database and then query it.
+
+== Setup
+
+image:http://gstatic.com/cloudssh/images/open-btn.svg[link=https://ssh.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https%3A%2F%2Fgithub.com%2FGoogleCloudPlatform%2Fspring-cloud-gcp&cloudshell_open_in_editor=spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-sample/README.adoc]
```
meltsufin, line 9:
```suggestion
image:http://gstatic.com/cloudssh/images/open-btn.svg[link=https://ssh.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https%3A%2F%2Fgithub.com%2FGoogleCloudPlatform%2Fspring-cloud-gcp&cloudshell_open_in_editor=spring-cloud-gcp-samples/spring-cloud-gcp-sql-postgres-r2dbc-sample/README.adoc]
```

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql-r2dbc/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>spring-cloud-gcp-starters</artifactId>
+        <groupId>com.google.cloud</groupId>
+        <version>2.0.7-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>spring-cloud-gcp-starter-sql-r2dbc</artifactId>
+    <name>Spring Cloud GCP Starter - R2DBC</name>
```
meltsufin, line 13: Why only one starter for both MySQL and Postgres R2DBC? We made the decision to split them for JDBC. The problem with combining them is that you're automatically bringing in dependencies that will not be needed.

meltsufin (changes requested):
I mostly commented just on the MySQL side as the comments equally apply to Postgres.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/README.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+4. Open the link:src/main/resources/application.properties[application.properties] file and replace the following:
+- *[database-name]* with the name of the database you created in step 3.
+- *[instance-connection-name]* with the instance name of the MySQL instance you created in step 1.
+- *[username]* with the username for the instance you created in step 1. This should be in the "Users" section of your instance in the console.
++
+If you set a root password, add the `spring.r2dbc.password` property with the root password as the value.
+Note that the password can also be set in the connection url.
+The template is as follows: `r2dbc:gcp:mysql://{username}:{password}@{instance-connection-name}/{database-name}`.
```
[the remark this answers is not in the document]
mpeddada1: That is a good idea! This should be possible for R2DBC. Similar to what is being done in JDBC support, we can create the url (with a default username if not provided) and set it to `spring.r2dbc.url`. I will try this out.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/main/java/com/example/WebController.java
```diff
@@ -0,0 +1,50 @@
...
+public class WebController {
+
+	private final ConnectionFactory connectionFactory;
+
+	public WebController(ConnectionFactory connectionFactory) {
+		this.connectionFactory = connectionFactory;
+	}
+
+	@GetMapping("/getTuples")
+	public List<String> getTuples() {
```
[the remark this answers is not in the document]
elefeint: We have an example [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/main/java/com/example/UserController.java). The difference is one dependency (webflux instead of web) and then reactive types can be returned as is, and are streamed back to the user.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql-r2dbc/pom.xml
mpeddada1: You make a valid point here. Introduced two starters (one for MySQL and PostgreSQL).

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/main/java/com/example/SqlApplication.java
```diff
@@ -0,0 +1,44 @@
...
+
+@SpringBootApplication
+public class SqlApplication {
+
+	public static void main(String[] args) {
+		SpringApplication.run(SqlApplication.class, args);
+	}
+
+	@Bean
+	ConnectionFactoryInitializer initializer(ConnectionFactory connectionFactory) {
```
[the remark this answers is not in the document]
mpeddada1: Good idea, done.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/main/resources/application.properties
```diff
@@ -0,0 +1,7 @@
+spring.r2dbc.url=r2dbc:gcp:mysql://username@instance-connection-name/database-name
+
+# Uncomment if username is not specified in url
+# spring.r2dbc.username=
```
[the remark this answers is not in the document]
mpeddada1: Done. Modified this file since we're now building the r2dbc url from instance connection name and database name.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/pom.xml
mpeddada1: Sorry about that. Removed this dependency.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,103 @@
...
+import org.springframework.core.env.MapPropertySource;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+
+/**
+ * Builds connection string for Cloud SQL through Spring R2DBC by requiring only a
+ * database and instance connection name.
+ *
+ */
+public class R2dbcCloudSqlEnvironmentPostProcessor implements EnvironmentPostProcessor {
```
meltsufin, line 40: Because there is a lot shared code, would it make sense to just enhance the `CloudSqlEnvironmentPostProcessor` instead of introducing another `EnvrionmentPostProcessor`?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcDatabaseType.java
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spring.autoconfigure.sql;
+
+public enum R2dbcDatabaseType {
```
meltsufin, line 19: Would it make sense to just add another property to `DatabaseType` for R2DBC url template?

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/main/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+spring.cloud.gcp.sql.database-name=[database-name]
+spring.cloud.gcp.sql.instance-connection-name=[instance-connection-name]
+# Leave empty for root, uncomment and fill out if you specified a user
+# spring.r2dbc.username=
```
meltsufin, line 4: Can we just piggy-back on the `spring.database.username` property? Same for password.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcDatabaseType.java
mpeddada1: Hm `DatabaseType` includes some extra information, particularly the driver class name that is later set through `spring.datasource.driverClassName`. A similar option doesn't seem to exist for R2DBC so I decided to separate them out on the basis of what data is required for each. Additionally, [`ConnectionFactoryOptions`](https://github.com/r2dbc/r2dbc-spi/blob/d27db1062a9e91417c8c9bc2918d0db9b9f30d26/r2dbc-spi/src/main/java/io/r2dbc/spi/ConnectionFactoryOptions.java#L53) in the R2DBC SPI includes a "Driver" option but our GcpConnectionFactoryProviders end up overriding that value ([ as seen here for example](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/2fc784f272c74c18b54269cd1bcf8214737aa11f/r2dbc/postgres/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProviderPostgres.java#L79)) 

However, I was wondering, was there a reason why we decided to provide the driver class name through the property instead of letting Spring boot determine it from the url?

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/main/resources/application.properties
mpeddada1: In Spring Boot auto-configuration, the connection factory configurations are derived from`[R2dbcProperties`](https://github.com/spring-projects/spring-boot/blob/e38e72132206578ecf3ddaaa07ea3632e670f936/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java#L58) so it may not be possible to replace `spring.r2dbc.*` with `spring.datasource.password`?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessor.java
mpeddada1: Ah that is a good idea! I was initially hesitant about putting them both in `CloudSqlEnvironmentPostProcessor` as it would result in extra `if-else` statements to determine whether it is the r2dbc path or the jdbc path. However, given the presence of so much shared code it might be worth it! I'll try this out.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -64,55 +64,116 @@ public void postProcessEnvironment(ConfigurableEnvironment environment, SpringAp
...
+
+		// Always set the spring.r2dbc.url property in the environment
+		Map<String, Object> primaryMap = new HashMap<>();
+		primaryMap.put("spring.r2dbc.url", r2dbcUrl);
+		environment.getPropertySources().addFirst(new MapPropertySource("CLOUD_SQL_R2DBC_URL", primaryMap));
 	}
 
-	private DatabaseType getEnabledDatabaseType(ConfigurableEnvironment environment) {
+	String createR2dbcUrl(R2dbcDatabaseType databaseType, GcpCloudSqlProperties sqlProperties) {
+		Assert.hasText(sqlProperties.getDatabaseName(), "A database name must be provided.");
```
mpeddada1, line 168: Currently holding off on adding support for enable IAM auth through the url as this doesn't seem to be working as expected for the R2DBC workflow. Extra options specified through the url are automatically typed as `String` so  when [GcpConnectionFactoryProvider#createFactory](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/20881af5ec8669fcb97899f232810ed03138ebfa/r2dbc/core/src/main/java/com/google/cloud/sql/core/GcpConnectionFactoryProvider.java#L103) calls `(Boolean)connectionFactoryOptions.getValue(ENABLE_IAM_AUTH)` this results in an error.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/test/java/com/example/SqlR2dbcMySqlSampleApplicationIntegrationTests.java
mpeddada1: The integration tests use different properties based on the database they're talking to. This might make it harder for them to be under the same profile? The non-R2DBC samples seem to use a hybrid approach where the mySQL test uses the profile while the postgreSQL test configures properties in the test. However, moving all the setup to the test classes could potentially help with readability. I could also be missing some maven-related knowledge here so let me know what you think.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/main/java/com/example/WebController.java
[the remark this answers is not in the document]
mpeddada1: Modified the samples to use WebFlux. Thanks for the suggestion and the example!

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcDatabaseType.java
meltsufin: Why would it be a problem if `DatabaseType` contained constants for both JDBC and R2DBC?
It could have:
`jdbcUrlTemplate, r2dbcUrlTemplate, defaultUsername`

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -52,8 +53,7 @@
  * @author Eddú Meléndez
  */
 public class CloudSqlEnvironmentPostProcessor implements EnvironmentPostProcessor {
```
meltsufin, line 55: Seeing now how handling both JDBC and R2DBC in one class here is not really reusing that much code, and making it less readable, maybe a separate `EventPostProcessor` is a better design after all. You can still re-use some code bt introducing a common `AbstractCloudSqlEnvironmentPostProcessor` that both would extend.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcDatabaseType.java
mpeddada1: My bad! You're right. I take back my comment.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/main/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+spring.cloud.gcp.sql.database-name=[database-name]
+spring.cloud.gcp.sql.instance-connection-name=[instance-connection-name]
```
elefeint, line 2: Could you add a comment that instance-connection-name needs to be in the format `[GCP_PROJECT_ID]:[INSTANCE_REGION]:[INSTANCE_NAME]`? I put just the instance, and got "Cloud SQL connection name is invalid, expected string in the form of "<PROJECT_ID>:<REGION_ID>:<INSTANCE_ID>", which I did not connect with the property until I saw the integration test.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/README.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+2. Create a Google Cloud MySQL instance from the https://console.cloud.google.com/sql/instances[Google Cloud Console Cloud SQL page].
+
+3. Still within the Google Cloud Console SQL page, create a new database in the instance from the "Databases" section.
+If you decided to set a root password for your instance footnoteref:[note,Google Cloud SQL allows for creating MySQL instances without a password,while still remaining safe,as the connection is using Cloud SDK authentication and encrypted communication: see docs on https://cloud.google.com/sql/docs/mysql/sql-proxy[Cloud SQL proxy] and https://cloud.google.com/sql/docs/mysql/connect-external-app#java[the java socketFactory],which is used in spring-cloud-gcp to create the connection!], then don't forget to take note of your root password, as you will need it in the next step!
+
+4. Open the link:src/main/resources/application.properties[application.properties] file and replace the following:
+- *[database-name]* with the name of the database you created in step 3.
+- *[instance-connection-name]* with the instance name of the MySQL instance you created in step 1.
```
elefeint, line 20: Add a bit of information -- "instance name (in the format `[GCP_PROJECT_ID]:[INSTANCE_REGION]:[INSTANCE_NAME]`)"

elefeint (changes requested):
I ran into a self-created issue (provided wrong format for instance), but as part of troubleshooting noticed that Spring Data R2DBC configuration has a concept of `ConnectionFactoryOptionsBuilderCustomizer`. This may be a more natural way to prepare options for the socket factory than the more generic environment post-processor. Could you try to replace R2dbcCloudSqlEnvironmentPostProcessor with a new autoconfiguration class? It will need to be listed in spring.factories under `EnableAutoConfiguration`.

https://docs.spring.io/spring-boot/docs/2.5.7/reference/html/features.html#features.sql.r2dbc

mpeddada1:
> I ran into a self-created issue (provided wrong format for instance), but as part of troubleshooting noticed that Spring Data R2DBC configuration has a concept of `ConnectionFactoryOptionsBuilderCustomizer`. This may be a more natural way to prepare options for the socket factory than the more generic environment post-processor. Could you try to replace R2dbcCloudSqlEnvironmentPostProcessor with a new autoconfiguration class? It will need to be listed in spring.factories under `EnableAutoConfiguration`.
> 
> https://docs.spring.io/spring-boot/docs/2.5.7/reference/html/features.html#features.sql.r2dbc

That is a great discovery! I agree with you that `ConnectionFactoryOptionsBuilderCustomizer` is the most natural way to customize the ConnectionFactory. However, it may come down to the kind of user experience we want to provide. The r2dbc url is expected if we go with this particular workflow. If we don't provide a url then we run into the following exception at the connection factory bean creation step: 
```  Error creating bean with name 'connectionFactory' defined in class path resource [org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations$Pool.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [io.r2dbc.pool.ConnectionPool]: Factory method 'connectionFactory' threw exception; nested exception is org.springframework.boot.autoconfigure.r2dbc.ConnectionFactoryOptionsInitializer$ConnectionFactoryBeanCreationException: Failed to determine a suitable R2DBC Connection URL```

If we want to only require the database name and the instance connection name to be provided without the need for the url (similar to what we have for the jdbc workflow) then the environment post processor will still be needed. Let me know what you think?

**Comment by elefeint** [comment]
@mpeddada1 That makes sense, thank you for checking!

File: spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessor.java [hunk 17]
```diff
@@ -0,0 +1,95 @@
...
+		Assert.hasText(sqlProperties.getDatabaseName(), "A database name must be provided.");
+		Assert.hasText(sqlProperties.getInstanceConnectionName(),
+				"An instance connection name must be provided in the format <PROJECT_ID>:<REGION>:<INSTANCE_ID>.");
+
+		return String.format(databaseType.getR2dbcUrlTemplate(),
+				sqlProperties.getInstanceConnectionName(), sqlProperties.getDatabaseName());
+	}
+
+	DatabaseType getEnabledDatabaseType(ConfigurableEnvironment environment) {
+		if (Boolean.parseBoolean(environment.getProperty("spring.cloud.gcp.sql.enabled", "true"))
```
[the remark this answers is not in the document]
mpeddada1: Done. This method now returns `null` if the dependency that is used to enable Spring R2DBC auto-configuration is not present.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/main/resources/application.properties
mpeddada1: Good idea! This will help make the samples clearer, thank you.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -117,6 +117,31 @@
             <scope>test</scope>
         </dependency>
 
+        <!--R2DBC : mySQL connector factory -->
```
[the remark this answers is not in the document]
mpeddada1: Done, thank you.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/DatabaseType.java
```diff
@@ -29,24 +29,37 @@
...
 	private final String jdbcDriverName;
 
 	private final String jdbcUrlTemplate;
 
+	private final String r2dbcUrlTemplate;
+
 	private final String defaultUsername;
 
-	DatabaseType(String jdbcDriverName, String jdbcUrlTemplate, String defaultUsername) {
+	DatabaseType(String jdbcDriverName, String jdbcUrlTemplate, String r2dbcUrlTemplate, String defaultUsername) {
```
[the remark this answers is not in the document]
mpeddada1: Good point!

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessorTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link R2dbcCloudSqlEnvironmentPostProcessor}.
+ */
+class R2dbcCloudSqlEnvironmentPostProcessorTest {
+
+	private R2dbcCloudSqlEnvironmentPostProcessor initializer = new R2dbcCloudSqlEnvironmentPostProcessor();
```
elefeint, line 32: s/initializer/r2dbcPostProcessor?

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/main/java/com/example/WebController.java
```diff
@@ -0,0 +1,50 @@
...
+public class WebController {
+
+	private final ConnectionFactory connectionFactory;
+
+	public WebController(ConnectionFactory connectionFactory) {
+		this.connectionFactory = connectionFactory;
+	}
+
+	@GetMapping("/getTuples")
+	public Mono<List<String>> getTuples() {
```
elefeint, line 42: `Mono<List>` is a bit odd. You likely don't need `.collectList()` -- just returning the `Flux<String>` should work.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/test/java/com/example/SqlR2dbcMySqlSampleApplicationIntegrationTests.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Simple integration test to verify the SQL sample application with MySQL.
+ *
+ */
+@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT, classes = { SqlApplication.class }, properties = {
+		"spring.cloud.gcp.sql.databaseName=code_samples_test_r2dbc_db",
+		"spring.cloud.gcp.sql.instanceConnectionName=spring-cloud-gcp-ci:us-central1:testmysql",
```
elefeint, line 40: We have those defined in `spring-cloud-gcp-samples/pom.xml` . (`spring.cloud.gcp.sql.instance-connection-name`, `spring.cloud.gcp.sql.database-name`). So it should be possible to omit these when running maven with `-Pspring-cloud-gcp-ci-it`. Omitting these also allows a developer to run against a personal project.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-postgres-r2dbc-sample/src/main/java/com/example/WebController.java
```diff
@@ -0,0 +1,50 @@
...
+public class WebController {
+
+	private final ConnectionFactory connectionFactory;
+
+	public WebController(ConnectionFactory connectionFactory) {
+		this.connectionFactory = connectionFactory;
+	}
+
+	@GetMapping("/getTuples")
+	public Mono<List<String>> getTuples() {
```
elefeint, line 42: Same as in the other sample -- try returning `Flux<String>`. Spring Webflux should handle this output just fine.

elefeint (approved):
Looks good, only a couple of minor changes.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/DatabaseType.java
```diff
@@ -36,26 +36,44 @@
 	 */
 	POSTGRESQL("org.postgresql.Driver", "jdbc:postgresql://google/%s?"
 			+ "socketFactory=com.google.cloud.sql.postgres.SocketFactory"
-			+ "&cloudSqlInstance=%s", "postgres");
+			+ "&cloudSqlInstance=%s", "postgres"),
+
+	/**
+	 * R2DBC MySQL constants.
+	 */
+	R2DBC_MYSQL(null,  "r2dbc:gcp:mysql://%s/%s", "root"),
```
meltsufin, line 44: It's not what I had in mind, but maybe it works better? I was thinking of the enum just having 4 fields: `jdbcDriver`, `jdbcUrlTeample`, `r2dbcUrlTemplate`, `defaultUsername`. In my opinion it makes more sense. Does it not work well?
Also, notice that semantically `R2DBC_MYSQL` is not really a "database type". The fact that you need to pass a `null` for the driver class is also a smell that something is off in design.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Builds connection string for Cloud SQL through Spring R2DBC by requiring only a
+ * database and instance connection name.
+ *
+ */
+public class R2dbcCloudSqlEnvironmentPostProcessor implements EnvironmentPostProcessor {
+	private static final Log LOGGER = LogFactory.getLog(R2dbcCloudSqlEnvironmentPostProcessor.class);
+
+	@Override
+	public void postProcessEnvironment(ConfigurableEnvironment environment, SpringApplication application) {
```
meltsufin, line 41: It's a minor thing, but I think there is even more code here that can be shared by adding some extra parameters to a helper method. It could make sense, unless it makes the code less readable.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/DatabaseType.java
elefeint: This came from my review -- having both `jdbcUrlTemplate` and `r2dbcUrlTemplate` is awkward because one of them is always set and the other is always null. They really represent the same concept -- the URL pattern, so I asked to collapse them. Funny, the "passing null in constructor" was also what tripped me up, but for the URL template field.

For the jdbcDriver field, there does not seem to be an applicable concept in R2DBC since the driver is inferred from the URL pattern. Not sure how you'd avoid setting it to null.

elefeint: Ah, I misunderstood -- I was thinking of enum having 4 constants. But if the enum only has 2 constants (JDBC and R2DBC), then having both, jdbcUrl and r2dbcUrl makes sense.

elefeint: Sorry, two types -- MySQL and Postgres; not JDBC and R2DBC.

mpeddada1: My bad, I was overthinking this. The presence of `jdbcDriver` for r2dbc confused me a little. However, have just `MySQL` and `Postgres` (as you mentioned) with appropriately named getters makes a lot of sense. Thank you!

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/src/test/java/com/example/SqlR2dbcMySqlSampleApplicationIntegrationTests.java
mpeddada1: I was previously using different databases for r2dbc and jdbc when we discussed this earlier (which was probably unnecessary). Modified the mySQL test to use the properties from the pom.xml. Thank you for the suggestion.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessor.java
mpeddada1: Hm that's a good point.  One potential area I can think of refactoring is `CloudJdbcInfoProvider` and `DefaultCloudSqlJdbcProvider`. We used this interface in the past because we also had `AppEngineCloudSqlJdbcInfoProvider` but we may not need this interface anymore -- this can help reduce added hierarchy in the code. Additionally some of the code that is used to create a url can be shared between the r2dbc and jdbc workflow. I can do this refactoring in a follow-up PR if you think this is a good idea.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-postgres-r2dbc-sample/src/main/java/com/example/WebController.java
mpeddada1: Done, thank you!

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
mpeddada1: Created https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/821 to track this.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/PropertiesRetriever.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.boot.context.properties.bind.PlaceholdersResolver;
+import org.springframework.boot.context.properties.bind.PropertySourcesPlaceholdersResolver;
+import org.springframework.boot.context.properties.source.ConfigurationPropertySources;
+import org.springframework.core.env.Environment;
+
+/**
+ * Helper class to derive Cloud SQL and GCP properties from the user configuration
+ * (application.properties, for example).
+ */
+class PropertiesRetriever {
```
meltsufin, line 32: Nice refactoring and reuse!

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessor.java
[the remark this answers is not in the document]
meltsufin: What if the user has the R2DBC dependency, but doesn't want our auto-configuration to kick in? I think we might need something like `spring.cloud.gcp.sql.r2dbc.enabled`.

meltsufin (approved):
LGTM, but note one comment on `spring.cloud.gcp.sql.r2dbc.enabled`.
@elefeint WDYT?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessor.java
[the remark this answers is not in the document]
elefeint: That would be instead of `spring.cloud.gcp.sql.enabled` for Cloud SQL R2DBC autoconfig being on, not in addition to, right?
I guess theoretically it's possible that an app wants `spring.cloud.gcp.sql.enabled` for blocking connection to Cloud SQL, and then an R2DBC driver to another type of database. It would be weird for a reactive app to have blocking components, but not impossible.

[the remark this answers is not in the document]
meltsufin: Ideally, `spring.cloud.gcp.sql.enabled=false` would disable all of our Cloud SQL starters.
Maybe we need to add two properties: `spring.cloud.gcp.sql.r2dbc.enabled` and `spring.cloud.gcp.sql.jdbc.enabled`.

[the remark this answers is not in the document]
elefeint: So for this PR, the check would be that both `spring.cloud.gcp.sql.enabled` and `spring.cloud.gcp.sql.r2dbc.enabled` are true. That sounds good.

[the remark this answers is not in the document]
meltsufin: Yep. Also, consider introducing `spring.cloud.gcp.sql.jdbc.enabled`.
Eventually, we might want to deprecate `spring.cloud.gcp.sql.enabled`.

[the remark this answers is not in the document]
mpeddada1: This is something that I did not think of before. Since it is possible for an app to expect a blocking connection while having the r2dbc driver on the classpath, introducing `spring.cloud.gcp.sql.r2dbc.enabled` sounds like a good idea.  I'll add `spring.cloud.gcp.sql.jdbc.enabled` in a separate PR since it might break existing behavior for current cloud sql customers.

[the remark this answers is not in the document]
mpeddada1: @meltsufin @elefeint Added `spring.cloud.gcp.sql.r2dbc.enabled` which will be `true` by default.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/PropertiesRetriever.java
mpeddada1: Thank you!

elefeint (approved):
Looks great.

Could you add a paragraph to refdoc in `cloud-sql.adoc`? Can be a followup PR.

mpeddada1:
Thank you! Yup, I'll do this in a follow-up PR.

PR #775, opened 2021-12-08
R2DBC: Allow external credentials to be provided to Cloud SQL  

mpeddada1:
Follow-up to #772

This PR sets the credentials system properties for credentials provided through `spring.cloud.gcp.credentials` and `spring.cloud.gcp.sql.credentials`. It moves all the credentials related logic from `CloudSqlEnvironmentPostProcessor` into a helper class that can be shared between the EnvironmentPostProcessors for JDBC and R2DBC support.


spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql-r2dbc/pom.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
elefeint, line 1: I am confused about this starter -- is it for both, postgres and mysql at the same time?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CredentialsPropertiesSetter.java
```diff
@@ -0,0 +1,91 @@
...
+
+    if (sqlProperties.getCredentials().hasKey()) {
+      // First tries the SQL configuration credential.
+      credentials = sqlProperties.getCredentials();
+    } else {
+      // Then, the global credential.
+      credentials = gcpProperties.getCredentials();
+    }
+
+    if (credentials.getEncodedKey() != null) {
```
elefeint, line 60: credentials could also be null here if neither sql nor gcp properties have them; add a check before this line to return without doing anything.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CredentialsPropertiesSetter.java
```diff
@@ -0,0 +1,91 @@
...
+      File credentialsLocationFile = credentialsLocation.getFile();
+
+      System.setProperty(SqlCredentialFactory.CREDENTIAL_LOCATION_PROPERTY_NAME,
+          credentialsLocationFile.getAbsolutePath());
+
+      // If there are specified credentials, tell sockets factory to use them.
+      System.setProperty(CredentialFactory.CREDENTIAL_FACTORY_PROPERTY,
+          SqlCredentialFactory.class.getName());
+    } catch (IOException ioe) {
+      logger.info("Error reading Cloud SQL credentials file.", ioe);
```
elefeint, line 88: I think this failure should propagate back to the user. If the user provided custom credentials and they did not work, the application is not really expected to work. Unless falling back to default credentials is acceptable (it might be in some cases? but it's hard to assume).

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -86,7 +80,7 @@ public void postProcessEnvironment(
           .getPropertySources()
           .addFirst(new MapPropertySource("CLOUD_SQL_DATA_SOURCE_URL", primaryMap));
 
-      setCredentials(sqlProperties, propertiesRetriever.getGcpProperties());
+      CredentialsPropertiesSetter.setCredentials(sqlProperties, propertiesRetriever.getGcpProperties(), LOGGER);
```
elefeint, line 83: No need to pass the logger; define a static logger in the `CredentialsPropertiesSetter` if it's needed. Per-class loggers are useful because end users can control logging verbosity per package/class if needed.

elefeint (changes requested):
I wonder if it makes sense to file an issue in socket factory to accept credentials in a less environment-dependent way. Or do you think they had technical constraints that led to the environment property being the only way to inject the key?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql-r2dbc/pom.xml
mpeddada1: Sorry about that. Removing this sample. I must've missed it while resolving merge conflicts.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CredentialsPropertiesSetter.java
mpeddada1: `Credentials` is initialized (` private Credentials credentials = new Credentials();`) in sql and gcp properties so we may not run into this problem? Unless I'm missing something here.

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/07d75000a704a2266ca4619e1b75534faec329a7/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/GcpCloudSqlProperties.java#L35

mpeddada1:
> I wonder if it makes sense to file an issue in socket factory to accept credentials in a less environment-dependent way. Or do you think they had technical constraints that led to the environment property being the only way to inject the key?

That is a good point. I think this idea was initially considered but was later dropped in favor of environment variables because the Postgres driver didn't allow arbitrary properties to be passed to the socket factory through the JDBC url. https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/41. This discussion happened a while ago so the status may have changed.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CredentialsPropertiesSetter.java
```diff
@@ -73,7 +71,8 @@ private static void setCredentialsEncodedKeyProperty(String encodedKey) {
         SqlCredentialFactory.class.getName());
   }
 
-  private static void setCredentialsFileProperty(Resource credentialsLocation, Log logger) {
+  static void setCredentialsFileProperty(Resource credentialsLocation)
+      throws IllegalArgumentException {
```
elefeint, line 75: `IllegalArgumentException` is a runtime exception; it does not need to be declared or propagated.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CredentialsPropertiesSetter.java
elefeint: No, that makes sense. Just looked suspicious at first glance.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/PropertiesRetriever.java
```diff
@@ -57,6 +57,8 @@ GcpProperties getGcpProperties() {
     return this.binder.bind(gcpPropertiesPrefix, GcpProperties.class).orElse(new GcpProperties());
   }
 
+
+
```
elefeint, line 61: (nit) revert extra whitespace?

spring-cloud-gcp-samples/spring-cloud-gcp-sql-mysql-r2dbc-sample/README.adoc
```diff
@@ -24,6 +24,9 @@ If you would like to use a different user instead of the default (`root`), set t
 If you set a root password, add the `spring.r2dbc.password` property with the root password as the value.
 
 5. https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login[If you are authenticated in the Cloud SDK], your credentials will be automatically found by the Spring Boot Starter for Google Cloud SQL.
++
+Alternatively, https://console.cloud.google.com/iam-admin/serviceaccounts[create a service account from the Google Cloud Console] and download its private key.
+Then, uncomment the `spring.cloud.gcp.credentials.location` property in the link:src/main/resources/[application.properties] file and fill its value with the path to your service account private key on your local file system, prepended with `file:`.
```
elefeint, line 29:
```suggestion
Then, uncomment the `spring.cloud.gcp.credentials.location` property in the link:src/main/resources/application.properties[application.properties] file and fill its value with the path to your service account private key on your local file system, prepended with `file:`.
```

spring-cloud-gcp-samples/spring-cloud-gcp-sql-postgres-r2dbc-sample/README.adoc
```diff
@@ -25,6 +25,9 @@ If you would like to use a different user instead of the default (`postgres`), s
 If you set a root password, add the `spring.r2dbc.password` property with the root password as the value.
 
 5. https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login[If you are authenticated in the Cloud SDK], your credentials will be automatically found by the Spring Boot Starter for Google Cloud SQL.
++
+Alternatively, https://console.cloud.google.com/iam-admin/serviceaccounts[create a service account from the Google Cloud Console] and download its private key.
+Then, uncomment the `spring.cloud.gcp.credentials.location` property in the link:src/main/resources/[application.properties] file and fill its value with the path to your service account private key on your local file system, prepended with `file:`.
```
elefeint, line 30:
```suggestion
Then, uncomment the `spring.cloud.gcp.credentials.location` property in the link:src/main/resources/application.properties[application.properties] file and fill its value with the path to your service account private key on your local file system, prepended with `file:`.
```

spring-cloud-gcp-samples/spring-cloud-gcp-sql-postgres-sample/README.adoc
```diff
@@ -31,7 +31,7 @@ If you would like to use a different user, set the `spring.datasource.username`
 5. https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login[If you are authenticated in the Cloud SDK], your credentials will be automatically found by the Spring Boot Starter for Google Cloud SQL.
 +
 Alternatively, https://console.cloud.google.com/iam-admin/serviceaccounts[create a service account from the Google Cloud Console] and download its private key.
-Then, uncomment the `spring.cloud.gcp.credentials.location` property in the link:src/main/resources/application.properties file and fill its value with the path to your service account private key on your local file system, prepended with `file:`.
+Then, uncomment the `spring.cloud.gcp.credentials.location` property in the link:src/main/resources/[application.properties] file and fill its value with the path to your service account private key on your local file system, prepended with `file:`.
```
elefeint, line 34:
```suggestion
Then, uncomment the `spring.cloud.gcp.credentials.location` property in the link:src/main/resources/application.properties[application.properties] file and fill its value with the path to your service account private key on your local file system, prepended with `file:`.
```

elefeint (changes requested):
Only minor things left!

elefeint (approved):

Issue #774, opened 2021-12-08
Spanner @Embedded annotation doesn't work with findById


gaeshi:
I'm trying to follow the latest official reference on how to use @Embedded annotation to reuse set of columns between multiple entities:
https://googlecloudplatform.github.io/spring-cloud-gcp/2.0.6/reference/html/index.html#embedded-objects

When I'm using `SpannerRepository.save(Entity)`, everything works as expected, but when I'm doing `SpannerRepository.findById(Entity.id)`, I'm getting `SpannerDataException: Column not found`. The logs from GapicSpannerRpc look good, it seems that the problem is somewhere during mapping.

I made a new sample project with Spring Initializr (Spring Boot 2.5.7, Gradle, Kotlin) and version 2.0.6 of spring-cloud-gcp-starter-data-spanner.

Here is the DDL I used to create a new table in Spanner:

```sql
CREATE TABLE Requests (
  requestId STRING(MAX),
  payload STRING(MAX),
  createdAt TIMESTAMP,
  createdBy STRING(MAX),
  updatedAt TIMESTAMP,
  updatedBy STRING(MAX)
) PRIMARY KEY (requestId)
```

Minimal example code:

```kt
@SpringBootApplication
class EmbeddedObjectsApplication

fun main(args: Array<String>) {
    runApplication<EmbeddedObjectsApplication>(*args)
}

data class EntityHistory(val createdAt: Date, val createdBy: String, val updatedAt: Date, val updatedBy: String) {
    constructor(userId: String) : this(
        createdAt = Date(System.currentTimeMillis()),
        createdBy = userId,
        updatedAt = Date(System.currentTimeMillis()),
        updatedBy = userId
    )
}

@Table(name = "Requests")
data class Request(@PrimaryKey val requestId: String, val payload: String, @Embedded val entityHistory: EntityHistory)

interface RequestRepository : SpannerRepository<Request, String>

@RestController
class ApiController(val requestRepository: RequestRepository) {

    @GetMapping("/requests/{requestId}")
    fun getRequest(@PathVariable requestId: String) = requestRepository.findById(requestId) // fails with "SpannerDataException: Column not found: entityHistory"

    @PostMapping("/requests")
    fun saveRequest(@RequestBody payload: String): String {
        val requestId = UUID.randomUUID().toString()
        requestRepository.save(Request(requestId, payload, EntityHistory("some-user-id"))) // works fine
        return requestId
    }
}

```

[...]

PR #784, opened 2021-12-09
Document restriction on embedded objects

suztomo:
Related to https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/774

### DIFF
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -317,8 +317,11 @@ CREATE TABLE items (
 ) PRIMARY KEY (grandParentId, parentId, id)
 ----
 
-Note that embedded properties' column names must all be unique.
+Note that the following restrictions apply when you use embedded objects:
 
+- embedded properties' column names must all be unique.
+- embedded properties must be mutable; otherwise you'll get an error, such as `SpannerDataException: Column not found: parent`.
+  Therefore, you cannot use Kotlin's data class, which is immutable, to hold an embedded property.
 
 ==== Relationships
 
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -319,9 +319,9 @@ CREATE TABLE items (
 
 Note that the following restrictions apply when you use embedded objects:
 
-- embedded properties' column names must all be unique.
-- embedded properties must be mutable; otherwise you'll get an error, such as `SpannerDataException: Column not found: parent`.
-  Therefore, you cannot use Kotlin's data class, which is immutable, to hold an embedded property.
+- Embedded properties' column names must all be unique.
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/spanner.adoc
```diff
@@ -317,8 +317,11 @@ CREATE TABLE items (
 ) PRIMARY KEY (grandParentId, parentId, id)
 ----
 
-Note that embedded properties' column names must all be unique.
+Note that the following restrictions apply when you use embedded objects:
 
+- embedded properties' column names must all be unique.
+- embedded properties must be mutable; otherwise you'll get an error, such as `SpannerDataException: Column not found: parent`.
+  Therefore, you cannot use Kotlin's data class, which is immutable, to hold an embedded property.
```
suztomo, line 324: This is not accurate. Kotlin's data class can have mutable field with 'var'. But the following definition (`@Embedded var entityHistory`) failed too:

```
@Table(name = "Requests")
data class Request(
    @PrimaryKey val requestId: String,
    val payload: String,
    @Embedded var entityHistory: EntityHistory? = null
)
```

What's the better description?

gaeshi: Can we rephrase somehow that the constructor should not contain embedded properties?

It seems that to use Kotlin data class with `@Embedded`, we can do something like:

```kt
@Table(name = "Requests")
data class Request(
    @PrimaryKey val requestId: String,
    val payload: String,
) {
    @Embedded lateinit var entityHistory: EntityHistory
}
```

or

```kt
@Table(name = "Requests")
data class Request(
    @PrimaryKey val requestId: String,
    val payload: String,
) {
    @Embedded var entityHistory: EntityHistory? = null
}
```

elefeint: How about "Embedded properties must not be passed through constructor"? This would apply to both Java and Kotlin.

suztomo: @elefeint @gaeshi Thank you for feedback. Updated.

suztomo (commented):
@elefeint PTAL.

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -317,8 +317,11 @@ CREATE TABLE items (
 ) PRIMARY KEY (grandParentId, parentId, id)
 ----
 
-Note that embedded properties' column names must all be unique.
+Note that the following restrictions apply when you use embedded objects:
 
+- Embedded properties' column names must all be unique.
+- Embedded properties must not be passed through a constructor and the property must be mutable; otherwise you'll get an error, such as `SpannerDataException: Column not found: parent`.
```
elefeint, line 323:
```suggestion
- Embedded properties must not be passed through a constructor and the property must be mutable; otherwise you'll get an error, such as `SpannerDataException: Column not found`.
```

elefeint (approved):

PR #798, opened 2021-12-14
Migrating "MultipleDataModuleIntegrationTest.java" and "MultipleNamespaceDatastoreIntegrationTest.java" to JUnit5

ddixit14:
Migrated the tests to JUnit5 and created a "multisample" integration test suite which is responsible for running "MultipleDataModuleIntegrationTest.java" in GHA.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
@@ -32,7 +32,7 @@ import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
  *
  * @author Chengyuan
  */
-////Please use "-Dit.spanner=true -Dit.datastore=true" to enable the tests
+////Please use "-Dit.multisample=true" to enable the tests
 @ExtendWith(SpringExtension.class)
 @EnabledIfSystemProperty(named = "it.multisample", matches = "true")
 @TestPropertySource("classpath:application-test.properties")
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
@@ -16,28 +16,29 @@
 
 package com.example;
 
-import org.junit.BeforeClass;
-import org.junit.Test;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
```diff
@@ -16,28 +16,29 @@
...
-import static org.hamcrest.Matchers.is;
-import static org.junit.Assume.assumeThat;
 
 /**
  * Tests that our Spring Data modules can be used with each other.
  *
  * @author Chengyuan
  */
-@RunWith(SpringRunner.class)
+////Please use "-Dit.spanner=true -Dit.datastore=true" to enable the tests
```
elefeint, line 35:
```suggestion
//Please use "-Dit.spanner=true -Dit.datastore=true" to enable the tests
```

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
```diff
@@ -16,28 +16,29 @@
...
+@ExtendWith(SpringExtension.class)
+@EnabledIfSystemProperty(named = "it.spanner", matches = "true")
```
elefeint, line 37: Well, this is an interesting case! I was going to comment that this changes behavior -- in the past this test would only run if both properties are on, and now it's one or the other. But that's an improvement, actually, since the old way must have never run at all for a year. Our GithubActions setup is split by module, so that only one property is set up at any time.

Could you file a follow-up issue to create a "multimodule" integration test suite? This will require changes in integrationTests.yaml and this test right here.

elefeint (approved):

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
ddixit14: Okay I'll create the issue. 
I investigated a little though. Do we just need to add something like "multiSample" in jobs/integrationTests/strategy/matrix/it in integrationTests.yaml and specify "-Dit.multiSample=true" in this sample? Or is there something else that needs to be done, something like specifying that "-Dit.multiSample=true" includes "-Dit.spanner=true" and "-Dit.datastore=true" ?

elefeint: It's the first scenario. Because if we turn a multi-sample into "switch spanner and datastore tests to run", then this test will run under both, Spanner matrix row and Datastore matrix row.
It would be better to have a special "multimodule" category of integration tests in which these would run once and only once.

elefeint (approved):

elefeint (approved):

PR #804, opened 2021-12-15
Remove spring-cloud-build dependency; add spring-boot BOM explicitly

### LINKED ISSUE #801: Switch to managing version of Spring Boot in this project
We still rely on Spring Cloud Build for the default Spring Boot version to build against. Reasoning about versions and keeping the `spring-cloud-gcp-samples` Spring Boot version in sync would be easier if it was explicitly managed here.

We likely don't need `spring-cloud-build-dependencies` at all, except for checkstyle, and that will go away when we switch to Google style next year.

Related issue: spring-cloud/spring-cloud-build#224

elefeint:
We currently derive Spring Boot versions indirectly, through Spring Cloud Build. That's not ideal because for Spring Cloud Build, it's a minor infrastructural detail, while users of Spring Cloud GCP can get the version of Spring Boot through our starters.

We should explicitly maintain the version of Spring Boot.

Fixes #801.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
 		<skipTests>false</skipTests>
 		<spring-cloud-build.version>3.0.4</spring-cloud-build.version>
 		<spring-cloud-dependencies.version>2020.0.4</spring-cloud-dependencies.version>
-		<spring-boot-dependencies.version>2.5.1</spring-boot-dependencies.version>
+		<spring-boot-dependencies.version>2.5.7</spring-boot-dependencies.version>
 		<spring-javaformat-checkstyle.version>0.0.29</spring-javaformat-checkstyle.version>
 		<zipkin-gcp.version>1.0.2</zipkin-gcp.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -64,6 +64,7 @@
 		<skipTests>false</skipTests>
 		<spring-cloud-build.version>3.0.4</spring-cloud-build.version>
```
meltsufin, line 65: Should you remove this then?

meltsufin (changes requested):

pom.xml
elefeint: No, we are still using it for the checkstyle plugin. This will go away soon -- I am thinking the best time to do the checkstyle changes is right before forking off Spring Cloud GCP 2.x (with Spring Boot 2.5 support) and leaving the main branch to be Spring Cloud GCP 3.x (with Spring Boot 2.6 support).

meltsufin (approved):

PR #825, opened 2021-12-28
Migrating "MetricsSampleApplicationIntegrationTests.java" to JUnit5

ddixit14:


elefeint (changes requested):
It's interesting that the test succeeds despite this message in logs: `i.m.s.StackdriverMeterRegistry           : failed to send metrics to Stackdriver
2021-12-28T22:01:00.6318835Z 
2021-12-28T22:01:00.6375763Z com.google.api.gax.rpc.InternalException: io.grpc.StatusRuntimeException: INTERNAL: One or more TimeSeries could not be written: Internal error encountered. Please retry after a few seconds. If internal errors persist, contact support at https://cloud.google.com/support/docs.: timeSeries[61]`.

Could you check if this error is reproducible, and if so what it is that it's failing to write to Cloud Metrics?

elefeint (approved):

elefeint (approved):

PR #826, opened 2021-12-28
Migrating "LocalSampleAppIntegrationTest.java" to JUnit5

ddixit14:


spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/src/test/java/com/example/LocalSampleAppIntegrationTest.java
```diff
@@ -143,7 +140,7 @@ public void closeResources() throws IOException {
 
 	}
 
-	private void startConfigServer() throws IOException {
```
elefeint, line 146: Going from `private` to package-private is relaxing visibility. Generally, it's best to have the least visibility that gets the job done. These private methods should stay private, unless there is a reason to change them.

elefeint (changes requested):

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/src/test/java/com/example/LocalSampleAppIntegrationTest.java
ddixit14: I had this in mind too but I thought uniformity across all the integration tests would be a better choice. No worries, I'll revert this change.

elefeint (approved):

PR #829, opened 2021-12-29
Migrating "PubSubSampleApplicationIntegrationTests.java" to JUnit5

ddixit14:


spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/test/java/com/example/PubSubSampleApplicationIntegrationTests.java
```diff
@@ -220,7 +213,7 @@ public void testMultiPull() {
 				.collect(Collectors.toList());
 	}
 
-	private void createTopic(String topicName) {
```
elefeint, line 223: Similar to a previous PR -- private methods should stay private unless they actually need to be accessed from outside of this class.

elefeint (changes requested):

elefeint (approved):

PR #835, opened 2022-01-04
Bump spring-cloud version to 2020.0.5

elefeint:
some test failed; don't recall which

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,7 @@
 		<skip.surefire.tests>${skipTests}</skip.surefire.tests>
 		<skipTests>false</skipTests>
 		<spring-cloud-build.version>3.0.4</spring-cloud-build.version>
-		<spring-cloud-dependencies.version>2020.0.4</spring-cloud-dependencies.version>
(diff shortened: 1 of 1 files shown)


q
```diff
@@ -0,0 +1,45 @@
+[1mdiff --git a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinderEmulatorIntegrationTests.java b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinderEmulatorIntegrationTests.java[m
```
zhumin8, line 1: looks like some temp diff file sneaked in here.

zhumin8 (approved):
LGTM.
I missed that we could indeed reuse the application context from the super class. I like this better than refresh() in #730. I will change it for 3.x also then.

q
elefeint: oops. Can't even blame the cat; this was me.

Issue #692, opened 2021-11-11
Spring Boot 2.6: QueryByExampleExecutor has a new findBy() method


elefeint:
RC1 introduced https://github.com/spring-projects/spring-data-commons/pull/2421

Release notes: https://spring.io/blog/2021/10/18/spring-data-2021-1-enters-rc-phase

This affects `SimpleDatastoreRepository`

PR #836, opened 2022-01-04
Implement new findBy() method with FluentQuery for SimpleDatastoreRepository

zhumin8:
Implementing new `findBy()` method in `SimpleDatastoreRepository` and corresponding `FluentQuery.FetchableFluentQuery` interface introduced by spring-data-commons-2.6.0 ([#2421](https://github.com/spring-projects/spring-data-commons/pull/2421)).
This is basically implementing already supported query methods to fit functional programming model.

Note: In this PR, leaving `project()` and `as()` methods as unsupported for now. As Datastore query by example does not support projection overall yet[(doc)](https://googlecloudplatform.github.io/spring-cloud-gcp/2.0.7/reference/html/index.html#projections-2). Let's have a separate discussion about it.

fix for #692

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/support/SimpleDatastoreRepository.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/support/SimpleDatastoreRepository.java
@@ -280,13 +280,9 @@ public class SimpleDatastoreRepository<T, I> implements DatastoreRepository<T, I
 			return stream().collect(Collectors.toList());
 		}
 
-		/**
-		 * Get a page of matching elements for {@link Pageable}.
-		 *
-		 * @param pageable must not be {@code null}. The given {@link Pageable} will override any
-		 *     previously specified {@link Sort sort} if the {@link Sort} object is not
-		 *     {@link Sort#isUnsorted()}.
-		 * @return
+		/*
+		 * (non-Javadoc)
+		 * @see org.springframework.data.repository.query.FluentQuery.FetchableFluentQuery#page(org.springframework.data.domain.Pageable)
 		 */
 		@Override
 		public Page<R> page(Pageable pageable) {
--- a/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/support/SimpleDatastoreRepository.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/support/SimpleDatastoreRepository.java
@@ -45,6 +45,7 @@ import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.repository.query.FluentQuery;
 import org.springframework.data.util.Streamable;
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
 /**
@@ -251,12 +252,15 @@ public class SimpleDatastoreRepository<T, I> implements DatastoreRepository<T, I
 		}
 
 		@Override
+		public Optional<R> one() {
+			return (Optional<R>) SimpleDatastoreRepository.this.findOne(example);
+		}
+
+		@Nullable
+		@Override
 		public R oneValue() {
-			Optional<S> one = SimpleDatastoreRepository.this.findOne(example);
-			if (one.isPresent()) {
-				return (R) one;
-			}
-			return null;
+			Optional<R> one = one();
+			return one.orElse(null);
 		}
 
 		@Override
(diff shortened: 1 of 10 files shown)


spring-cloud-gcp-data-datastore/src/test/resources/index.yaml
```diff
@@ -2,8 +2,11 @@ indexes:
 
 - kind: test_entities_ci
   properties:
-  - name: embeddedEntity.stringField
-  - name: blobField
+  - name: size
   - name: color
```
zhumin8, line 6: This part is copied from main branch.

spring-cloud-gcp-data-datastore/src/test/resources/index.yaml
```diff
@@ -2,8 +2,11 @@ indexes:
...
-  - name: shape
-  - name: size
\ No newline at end of file
+
+- kind: test_entities_ci
+  properties:
+  - name: color
+  - name: __key__
+    direction: desc
```
zhumin8, line 12: This index is added for integration tests.

spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/support/SimpleDatastoreRepository.java
```diff
@@ -218,7 +230,95 @@ public void deleteAllById(Iterable<? extends I> iterable) {
...
+		@Nullable
+		@Override
+		public R oneValue() {
+			Optional<R> one = one();
+			return one.orElse(null);
+		}
+
+		@Override
+		public R firstValue() {
+			return (R) SimpleDatastoreRepository.this.findFirstSorted(this.example, this.sort);
```
elefeint, line 282: Should this assert that the current sort is not `Sort.unsorted()` before trying to find first sorted? Or at least log a warning. So users don't have to troubleshoot non-repeatable behavior...

spring-cloud-gcp-data-datastore/src/test/resources/index.yaml
```diff
@@ -2,8 +2,11 @@ indexes:
...
   properties:
-  - name: embeddedEntity.stringField
-  - name: blobField
+  - name: size
   - name: color
-  - name: shape
-  - name: size
\ No newline at end of file
+
+- kind: test_entities_ci
```
elefeint, line 8: this is a very minor quibble, and I don't even know if it's worth changing, but the index might not be used from CI. I just applied it locally to test.

spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/repository/support/SimpleDatastoreRepository.java
```diff
@@ -218,7 +230,95 @@ public void deleteAllById(Iterable<? extends I> iterable) {
...
+		@NonNull
+		@Override
+		public List<R> all() {
+			return stream().collect(Collectors.toList());
+		}
+
+		@NonNull
+		@Override
+		public Page<R> page(@NonNull Pageable pageable) {
+			return (Page<R>) SimpleDatastoreRepository.this.findAll(this.example, pageable);
```
elefeint, line 294: The cast from `page<source>` to `page<returnType>` gave me pause, so I looked into it. Not supporting type conversions for now sounds fine, but then we should probably remove the `R` parameter completely. The class signature would then be `DatastoreFluentQueryByExample<S extends T> implements FluentQuery.FetchableFluentQuery<S>`, and the only place `R` would show up is in the signature of `findBy`. This should be reasonably safe, since `DatastoreFluentQueryByExample` is a package-private class that's an implementation detail, and we can change it if we add support for `as`.

elefeint (changes requested):
This looks overall very good. Let's talk about support for return type conversion a bit more today.

spring-cloud-gcp-data-datastore/src/test/resources/index.yaml
zhumin8: Interesting, when testing on my gcp project, it would fail with this error below if not setting this index.
```
com.google.cloud.datastore.DatastoreException: no matching index found. recommended index is:
- kind: test_entities_ci
  properties:
  - name: color
  - name: __key__
    direction: desc
```

elefeint: Right, on mine too. Sorry, I thought you've introduced this "kind", but I see it's been there forever. Ignore my comment.

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/Singer.java
```diff
@@ -165,6 +165,11 @@ public String toString() {
 				+ '}';
 	}
 
+	public String FirstAndLastName() {
```
elefeint, line 168:
```suggestion
	public String firstAndLastName() {
```

elefeint (approved):
I think at this point we'll end up merging this PR straight into main, not into 3.x.

zhumin8:
I'll update base branch once #838 gets merged in.

elefeint (approved):

PR #841, opened 2022-01-06
Added scheduled CI workflows for 2.x

elefeint:
GHA "schedule" task does not allow you to set custom branches; it always runs on the last commit of the main branch. So I added two new actions to handle the scheduled unit/integration tests for the 2.x branch.

Pull requests and pushes to the branch will continue being handled by the existing yaml configuration files.

I also removed all native IT setup from the 2.x integration configuration. No reason to maintain it for the 2.x branch, as those are not released artifacts.

### DIFF
--- a/.github/workflows/integrationTests-2.x.yaml
+++ b/.github/workflows/integrationTests-2.x.yaml
@@ -1,0 +1,131 @@
+name: Scheduled Integration Tests for 2.x
+
+on:
+  # push to 2.x branch is handled by default integrationTests.yaml
+  workflow_dispatch:
+  schedule:
+    - cron: '05 5 * * *' # 05:00 UTC every day
+
+
+jobs:
+  integrationTests:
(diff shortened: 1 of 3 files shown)


.github/workflows/integrationTests-2.x.yaml
```diff
@@ -0,0 +1,132 @@
...
+          - storage
+          - trace
+          - vision
+    steps:
+      - name: Get current date
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d' --utc)"
+      - uses: actions/checkout@v2
+        with:
+          ref: '2.x'
```
zhumin8, line 43: Not sure how tolerant GHA is on config. According to this guide I found, branch name goes without quotes?
https://github.com/actions/checkout#Checkout-a-different-branch

zhumin8 (approved):
LGTM

PR #884, opened 2022-01-13
Update changelog for 3.0.0

elefeint:


zhumin8 (approved):
I think we should also update supported security policy versions in `SECURITY.md`. 
Or we could do that in a separate PR.

elefeint:
Good point; updated.

PR #897, opened 2022-01-20
fix table columns.

zhumin8:
Fix column mismatch in doc:
image


meltsufin (approved):
Thank you for fixing this!

elefeint (approved):

Issue #864, opened 2022-01-11
[pubsub] Graceful shutdown for underlying Publisher


njkevlani:
Hello,

Does [`PubSubPublisherTemplate`](https://docs.spring.io/spring-cloud-gcp/docs/1.1.0.M2/api/org/springframework/cloud/gcp/pubsub/core/publisher/PubSubPublisherTemplate.html) provide functionality to call [`.shutdown`](https://googleapis.dev/java/google-cloud-pubsub/latest/com/google/cloud/pubsub/v1/Publisher.html#shutdown--) and [`.awaitTermination`](https://googleapis.dev/java/google-cloud-pubsub/latest/com/google/cloud/pubsub/v1/Publisher.html#awaitTermination-long-java.util.concurrent.TimeUnit-) methods from underlying [`com.google.cloud.pubsub.v1.Publisher`](https://googleapis.dev/java/google-cloud-pubsub/latest/com/google/cloud/pubsub/v1/Publisher.html)?

We have a use case where we want to be able to control the shutdown for the underlying publisher.

> **Describe the solution you'd like**

We would want to be able to have a method in PubSubPublisherTemplate that shuts down underlying Publishers. Or to have an API to access underlying Publishers.

elefeint:
It's a valid feature request. In fact, it's a bug/omission that we don't shut down publishers on Spring context shutdown ourselves.

@njkevlani Is your usecase related to application shutdown issues, or do you want to clean up publishers throughout the application lifespan?

njkevlani:
Hey @elefeint,

Yes, our usecase is about application's graceful shutdown.

PR #900, opened 2022-01-21
Replace PublisherInterface with Publisher

### LINKED ISSUE #790: PublisherFactory returns PublisherInterface instead of Publisher
**Describe the bug**
There was a change in the PublisherFactory in `com.google.cloud:spring-cloud-gcp-pubsub` version 2.0.6 used in `com.google.cloud:spring-cloud-gcp-starter-pubsub` which caused one of our applications to have a compilation error on the following line:
```
pubsubTemplate.getPublisherFactory.createPublisher(topic).publishAllOutstanding()
```

The reason for this is as follows:

In version 2.0.5 the PublisherFactory returns a Publisher 
```
public interface PublisherFactory {
    Publisher createPublisher(String topic);
}
``` 
while in v 2.0.6 it returns a PublisherInterface 
```
public interface PublisherFactory {
    PublisherInterface createPublisher(String topic);
}
``` 

The PublisherInterface doesn't expose the Publisher method `publishAllOutstanding()`.

This change is not mentioned in the change log for (release 2.0.6)[https://github.com/GoogleCloudPlatform/spring-cloud-gcp/releases/tag/v2.0.6].

The change was part of PR #548. 
In this change I notice that the `CachingPublisherFactory` implementation of PublisherFactory was also change to return a PublisherInterface instead of a Publisher, but the `DefaultPublisherFactory` was not (and still returns a Publisher).

Since this isn't mentioned in the release notes, I wonder if this change was intentional? ~And also whether we can fix our compilation error by casting to a `DefaultPublisherFactory` going forward, or~ should assume that the `publishAllOutstanding()` is not intended to be exposed (as it's not part of the PublisherInterface).

Update: casting to DefaultPublisherFactory didn't work.

elefeint:
After the discussion in googleapis/java-pubsub#949, reverting back to using `Publisher` for all API return types. I've ported the trace header injection functionality from `TracingPublisher` into `TracingPublisherFactory`.

The Trace information will now get into a standard `Publisher` through a builder callback `PubsubMessage.Builder.setTransform()`. 

This PR is a pre-requisite for fixing #864.

Fixes #790


elefeint:
I think I might be able to remove `TracingPublisherFactory` and the new `PublisherFactory` constructor if I allow an extra customizable field in `DefaultPublisherFactory`.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -347,7 +348,8 @@ public PublisherFactory defaultPublisherFactory(
       @Qualifier("publisherBatchSettings") ObjectProvider<BatchingSettings> batchingSettings,
       @Qualifier("publisherRetrySettings") ObjectProvider<RetrySettings> retrySettings,
       @Qualifier("publisherTransportChannelProvider")
-          TransportChannelProvider publisherTransportChannelProvider) {
+          TransportChannelProvider publisherTransportChannelProvider,
+      Optional<PublisherCustomizer> customizer) {
```
meltsufin, line 352: Why not `ObjectProvider` like the other arguments? Also, do we need to add the `@Qualifier` like for the others.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/publisher/PublisherCustomizer.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.spring.pubsub.core.publisher;
+
+import com.google.cloud.pubsub.v1.Publisher;
+
+public interface PublisherCustomizer {
```
meltsufin, line 21: This needs a javadoc.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/publisher/PublisherCustomizer.java
```diff
@@ -0,0 +1,23 @@
...
+  public void apply(Publisher.Builder builder, String topic);
```
meltsufin, line 22:
```suggestion
  public void apply(Publisher.Builder publisherBuilder, String topic);
```

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
```diff
@@ -1285,6 +1294,26 @@ void createSubscriberStub_flowControlSettings_noPropertiesSet() {
...
+      builder.setBatchingSettings(testBatchingSettings);
+    };
+
+    baseContextRunner
+        .withBean(PublisherCustomizer.class, () -> customizer)
+        .run(ctx -> {
+          PublisherFactory factory =
+              ctx.getBean("defaultPublisherFactory", PublisherFactory.class);
+          Publisher testPublisher = factory.createPublisher("unused");
+          assertThat(testPublisher.getBatchingSettings()).isSameAs(testBatchingSettings);
```
meltsufin, line 1313: Should you also verify that endpoint is `"unused:433"`?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/TracePubSubBeanPostProcessor.java
```diff
@@ -35,10 +33,7 @@
   @Override
   public Object postProcessBeforeInitialization(Object bean, String beanName)
       throws BeansException {
-    if (bean instanceof PublisherFactory) {
-      return new CachingPublisherFactory(
```
meltsufin, line 39: Would it make sense to just register the customizer here instead of autoconfig? I'm not sure what the tradeoffs are here.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/TracePubSubAutoConfiguration.java
```diff
@@ -50,4 +55,14 @@ static TracePubSubBeanPostProcessor tracePubSubBeanPostProcessor(BeanFactory bea
   PubSubTracing pubSubTracing(MessagingTracing messagingTracing) {
     return PubSubTracing.newBuilder(messagingTracing).build();
   }
+
+  @Bean
+  @ConditionalOnMissingBean
+  PublisherCustomizer publisherCustomizer(PubSubTracing pubSubTracing) {
```
meltsufin, line 61: This is specifically the `tracePublisherCustomizer`. So, I think it needs to be named that way.

meltsufin (changes requested):
My main concern here is that the `PublisherCustomizer` is part of the public API now, and nothing really requires it to be used for trace instrumentation only. There is a risk that users will provide a `PublisherCustomizer` and inadvertently disable tracing during publishing. You might be able to get around this by supporting multiple `PublisherCustomizer` beans, but then things get a bit more complicated.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
elefeint: `ObjectProvider` is useful when you don't know whether you have one or multiple.

meltsufin:
@elefeint Maybe what you can do is just keep `TracingPublisherFactory`, but add the customizer to it rather than the `DefaultPublisherFactory`.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
elefeint: Qualifier is there for non-pubsub-specific types, to distinguish, for example, a Spanner `RetrySettings` object from a Pub/Sub one. `PublisherCustomizer` is already a pubsub-specific interface.

meltsufin: Why do you think there will never be multiple? It's a very generic customizer.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/trace/pubsub/TracePubSubBeanPostProcessor.java
elefeint: That would address the concern of users potentially overriding trace; would also keep all trace-related config in post-processor. I'll try this.

elefeint: As discussed -- tried it, but because only `Publisher.Builder` can have a transformer injected, we'd have to unwrap `DefaultPublisherFactory` and inject a customizer here. It's doable but awkward.

meltsufin (approved):
Looks good, aside from missing documentation, which you might want to do in a separate PR. Up to you.

meltsufin (approved):

PR #906, opened 2022-01-25
Differentiate scheduled 2.x job cache from main cache

elefeint:
I think one of the reasons the flakes got more frequent is that the 2.x unit/integration tests run first, and set up the older set of artifacts. Then all the new code (`main` branch and PRs off of it) has to re-download Spring Boot 2.6-related artifacts on the fly.

This PR makes 2.x unit and integration tests use different caches.

After this if, once we start backporting changes to 2.x, we notice that 2.x flakes more than `main`, it will be confirmation of this hypothesis.

### DIFF
--- a/.github/workflows/integrationTests-2.x.yaml
+++ b/.github/workflows/integrationTests-2.x.yaml
@@ -49,7 +49,7 @@ jobs:
         id: mvn-cache
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-unified-${{ steps.date.outputs.date }}
+          key: ${{ runner.os }}-maven-unified-2.x-${{ steps.date.outputs.date }}
       - name: Setup gcloud
         uses: google-github-actions/setup-gcloud@master
         with:
--- a/.github/workflows/unitTests-2.x.yaml
+++ b/.github/workflows/unitTests-2.x.yaml
@@ -31,7 +31,7 @@ jobs:
       id: mvn-cache
       with:
         path: ~/.m2/repository
-        key: ${{ runner.os }}-maven-unified-${{ steps.date.outputs.date }}
+        key: ${{ runner.os }}-maven-unified-2.x-${{ steps.date.outputs.date }}
 
     - name: Maven go offline
       id: mvn-offline


meltsufin (approved):
LGTM.
Side question: Why do we even have the `integrationTests-2.x.yaml`? Is it not possible to include the 2.x branch in the main `integrationTests.yaml`?

elefeint:
It's for the scheduled job, so the right branch is checked out. The non-scheduled PR/branch integration tests are handled with the regular configuraion.

mpeddada1 (approved):

meltsufin:
@elefeint Is there a reason you can't apply the same script to multiple branches like explained [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore)?

elefeint:
Yes; because this documentation for non-scheduled events "When using the pull_request and pull_request_target". Pull requests (by way of branch pushes) [are already handled](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/.github/workflows/integrationTests.yaml#L4-L7) by the standard `integrationTests.yaml`

However, if you look at [on.schedule documentation[(https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule), it says _"Scheduled workflows run on the latest commit on the default or base branch. "_. So it's not possible to run a scheduled task on a different branch, and the only way to run integration tests on `2.x` on schedule, and not just on push, is to check out `2.x` through the `actions/checkout@v2` action, which does support different branches/tags.

Here are some relevant github threads:
https://github.community/t/scheduled-builds-of-non-default-branch/16306/22
https://github.community/t/on-schedule-per-branch/17525/2

meltsufin:
@elefeint Thanks for the refs. What I was also thinking was along the lines of using the matrix. See this comment: https://github.community/t/scheduled-builds-of-non-default-branch/16306/14.

elefeint:
That's something we could certainly do, as long as the new matrix dimension makes it into the cache key. Do you want to send a PR?

PR #912, opened 2022-01-26
Replace 2.x CI config with matrix

meltsufin:
Follow-up to https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/906#issuecomment-1022309720.


meltsufin:
This doesn't seem to work well because now all PRs would cause both branches to be tested. Additionally, config for each branch may diverge in the future making it hard to keep in the same file.

PR #914, opened 2022-01-27
Remove Spring Milestones/Snapshots repositories

elefeint:
We don't rely on snapshots/milestones in the releasable branches.

This may help with the flakes -- I suspect that maven's multi-threading tries to fetch jars from Maven Central and from Spring Milestones at the same time, but milestones don't actually have the artifacts that cause issues.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,1 +1,0 @@
--P spring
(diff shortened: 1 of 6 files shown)


meltsufin (commented):
Can we entirely remove the `spring` profile?
Spring releases should be in Maven Central. No?

elefeint:
Except for experimental modules/plugins like spring-native.
I am happy to remove the global profiles and restrict spring-releases to only the places that require them!

meltsufin (approved):

PR #923, opened 2022-01-28
Autoconfigure R2DBC support in Cloud Sql (backport)

elefeint:
Backports #772.


meltsufin (approved):
Since this is not a bug fix, I'm not sure you need to backport it, but it's up to you.

PR #937, opened 2022-02-03
allow user override configurations on Gson object for JSON conversion

### LINKED ISSUE #650: Enhancement to spring data spanner JSON support - allow setting changes on Gson object
This is a follow up to #635. To allow users the flexibility to configure on Gson object used in JSON conversion, we should make it a bean in auto-configuration, and inject in where it's needed in write and read.

_Originally discussed in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/635#discussion_r725072095_

zhumin8:
TODO:
add description.
add doc.
add test? 
cleanup spannerMappingCTX creation in testing.

This should close #650

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
@@ -104,11 +104,17 @@ public class StructAccessor {
 
   private Set<String> columnNamesIndex;
 
-  private static final Gson gson = new Gson();
+  private Gson gson;
 
   public StructAccessor(Struct struct) {
     this.struct = struct;
     this.columnNamesIndex = indexColumnNames();
+  }
+
+  public StructAccessor(Struct struct, Gson gson) {
+    this.struct = struct;
+    this.columnNamesIndex = indexColumnNames();
+    this.gson = gson;
   }
 
   Object getSingleValue(String colName) {
--- a/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerMappingContext.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerMappingContext.java
@@ -16,7 +16,9 @@
 
 package com.google.cloud.spring.data.spanner.core.mapping;
 
+import com.google.gson.Gson;
 import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.data.mapping.context.AbstractMappingContext;
@@ -25,6 +27,7 @@ import org.springframework.data.mapping.model.Property;
 import org.springframework.data.mapping.model.PropertyNameFieldNamingStrategy;
 import org.springframework.data.mapping.model.SimpleTypeHolder;
 import org.springframework.data.util.TypeInformation;
+import org.springframework.stereotype.Component;
 
 /**
  * A mapping context for Cloud Spanner that provides ways to create persistent entities and
@@ -32,6 +35,7 @@ import org.springframework.data.util.TypeInformation;
  *
  * @since 1.1
  */
+@Component
 public class SpannerMappingContext
     extends AbstractMappingContext<SpannerPersistentEntity<?>, SpannerPersistentProperty>
     implements ApplicationContextAware {
@@ -42,6 +46,13 @@ public class SpannerMappingContext
   private FieldNamingStrategy fieldNamingStrategy = DEFAULT_NAMING_STRATEGY;
 
   private ApplicationContext applicationContext;
+
+  private Gson gson;
+
+  @Autowired
+  public void setGson(Gson gson) {
+    this.gson = gson;
+  }
 
   /**
    * Set the field naming strategy used when creating persistent properties.
@@ -75,7 +86,7 @@ public class SpannerMappingContext
 
   protected <T> SpannerPersistentEntityImpl<T> constructPersistentEntity(
       TypeInformation<T> typeInformation) {
-    return new SpannerPersistentEntityImpl<>(typeInformation);
+    return new SpannerPersistentEntityImpl<>(typeInformation, this.gson);
   }
 
   @Override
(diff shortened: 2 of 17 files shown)


spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerMappingContext.java
```diff
@@ -43,6 +46,19 @@
 
   private ApplicationContext applicationContext;
 
+  private Gson gson;
+
+  public SpannerMappingContext() {
+  }
+
+  public SpannerMappingContext(Gson gson) {
```
elefeint, line 54: Could you add a non-null check here? And then `getGson()` can be annotated with `@NonNull`, so the invocations are known to be safe.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -91,18 +91,6 @@
 
   private final Set<Class<?>> jsonProperties = new HashSet<>();
 
-  /**
-   * Creates a {@link SpannerPersistentEntityImpl}.
-   *
-   * @param information type information about the underlying entity type.
-   */
-  public SpannerPersistentEntityImpl(TypeInformation<T> information) {
```
elefeint, line 99: It's a public method, unfortunately, so deprecate it instead. We can remove it in the next major release.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
```diff
@@ -104,13 +104,19 @@
 
   private Set<String> columnNamesIndex;
 
-  private static final Gson gson = new Gson();
+  private Gson gson;
 
   public StructAccessor(Struct struct) {
```
elefeint, line 109: I was looking for where the single-arg version was still used, and now I wonder if there is an issue with `ConverterAwareMappingSpannerEntityReader` not being able to process JSON fields. Could you figure out what path `ConverterAwareMappingSpannerEntityReader.read()` goes through?

My original thought was to deprecate the single-arg version and stop using it, since having Gson object be null in some cases but not others could easily lead to bugs.

elefeint (changes requested):
Nice cleanup along the way!

I only have one concern --  it's about the possibility of Gson object being null in some but not all of struct readers.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
zhumin8: hmm, the single-arg version is now only used in `SqlSpannerQuery` [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/98756c2ffd27ad207aca405290d90acf33a2752c/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SqlSpannerQuery.java#L241). it will make no difference to functionality to pass a second gson parameter in. So I think deprecating makes sense.

As for `ConverterAwareMappingSpannerEntityReader.read()` , it eventually processes each property of entity in `StructPropertyValueProvider` and JSON fields are read with [this line](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructPropertyValueProvider.java#L115). `this.structAccessor` here is passed down from `ConverterAwareMappingSpannerEntityReader` which has access to `SpannerMappingContext` and thus `Gson` bean.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerMappingContext.java
```diff
@@ -75,7 +91,9 @@ public FieldNamingStrategy getFieldNamingStrategy() {
 
   protected <T> SpannerPersistentEntityImpl<T> constructPersistentEntity(
       TypeInformation<T> typeInformation) {
-    return new SpannerPersistentEntityImpl<>(typeInformation);
+    return new SpannerPersistentEntityImpl<>(typeInformation, this,
+        this.applicationContext == null ? new ConverterAwareMappingSpannerEntityProcessor(this)
+            : this.applicationContext.getBean(SpannerEntityProcessor.class));
```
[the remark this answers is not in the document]
zhumin8: Expanded this into if-else blocks. 
I thought of getting rid of the if-else block completely and replace with non-null checks, my reasonings:
`this.applicationContext` is set from `ApplicationContextAware` in `SpannerContextMapping` bean post creation process.
So by the time `constructPersistentEntity()` is called, we are in post process of entity bean creations, and `this.applicationContext` and `this.applicationContext.getBean(SpannerEntityProcessor.class)` should both be present already. 
But eventually dial back to the current if-else because the change breaks a lot of unit tests and setting up mock applicationContext only fixes some of them. So I feel that this removal probably should belong to a separate future pr along with some test cleanups. what are your thoughts?

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerMappingContext.java
```diff
@@ -91,9 +95,15 @@ public FieldNamingStrategy getFieldNamingStrategy() {
...
   protected <T> SpannerPersistentEntityImpl<T> constructPersistentEntity(
       TypeInformation<T> typeInformation) {
-    return new SpannerPersistentEntityImpl<>(typeInformation, this,
-        this.applicationContext == null ? new ConverterAwareMappingSpannerEntityProcessor(this)
-            : this.applicationContext.getBean(SpannerEntityProcessor.class));
+    SpannerEntityProcessor processor;
+    if (this.applicationContext == null) {
+      processor = new ConverterAwareMappingSpannerEntityProcessor(this);
+    } else if (!this.applicationContext.containsBean("spannerConverter")) {
+      processor = new ConverterAwareMappingSpannerEntityProcessor(this);
```
elefeint, line 102: These can be collapsed into `if (A || B)` ingle branch, since the action is the same

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerMappingContext.java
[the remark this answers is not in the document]
elefeint: Yeah, keep it. It's easy to follow now, and prevents future issues.

elefeint (approved):

elefeint (approved):

Issue #943, opened 2022-02-10
preparePartTreeSqlTagParameterMap breaks if the Pageable parameter is not the last


amrnarvar:
**Describe the bug**
Please provide details of the problem, including the version of Spring Cloud that you
are using. 
spring cloud version 3.1.0
**Sample**
Page<ConsumerEvent> findAllByXAndY(
      Pageable pageable, String X, String Y)
If possible, please provide a test case or sample application that reproduces
the problem. This makes it much easier for us to diagnose the problem and to verify that
we have fixed it.
```
SpannerStatementQueryExecutor
  private static Map<String, Parameter> preparePartTreeSqlTagParameterMap(
```

The code should maintain an independent variable for `sqlStringAndPlaceholders.getPlaceholders`, otherwise the code will break if the Sort and Page parameters are not last.

zhumin8:
Thanks for catching this and pointing to the method that breaks. We'll fix this soon. 
For the time being, you are right Pageable parameter need to be the last parameter for it to work.

PR #958, opened 2022-02-18
Fix query breaks when Pageable parameter is not the last

zhumin8:
This is a quick fix for issue #943 to correctly skip Pageable/Sort when it might not be the last parameter.

Note for future refactoring: what case makes `originalParam == null` [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/repository/query/SpannerStatementQueryExecutor.java#L459)? It's never used in current test cases. might worth look into remove building of `paramMetadataMap`  if not used.


elefeint (changes requested):
I think it would be easier to just keep track of the next placeholder index. Then you would not have to worry about multiple skippable parameters at all.

```
   int placeholderIndex = 0;
    for (int i = 0; i < paramsMetadata.length; i++) {
      Parameter param = paramsMetadata[i];
      // Skip Pageable and Sort parameters because they don't need to be bound to the tags in the
      // query.
      // They are processed separately in applySort and buildLimit methods.
      if (param.getType() != Pageable.class && param.getType() != Sort.class) {
        paramMetadataMap.put(sqlStringAndPlaceholders.getPlaceholders().get(placeholderIndex), param);
        placeholderIndex++;
      }
    }
```

elefeint (approved):

Issue #844, opened 2022-01-07
Support "Annotate File" endpoint


gabac:
**Is your feature request related to a problem? Please describe.**
Google Vision also offers an API to annotate files, specifically PDFs. In my use case I have many file types and I would like not to convert a PDF to an image on my side

- https://cloud.google.com/vision/docs/reference/rest/v1/files/annotate
- https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.BatchAnnotateFilesRequest

**Describe the solution you'd like**
Support this additional endpoint as well.

**Describe alternatives you've considered**
Convert the file in my custom service

**Additional context**
Would be happy to provide a PR if you think this is a sensible addition

meltsufin:
I think this could make sense, especially if it's a simple addition to the `VisionTemplate`. @dzou @elefeint What do you think?

gabac:
That would have been my idea 💡

elefeint:
@gabac Have you tried using `CoudVisionTemplate.analyzeImage()` on the PDFs? PDF is [considered one of the supported "image" types](https://cloud.google.com/vision/docs/supported-files).

gabac:
I tried, it's not 😉
That's why I opened up the feature request in the first place

gabac:
That would be the error message I receive
```
code: 13
message: "Internal server error. Failed to process features."
```

elefeint:
We've tested this a bit with the existing Vision sample. Could not reproduce this exact error, but did get "bad image" response codes for every PDF tried. 
I've reached out to Vision team for guidance on whether the "annotate-image" set of APIs is expected to work on PDFs, or if "annotate-files" should be used instead.

gabac:
This is my mistake. I copied in an internal error message from my service. The error message by the API would be the same as you just mentioned
```
{
  "responses": [
    {
      "error": {
        "code": 3,
        "message": "Bad image data."
      }
    }
  ]
}

```
I appreciate that you reached out to the Vision team to clarify things.

Independent of the outcome, would you be interested in supporting the `annotate-files` endpoint in this library?

elefeint:
@gabac I've confirmed that PDF, TIFF and GIF files are expected to use `BatchAnnotateFiles` endpoint. 

If you are interested in sending a PR, we'll be happy to review and accept it.

meltsufin:
[...]

PR #966, opened 2022-02-28
Add support for Vision BatchAnnotateFiles endpoint

florian-3ap:
fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/844


spring-cloud-gcp-vision/src/main/java/com/google/cloud/spring/vision/CloudVisionTemplate.java
```diff
@@ -78,6 +94,31 @@ public String extractTextFromImage(Resource imageResource, ImageContext imageCon
     return result;
   }
 
+  /**
+   * Extract the text out of a file and return the result as a String.
+   *
+   * @param fileResource the file one wishes to analyze
+   * @param mimeType the mime type of the fileResource
```
meltsufin, line 101: Either list the supported mime types or link to docs that do.

spring-cloud-gcp-vision/src/main/java/com/google/cloud/spring/vision/CloudVisionTemplate.java
```diff
@@ -78,6 +94,31 @@ public String extractTextFromImage(Resource imageResource, ImageContext imageCon
...
+   * @return the text extracted from the pdf aggregated to a String
+   * @throws CloudVisionException if the image could not be read or if text extraction failed
+   */
+  public String extractTextFromFile(Resource fileResource, String mimeType) {
+    AnnotateFileResponse response = analyzeFile(fileResource, mimeType, Type.TEXT_DETECTION);
+
+    List<AnnotateImageResponse> annotateImageResponses = response.getResponsesList();
```
meltsufin, line 108: The 10 lines of code below are the same as what we need to do for `extractTextFromImage` and should be refactored into a shared private utility method.

meltsufin (changes requested):
Thanks for the contribution! It looks very reasonable. I just added a few minor comments.

zhumin8 (changes requested):
Thanks for contributing! Code looks good in general. 
I only have one concern about `extractTextFromFile()` that file can have multiple pages, details in comment below. 
Additionally, I would suggest to add an integration test or a case in the existing sample code if possible.

spring-cloud-gcp-vision/src/main/java/com/google/cloud/spring/vision/CloudVisionTemplate.java
florian-3ap: added the supported types to the doc

spring-cloud-gcp-vision/src/main/java/com/google/cloud/spring/vision/CloudVisionTemplate.java
```diff
@@ -78,6 +94,31 @@ public String extractTextFromImage(Resource imageResource, ImageContext imageCon
...
+    if (annotateImageResponses.isEmpty()) {
+      throw new CloudVisionException(
+          "Failed to receive valid response Vision APIs; empty response received.");
+    }
+
+    String result = annotateImageResponses.get(0).getFullTextAnnotation().getText();
```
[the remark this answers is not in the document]
florian-3ap: I changed the return value to a list of strings. This should now contain all the extracted text for all the pages

florian-3ap:
> Thanks for contributing! Code looks good in general.
> I only have one concern about `extractTextFromFile()` that file can have multiple pages, details in comment below.
> Additionally, I would suggest to add an integration test or a case in the existing sample code if possible.

I enhanced the `spring-cloud-gcp-vision-api-sample` with a PDF example. It's now working with single page and also multi page PDFs.

spring-cloud-gcp-vision/src/test/java/com/google/cloud/spring/vision/CloudVisionTemplateTests.java
```diff
@@ -162,6 +211,15 @@ public void testResourceReadingError() {
     this.cloudVisionTemplate.analyzeImage(new BadResource(), Type.LABEL_DETECTION);
   }
 
+  @Test
+  public void testFileResourceReadingError() {
+    this.expectedException.expect(CloudVisionException.class);
+    this.expectedException.expectMessage("Failed to read image bytes from provided resource.");
```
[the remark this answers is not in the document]
florian-3ap: added the error message to constants

meltsufin (changes requested):
I think the code looks good, except for one minor comment that's unresolved.

If it's not too much to ask, two more things would be ideal:
1) Can you add an integration test for the sample endpoint?
2) Reference documentation in `vision.adoc` should be updated to describe the new functionality.

Thanks!

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/resources/static/index.html
```diff
@@ -31,5 +31,17 @@ <h1>Text Extraction</h1>
     </form>
 </div>
 
+<div>
+    <h1>Text Extraction PDF</h1>
+    <p>Read and extract the text from a PDF:</p>
+    <form action="/extractTextFromPdf">
+        Web URL of a PDF to analyze:
```
zhumin8, line 38: This might be a good place to also mention this is meant for small PDF files less than 5 pages.

zhumin8 (changes requested):
The added sample looks great. Made a few minor comments below. 
Can you please also add to the existing Vision doc [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/docs/src/main/asciidoc/vision.adoc#image-analysis) with the newly added functionalities in `CloudVisionTemplate`? This way they are discoverable by more users.

spring-cloud-gcp-vision/src/main/java/com/google/cloud/spring/vision/CloudVisionTemplate.java
florian-3ap: With the latest changes, the 10 lines below are not the exact same anymore. The return value changed now to a list of strings instead of a single string. Do you think there are still some lines left that are worth moving into a separate utility method?

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/resources/static/index.html
florian-3ap: I added a small comment above regarding the maximum pages.

spring-cloud-gcp-vision/src/main/java/com/google/cloud/spring/vision/CloudVisionTemplate.java
```diff
@@ -143,8 +192,59 @@ public AnnotateImageResponse analyzeImage(
...
+
+    List<Feature> featureList =
+        Arrays.stream(featureTypes)
+            .map(featureType -> Feature.newBuilder().setType(featureType).build())
+            .collect(Collectors.toList());
+
+    BatchAnnotateFilesRequest request =
+        BatchAnnotateFilesRequest.newBuilder()
+            .addRequests(
+                AnnotateFileRequest.newBuilder()
```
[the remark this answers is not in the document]
florian-3ap: I mentioned it above in the method doc and also added it to the `vision.adoc` documentation

spring-cloud-gcp-vision/src/main/java/com/google/cloud/spring/vision/CloudVisionTemplate.java
meltsufin: This actually got me thinking, for the sake of simplicity of use, maybe we should just return a single concatenated string, separated by newlines, or maybe a customizable separator that can be configured on the class level. What do you think?

docs/src/main/asciidoc/vision.adoc
```diff
@@ -99,6 +99,35 @@ For example, if you analyzed an image using the `LABEL_DETECTION` feature, you w
 `AnnotateImageResponse` is provided by the Google Cloud Vision libraries; please consult the https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.AnnotateImageResponse[RPC reference] or https://googleapis.github.io/googleapis/java/all/latest/apidocs/com/google/cloud/vision/v1/AnnotateImageResponse.html[Javadoc] for more details.
 Additionally, you may consult the https://cloud.google.com/vision/docs/[Cloud Vision docs] to familiarize yourself with the concepts and features of the API.
 
+=== PDF, TIFF and GIF Analysis
+
+The `CloudVisionTemplate` allows you to easily analyze PDF, TIFF and GIF documents; it provides the following method for interfacing with Cloud Vision:
+
+`public AnnotateFileResponse analyzeFile(Resource fileResource, String mimeType, Feature.Type... featureTypes)`
```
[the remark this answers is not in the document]
florian-3ap: I did not include it because `extractTextFromImage` is also not part of the documentation

spring-cloud-gcp-vision/src/main/java/com/google/cloud/spring/vision/CloudVisionTemplate.java
florian-3ap: In my opinion, this really depends on what the caller is doing with the extracted text. If you want to process the text page by page it does not make too much sense to merge it into one string. But by returning a list of strings you leave the option open to the caller how he then wants to handle the response.

docs/src/main/asciidoc/vision.adoc
```diff
@@ -99,6 +99,35 @@ For example, if you analyzed an image using the `LABEL_DETECTION` feature, you w
 `AnnotateImageResponse` is provided by the Google Cloud Vision libraries; please consult the https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.AnnotateImageResponse[RPC reference] or https://googleapis.github.io/googleapis/java/all/latest/apidocs/com/google/cloud/vision/v1/AnnotateImageResponse.html[Javadoc] for more details.
 Additionally, you may consult the https://cloud.google.com/vision/docs/[Cloud Vision docs] to familiarize yourself with the concepts and features of the API.
 
+=== PDF, TIFF and GIF Analysis
```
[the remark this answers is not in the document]
florian-3ap: Sounds good. I moved the section and renamed the title. Also added a small `Running Text Detection Example`

docs/src/main/asciidoc/vision.adoc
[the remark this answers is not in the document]
meltsufin: Fair.

spring-cloud-gcp-vision/src/main/java/com/google/cloud/spring/vision/CloudVisionTemplate.java
meltsufin: I was just wondering if the most common usecase would be to process only 1 page anyway. In that case, a list would make it a bit more verbose to use. Which option seems more convenient for your usecase, for example?

meltsufin (approved):
@florian-3ap Thank you so much for this contribution and your patience with the reviews!

zhumin8 (approved):
Doc addition looks great to me! Thank you for contributing!

PR #977, opened 2022-03-04
Migrating tests to JUnit5

ddixit14:
Migrated "UserAgentHeaderProviderIntegrationTests.java" and "OnGcpEnvironmentConditionTests.java" to JUnit5

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/environment/OnGcpEnvironmentConditionTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/environment/OnGcpEnvironmentConditionTests.java
@@ -23,13 +23,10 @@ import static org.mockito.Mockito.when;
 import com.google.cloud.spring.core.GcpEnvironment;
 import com.google.cloud.spring.core.GcpEnvironmentProvider;
 import java.util.Collections;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
-import org.mockito.junit.jupiter.MockitoSettings;
-import org.mockito.quality.Strictness;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.boot.autoconfigure.condition.ConditionOutcome;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/environment/OnGcpEnvironmentConditionTests.java
```diff
@@ -18,16 +18,19 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.lenient;
```
zhumin8, line 21: This import is unused.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/environment/OnGcpEnvironmentConditionTests.java
```diff
@@ -52,15 +55,16 @@
 
   @Mock GcpEnvironmentProvider mockGcpEnvironmentProvider;
 
-  @Before
-  public void setUp() {
+  @BeforeEach
+  void setUp() {
     when(this.mockContext.getBeanFactory()).thenReturn(this.mockBeanFactory);
     when(this.mockBeanFactory.getBean(GcpEnvironmentProvider.class))
```
zhumin8, line 61: This might be nit, but I feel like there is a slight advantage to have mock initialized with `Mockito.lenient()` here instead of having the `@MockitoSettings(strictness = Strictness.LENIENT)` annotations on each test. This way it's more clear to me which mock is optional and allowed to be not called in some of the test methods. What do you think?
```suggestion
    Mockito.lenient().when(this.mockBeanFactory.getBean(GcpEnvironmentProvider.class))
```

zhumin8 (changes requested):
Learning about  MockitoExtension `UnnecessaryStubbingException` with your PR. I have one minor comment that may make maintenance a bit easier later.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/environment/OnGcpEnvironmentConditionTests.java
ddixit14: I would have preferred this if all the tests had UnnecessaryStubbingException. Since only half of them have it (and the other half works fine with strict stubbing), I think it is better to not initialize mockito with lenient settings. Strict stubbing is actually preferable over lenient one as it contributes towards a cleaner and more maintainable codebase so the more we have of it, the better. Moreover, all the spring-cloud-gcp tests have mock initialized normally so this would maintain the uniformity as well. What do you think?

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/environment/OnGcpEnvironmentConditionTests.java
ddixit14: Removed it.

zhumin8 (approved):

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/core/environment/OnGcpEnvironmentConditionTests.java
ddixit14: Let me try removing the lenient annotation for this one. I'm working on it.

zhumin8 (approved):

elefeint (approved):
Very good commit messages! Thank you for figuring out which mock setup was necessary for which test.

PR #981, opened 2022-03-07
Migrating "DatastoreHealthIndicatorTests.java" to JUnit5

ddixit14:
Migrating the test.


elefeint (changes requested):
This test is a good opportunity to discuss what is the point of unit testing -- why do we do it? What purpose does each test serve? This is intentionally vague; see the comment in the test class for the specific next steps.

elefeint (approved):

PR #988, opened 2022-03-07
Migrating "IapAuthenticationAutoConfigurationTests.java" to JUnit5

ddixit14:
Removing @ Rule annotation.


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
```diff
@@ -88,33 +86,31 @@
 
   @Mock static MetadataProvider mockMetadataProvider;
 
-  @Before
-  public void httpRequestSetup() {
+  @BeforeEach
+  void httpRequestSetup() {
     when(this.mockIapRequest.getHeader("x-goog-iap-jwt-assertion")).thenReturn("very fake jwt");
```
elefeint, line 91: By the time we have to enable LENIENT mode on half the tests, it makes sense to move this setup from being global to only being done for each test that needs it. Especially since the setup is only one line.

elefeint (changes requested):
To deal with half the tests not needing the `mockIapRequest` header setup, remove the `@BeforeEach` and put the when/then setup into each test that needs it. Then Lenient mode can go away.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
ddixit14: Removed the global set up and tried to put it in the tests that needs it along with removing the lenient mode as well but it is still giving "UnnecessaryStubbing" error.

elefeint: Did you add setup to the tests _with_ lenient or to the tests _without_? 
The setup should be needed in tests that did not have the lenient setup.

elefeint (approved):

PR #990, opened 2022-03-08
Migrating tests in spring-cloud-gcp-bigquery to JUnit5

ddixit14:
Migrated the following tests to JUnit5:

1.  BigQueryTemplateIntegrationTests.java
2.  BigQueryFileMessageHandlerIntegrationTests.java
3. BigQueryFileMessageHandlerTests.java

Introduced **@ TestInstance(TestInstance.Lifecycle.PER_CLASS)** annotation in **BigQueryTemplateIntegrationTests** to enable **@ AfterAll annotation** to work with non-static fields. 
**@ TestInstance(TestInstance.Lifecycle.PER_CLASS)** enables us to ask JUnit5 to create only one instance of the test class and reuse it between tests. This enables the usage of @ Autowired fields with @ AfterAll methods (otherwise not possible because @ AfterAll is used in static context and @ Autowired doesn't work with static)

### DIFF
--- a/spring-cloud-gcp-bigquery/src/test/java/com/google/cloud/spring/bigquery/core/BigQueryTemplateIntegrationTests.java
+++ b/spring-cloud-gcp-bigquery/src/test/java/com/google/cloud/spring/bigquery/core/BigQueryTemplateIntegrationTests.java
@@ -18,8 +18,6 @@ package com.google.cloud.spring.bigquery.core;
 
 import static com.google.cloud.spring.bigquery.core.BigQueryTestConfiguration.DATASET_NAME;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.hamcrest.Matchers.is;
-import static org.junit.Assume.assumeThat;
 
 import com.google.cloud.bigquery.BigQuery;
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-bigquery/src/test/java/com/google/cloud/spring/bigquery/core/BigQueryTemplateIntegrationTests.java
```diff
@@ -35,26 +33,30 @@
...
+import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.util.concurrent.ListenableFuture;
 
 /**
  * Integration tests for BigQuery.
  *
  * @since 1.2
  */
-@RunWith(SpringRunner.class)
+//Please create a table "test_dataset" in BigQuery to run the tests successfully
```
elefeint, line 54: Could you add this to "Running integration tests locally" g3doc?

elefeint (approved):
Using one per-class JUnit instance is an interesting approach. I did not know about this, but it does seem to be a good fit for these types of integration tests. It should be safe, since the table is cleared before each test.

Issue #982, opened 2022-03-07
Spring Cloud refresh causes CloudSqlEnvironmentPostProcessor to throw IllegalStateException


stefanscheidt:
When exposing the actuator endpoint `refresh` provided by Spring Cloud, calling this endpoint will lead to a re-execution of `CloudSqlEnvironmentPostProcessor:: postProcessEnvironment`. At the end of this method, the static method `CoreSocketFactory::setApplicationName` will be called. In this state, the `coreSocketFactory` singleton inside of `CoreSocketFactory` will not be null, thus `CoreSocketFactory::setApplicationName` will throw an `IllegalStateException("Unable to set ApplicationName - SQLAdmin client already initialized.")`.

Stacktrace:

```
java.lang.IllegalStateException: Unable to set ApplicationName - SQLAdmin client already initialized.
	at com.google.cloud.sql.core.CoreSocketFactory.setApplicationName(CoreSocketFactory.java:449)
	at com.google.cloud.spring.autoconfigure.sql.CloudSqlEnvironmentPostProcessor.postProcessEnvironment(CloudSqlEnvironmentPostProcessor.java:92)
	at org.springframework.cloud.context.refresh.ConfigDataContextRefresher.updateEnvironment(ConfigDataContextRefresher.java:92)
	at org.springframework.cloud.context.refresh.ContextRefresher.refreshEnvironment(ContextRefresher.java:101)
	at org.springframework.cloud.context.refresh.ContextRefresher.refresh(ContextRefresher.java:94)
	at org.springframework.cloud.endpoint.RefreshEndpoint.refresh(RefreshEndpoint.java:41)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.springframework.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:282)
	at org.springframework.boot.actuate.endpoint.invoke.reflect.ReflectiveOperationInvoker.invoke(ReflectiveOperationInvoker.java:74)
	at org.springframework.boot.actuate.endpoint.annotation.AbstractDiscoveredOperation.invoke(AbstractDiscoveredOperation.java:60)
[...]

PR #1002, opened 2022-03-14
fix: refresh causes CloudSqlEnvironmentPostProcessor to throw IllegalStateException

meltsufin:
This is just a workaround because we're not able to refresh `CoreSocketFactory`.

Fixes: #982.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
@@ -80,11 +80,15 @@ public class CloudSqlEnvironmentPostProcessor implements EnvironmentPostProcesso
           .getPropertySources()
           .addFirst(new MapPropertySource("CLOUD_SQL_DATA_SOURCE_URL", primaryMap));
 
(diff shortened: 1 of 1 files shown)


elefeint (changes requested):
Would it make sense to log a warning, so users are not mislead about this working?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessor.java
```diff
@@ -80,11 +80,15 @@ public void postProcessEnvironment(
...
-      CredentialsPropertiesSetter.setCredentials(sqlProperties, propertiesRetriever.getGcpProperties());
+      CredentialsPropertiesSetter.setCredentials(sqlProperties,
+          propertiesRetriever.getGcpProperties());
 
       // support usage metrics
-      CoreSocketFactory.setApplicationName(
-          "spring-cloud-gcp-sql/" + this.getClass().getPackage().getImplementationVersion());
+      String applicationName =
+          "spring-cloud-gcp-sql/" + this.getClass().getPackage().getImplementationVersion();
+      if (!CoreSocketFactory.getApplicationName().contains(applicationName)) {
```
stefanscheidt, line 89: Hi. I'm just wondering if this kind of logic actually belongs inside of `CoreSocketFactory::setApplicationName`. Just a thought ...

PR #1006, opened 2022-03-15
Migrating tests to JUnit5 (spring-cloud-gcp-data-spanner) 

ddixit14:
Tests Migrated:

1. SpannerMutationFactoryImplTests.java
2. SpannerReadOptionsTests.java
3. SpannerSortPageQueryOptionsTests.java
4. SpannerTemplateAuditingTests.java
5. SpannerTemplateTests.java
6. SpannerTemplateTransactionManagerTests.java
7. SpannerTransactionManagerTests.java

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/SpannerTemplateTests.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/SpannerTemplateTests.java
@@ -17,6 +17,7 @@
 package com.google.cloud.spring.data.spanner.core;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.isNull;
@@ -71,11 +72,10 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.function.Supplier;
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.rules.ExpectedException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 import org.mockito.InOrder;
 import org.mockito.Mockito;
 import org.springframework.context.ApplicationEvent;
@@ -84,12 +84,9 @@ import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 
 /** Tests for the Spanner Template. */
-public class SpannerTemplateTests {
+class SpannerTemplateTests {
 
   private static final Statement DML = Statement.of("update statement");
-
-  /** used for checking exception messages and tests. */
-  @Rule public ExpectedException expectedException = ExpectedException.none();
 
   private DatabaseClient databaseClient;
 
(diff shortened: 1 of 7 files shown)


spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/SpannerTemplateTests.java
```diff
@@ -287,49 +284,48 @@ public void readOnlyTransactionTest() {
     verify(readOnlyTransaction, times(2)).read(eq("custom_test_table"), any(), any());
   }
 
-  @Test
-  public void readOnlyTransactionDmlTest() {
 
-    this.expectedException.expectMessage("A read-only transaction template cannot execute DML.");
+  @Test
+  <T> void readOnlyTransactionDmlTest() {
```
elefeint, line 289: Generics are useful when the callers of this method may need to provide different types. Since this method is only called by the JUnit framework, and the test itself is defining the operation lambda, it is at this point known what the return type is. Well, kind of. The lambda returns `null`, so the type could be anything -- Long (because that would have been the result of `executeDmlStatement()`), Object (technically correct for any return value. In fact, the original code had Java infer Object as the type), or, my personal favorite when return value is `null` and return type does not matter, `Void`.

Looking at javadoc for `performReadOnlyTransaction`, it says "@param <T> the final return type of the operations." Since the return value of all these test lambdas is `null`, I'd use `Void` for the second parameterized type.

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/SpannerTemplateTests.java
```diff
@@ -287,49 +284,48 @@ public void readOnlyTransactionTest() {
...
-        new SpannerReadOptions().setTimestamp(Timestamp.ofTimeMicroseconds(333)));
+    SpannerReadOptions testSpannerReadOptions = new SpannerReadOptions().setTimestamp(Timestamp.ofTimeMicroseconds(333));
+    Function<SpannerTemplate, T> testSpannerOperations = spannerOperations -> {
+      spannerOperations.executePartitionedDmlStatement(Statement.of("fail"));
+      return null;
+    };
+
+    assertThatThrownBy(() -> this.spannerTemplate.performReadOnlyTransaction(
+            testSpannerOperations, testSpannerReadOptions))
+            .hasMessage("A read-only transaction template cannot execute partitioned" + " DML.");
```
elefeint, line 324: Could you remove the concatenation while you are here? I looked into how this came to be, and it was a result of original formatting being on two lines, and then getting joined to a single line when I ran autoformatting to google style.

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/SpannerTemplateTests.java
```diff
@@ -342,65 +338,62 @@ public void readWriteTransactionPartitionedDmlTest() {
...
-        null,
-        this.mutationFactory,
-        this.schemaUtils);
+    assertThatThrownBy(() ->  new SpannerTemplate(
+            () -> this.databaseClient,
+            this.mappingContext,
+            null,
+            this.mutationFactory,
+            this.schemaUtils))
+            .isInstanceOf(IllegalArgumentException.class)
```
elefeint, line 377: nit: Could you reformat this paragraph, so that the `isInstanceOf()` indentation is smaller than "carryover line" indentation? The easiest way is probably to split `))` to two different lines.

elefeint (changes requested):

elefeint (approved):

PR #1009, opened 2022-03-16
Migrating "KmsTemplateTests.java" to JUnit5

ddixit14:
Migrated "KmsTemplateTests.java" to JUnit5.

In both "testEncryptPermissionDenied()" and "testEncryptNotFound()", we have nothing to do with decryption, right? The exception thrown is in encryptText() method and for each of the tests we are setting up a new kmsTemplate, so what is the need for decryption?
I've omitted the decryption part for now, please let me know about your thoughts on this.


elefeint (approved):
All actual encryption and decryption are handled by the underlying client library (`KeyManagementServiceClient`). So this test is validating what happens if the client library behaves a particular way. For example "if client library successfully encrypts/decrypts, the KmsTemplate will pass the good value through". Likewise, "if client library fails, KmsTemplate will have the expected failure".

PR #1012, opened 2022-03-17
Migrating pubsub tests to JUnit5

ddixit14:
Tests Migrated:

1. PubSubAdminTests.java
2. PubSubTemplateTests.java

### DIFF
--- a/spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubTemplateTests.java
+++ b/spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubTemplateTests.java
@@ -46,21 +46,22 @@ import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubTemplateTests.java
```diff
@@ -149,16 +142,19 @@ public void testPublish_Object() throws IOException {
         .when(pubSubPublisherTemplate)
         .publish(eq("test"), any());
 
-    pubSubPublisherTemplate.publish("test", allowedPayload);
+    pubSubPublisherTemplate.publish("test", allowedPayload, null);
```
elefeint, line 145: Did you mean to change the `publish()` signature that's used here?

elefeint (approved):
I think it's good that you've brought mock behavior setup into the methods, making it obvious what setup each test requires, even though it creates more code duplication. In our production code, we want to avoid code duplication, but in unit tests, it's often useful.

PR #1018, opened 2022-03-21
Remove startup pubsub validation

elefeint:
We've had failures reported on startup due to startup subscription checking taking too long. While this can be worked around with a longer timeout, actuator healthchecks are a cross-cutting concern that should not prevent application startup.


zhumin8 (approved):
LGTM. We never mentioned this in our doc in the first place, but do you think it worth to put a note in there that it's up to application to validate health checker?

elefeint:
I don't think so -- we are just bringing the behavior back to what actuator healthchecks normally do. These healthchecks can be exposed to dashboards, rolling up to Kubernetes liveness (although that's a bad idea because Kubernetes will then restart the pod, which is pointless if there is a connectivity issue to Pub/Sub).

PR #1028, opened 2022-03-22
Enforce license

elefeint:
This PR has two checkstyle changes:
1) Enables checkstyle rule for flagging unused imports. We lost this rule in Spring-to-Google style transition, so now Sonar picks up on those -- that's a bit late in the process.
2) Enforces Spring, Google or combined license. Going forward, all new files will have "Google LLC" copyright. All existing files will retain the original Spring header, but when they get modified, they can now have two copyright lines. So, the three following copyright statements are all correct:

```
 * Copyright 2017-2018 the original author or authors.
```

```
 * Copyright 2022-2022 Google LLC
```

```
 * Copyright 2017-2021 the original author or authors.
 * Copyright 2022-2022 Google LLC
```

As part of enforcing licenses, I've added old-style license headers to all `package-info.java` files. See [this discussion](https://github.com/spring-attic/spring-cloud-gcp/pull/1363#discussion_r245776156) for context -- it seems adding license headers to those was the original intent, but we never quite did it.

I'd also like to explain the regex/property mess that I had to add to enforce licenses:

```
^/\*$
^ \* Copyright 20\d\d-20\d\d (Google LLC|the original author or authors.)$  # This allows either a "Google LLC" or "the original authors" verbiage on line 2.
^ \* Copyright 20\d\d-20\d\d Google LLC$  # line 3 seems like it's required, but it's not. See explanation below.
^ \*$
^ \* Licensed under the Apache License, Version 2\.0 \(the "License"\);$
^ \* you may not use this file except in compliance with the License\.$
^ \* You may obtain a copy of the License at$
^ \*$
^ \*\s?     https://www.apache.org/licenses/LICENSE-2.0$ # there is a single space difference between two licenses. "\s?" is an optional extra single space.
```

In custom-checks.xml, `<property name="multiLines" value="3"/>` effectively makes line 3 (the second copyright statement) optional. It's technically "repeatable 0 or more times", so it would allow multiple instances of Google license, but that does not seem harmful.


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
```diff
@@ -27,7 +27,6 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
-import org.springframework.beans.factory.BeanInitializationException;
```
mpeddada1, line 30: Is this an intentional change?

mpeddada1 (approved):
Thanks for the super detailed description! LGTM

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java
mpeddada1: Aha! Nice, that's a good test for point #1 :)

PR #1031, opened 2022-03-23
Migrating datastore tests to JUnit5

ddixit14:
Tests Migrated:

1. DatastoreIntegrationTests.java
2. ParallelDatastoreIntegrationTests.java
3. SubclassesDescendantsIntegrationTests.java
4. SubclassesReferencesIntegrationTests.java
5. PartTreeDatastoreQueryTests.java
6. DatastoreRepositoryFactoryTests.java
7. SimpleDatastoreRepositoryTests.java

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQueryTests.java
+++ b/spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQueryTests.java
@@ -17,6 +17,7 @@
 package com.google.cloud.spring.data.datastore.repository.query;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.isA;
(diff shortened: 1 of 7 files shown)


spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQueryTests.java
```diff
@@ -809,263 +798,252 @@ private void prepareDeleteResults(boolean isCollection) {
...
+  void unSupportedOrTest() {
+
+    assertThatThrownBy(() -> queryWithMockResult("countByTraderIdOrPrice", null, getClass().getMethod("traderAndPrice")))
+            .hasMessage("Cloud Datastore only supports multiple filters combined with AND.");
 
-    queryWithMockResult("countByTraderIdOrPrice", null, getClass().getMethod("traderAndPrice"));
 
     // this.partTreeDatastoreQuery = createQuery();
-    this.partTreeDatastoreQuery.execute(new Object[] {123L, 45L});
+    //this.partTreeDatastoreQuery.execute(new Object[] {123L, 45L});
```
elefeint, line 871: This test was unfortunately structured in the first place with mock setup faling. Could you remove the commented-out lines, and add a comment to this test saying that "PartTreeDatastoreQuery constructor will fail as part of queryWithMockResult setup"?

The commented out line that was there in the first place shows why it was done this way -- getting all the setup in place before calling the constructor was too hard. This is commonly a sign that code under test has a structure that does not lend itself to testing.

elefeint (approved):

elefeint:
The reason checkstyle complains about the file name not matching the top-level public class name is a minor false-positive error with checkstyle. It sees that there is an interface above the no-longer-public classname, and assumes the file name should match the interface name instead.

If you move the interface from being above the class to being below the class, you'll be able to remove the `public` keyword without running into problems with checkstyle.

If this was an important false-positive, we'd file an issue with checkstyle. But putting a repository interface first, above the class that's the main point of this file, was an odd choice anyway, so it's better to fix it in our repo.

PR #1038, opened 2022-03-30
Allow customizing Pub/Sub Spring Cloud Stream header mapping - Draft for Review

ddixit14:
I have included the modified app sample as well, one just need to trigger the function by clicking "post it" on localhost:8080, the message will always be the same.


ddixit14:
The failing tests are (1) Sonatype code coverage - which I need to discuss with you and (2) dead-letter-sample, the application that  I modified to check things. 
Once you approve that non-sample changes are okay, I'll revert back the sample to what it was and then see if we are getting any errors.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -110,6 +110,10 @@ public void setHeaderMapper(HeaderMapper<Map<String, String>> headerMapper) {
     this.headerMapper = headerMapper;
   }
 
+  public HeaderMapper<Map<String, String>> getHeaderMapper() {
+    return this.headerMapper;
+  }
+
```
elefeint, line 116: Aha! I thought it was odd that the `headerMapper` property ended up being mutable!
Generally, returning an object that the caller can then change is a dangerous practice because then this class (`PubSubInboundChannelAdapter`) can't guarantee the integrity of its own state. 
In this case, you are explicitly returning it in order to change it -- creating a great example of why this kind of "getter of a mutable object/collection" pattern is dangerous.

Please remove the getters here and in the `PubSubMessageHandler`, and use the existing setter instead.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-dead-letter-sample/pom.xml
```diff
@@ -63,6 +63,10 @@
             <version>2.11.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-messaging</artifactId>
+        </dependency>
```
elefeint, line 69: This dependency is already brought in transitively -- Spring Cloud Stream is built on Spring Integration.

You can verify this by running the dependency tree in the stream binder module or the sample app:
```
mvn dependency:tree -Dincludes=org.springframework:spring-messaging
```

elefeint (changes requested):
Good job with manual testing -- I see that the sample app modifications would allow these. We won't keep the sample app changes because, like we discussed, the app is already complicated enough.

For unit testing, take a look at what `PubSubMessageChannelBinderTests.producerSyncPropertyPropagatesToMessageHandler()`, `testConsumerEndpointCreation()` and other tests in that class are doing. These are unit tests -- they only make sure that if a property is provided by a user, that the Pub/Sub messaging objects that get created will be aware of them. Note that you might have to add _some_ getter(s) to the inbound/outbound adapter classes, but make sure you don't return mutable objects.

For integration testing, I'd probably try to add a test to the emulator-based `PubSubMessageChannelBinderEmulatorIntegrationTests`.

Try to get unit tests to work, and if you have time -- give a try to the integration test. Let's then pair on the testing tomorrow.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -75,8 +78,15 @@ protected MessageHandler createProducerMessageHandler(
       ExtendedProducerProperties<PubSubProducerProperties> producerProperties,
       MessageChannel errorChannel) {
 
-    PubSubMessageHandler messageHandler =
-        new PubSubMessageHandler(this.pubSubTemplate, destination.getName());
+    PubSubMessageHandler messageHandler = new PubSubMessageHandler(this.pubSubTemplate, destination.getName());
+
+    HeaderMapper<Map<String, String>> headerMapper = new PubSubHeaderMapper();
```
elefeint, line 83: Use ` PubSubHeaderMapper` as a handle to avoid casting below.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -75,8 +78,15 @@ protected MessageHandler createProducerMessageHandler(
...
+
+    if (producerProperties.getExtension().getAllowedHeaders() != null) {
```
elefeint, line 85: Consider extracting `producerProperties.getExtension().getAllowedHeaders()` into a local variable; also check if it can be null.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -76,6 +78,12 @@ public void setAckMode(AckMode ackMode) {
     this.ackMode = ackMode;
   }
 
+  public String[] getAllowedHeaders() {return this.allowedHeaders;}
+
+  public void setAllowedHeaders(String[] allowedHeaders) {
+    this.allowedHeaders = allowedHeaders;
+  }
```
elefeint, line 85: Instead of those, use `FieldUtils.readField(..., ..., true)`

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -75,8 +78,15 @@ protected MessageHandler createProducerMessageHandler(
...
+      ((PubSubHeaderMapper) headerMapper).setOutboundHeaderPatterns(producerProperties.getExtension().getAllowedHeaders());
+      messageHandler.setAllowedHeaders(producerProperties.getExtension().getAllowedHeaders());
+    }
+    messageHandler.setHeaderMapper(headerMapper);
```
elefeint, line 89: Only change the `messageHandler`'s mapper if the users passed the property (move this into the if statement)

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -75,8 +78,15 @@ protected MessageHandler createProducerMessageHandler(
...
 
-    PubSubMessageHandler messageHandler =
-        new PubSubMessageHandler(this.pubSubTemplate, destination.getName());
+    PubSubMessageHandler messageHandler = new PubSubMessageHandler(this.pubSubTemplate, destination.getName());
+
+    HeaderMapper<Map<String, String>> headerMapper = new PubSubHeaderMapper();
+
+    if (producerProperties.getExtension().getAllowedHeaders() != null) {
+      ((PubSubHeaderMapper) headerMapper).setOutboundHeaderPatterns(producerProperties.getExtension().getAllowedHeaders());
+      messageHandler.setAllowedHeaders(producerProperties.getExtension().getAllowedHeaders());
```
elefeint, line 87: This will go away when you remove the allowedHeaders that's used for testing.

elefeint (changes requested):
from in-person

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -75,8 +76,15 @@ protected MessageHandler createProducerMessageHandler(
       ExtendedProducerProperties<PubSubProducerProperties> producerProperties,
       MessageChannel errorChannel) {
 
-    PubSubMessageHandler messageHandler =
-        new PubSubMessageHandler(this.pubSubTemplate, destination.getName());
+    PubSubMessageHandler messageHandler = new PubSubMessageHandler(this.pubSubTemplate, destination.getName());
+
+    String [] headers = producerProperties.getExtension().getAllowedHeaders();
```
elefeint, line 81: The next line checks that `producerProperties.getExtension()` is not null, but if it is null, then there will be a NullPointerException thrown on this line.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -91,6 +99,13 @@ protected MessageProducer createConsumerEndpoint(
     PubSubInboundChannelAdapter adapter =
         new PubSubInboundChannelAdapter(this.pubSubTemplate, destination.getName());
 
+    String [] headers =  properties.getExtension().getAllowedHeaders();
+    if (properties.getExtension() != null && headers != null) {
```
elefeint, line 103: Same, NPE on previous line.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -91,6 +99,13 @@ protected MessageProducer createConsumerEndpoint(
     PubSubInboundChannelAdapter adapter =
         new PubSubInboundChannelAdapter(this.pubSubTemplate, destination.getName());
 
+    String [] headers =  properties.getExtension().getAllowedHeaders();
```
elefeint, line 102: nit: `String[]`

elefeint (changes requested):
Almost there for the actual functionality-driving code! See inline comments.
Also:
* add two more unit tests to verify that if the user did *not* provide any special properties, the inbound/outbound headers remain the default ones.
* Try to add an integration test with the emulator to this PR; it will replace the sample app modifications.
* Add reference documentation in `docs/.../spring-stream.adoc`

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinderEmulatorIntegrationTests.java
```diff
@@ -74,9 +85,116 @@ public void testClean(TestInfo testInfo) throws Exception {
...
+        latch.countDown();
+      }
+    });
+
+    moduleOutputChannel.send(message);
+    Assert.isTrue(latch.await(5, TimeUnit.SECONDS), "Failed to receive message");
+
+    assertThat(inboundMessageRef.get().getPayload()).isEqualTo("insert some random stuff here".getBytes());
+    assertThat(inboundMessageRef.get().getHeaders().get("secondHeader")).isNull();
+    assertThat(inboundMessageRef.get().getHeaders().get("firstHeader")).hasToString("firstHeaderValue");
```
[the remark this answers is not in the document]
ddixit14: I initially did .equals() only but then sonar-cloud suggested these changes on my previous commit.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubMessageChannelBinderEmulatorIntegrationTests.java
```diff
@@ -74,9 +85,116 @@ public void testClean(TestInfo testInfo) throws Exception {
   }
 
   @Test
-  @Disabled("Looks like there is no Kryo support in SCSt")
+  //@Disabled("Looks like there is no Kryo support in SCSt")
```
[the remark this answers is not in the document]
ddixit14: Yes, my bad! Will clean this up.

PR #1057, opened 2022-04-07
Update spring cloud functions versions

elefeint:
The next Spring Cloud train release is next week. Update spring-cloud-functions versions in the meantime.

Also:
* update Spring Cloud BOM manually; dependabot missed the last one because we applied second-to-last manually.
* Remove the unnecessary `spring-cloud-dependencies-parent`; this was only needed for when the code was under Spring Cloud org.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,15 +80,6 @@
 	<dependencyManagement>
 		<dependencies>
 
-			<!-- Addresses https://tanzu.vmware.com/security/cve-2022-22963 -->
-			<dependency>
-				<groupId>org.springframework.cloud</groupId>
-				<artifactId>spring-cloud-function-dependencies</artifactId>
-				<version>3.2.3</version>
-				<type>pom</type>
-				<scope>import</scope>
-			</dependency>
-
 			<dependency>
 				<groupId>com.google.cloud</groupId>
 				<artifactId>spring-cloud-gcp-dependencies</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,7 @@
 	<dependencyManagement>
 		<dependencies>
 
+			<!-- Addresses https://tanzu.vmware.com/security/cve-2022-22963 -->
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-function-dependencies</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,15 @@
 
 	<dependencyManagement>
 		<dependencies>
+
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-function-dependencies</artifactId>
+				<version>3.2.3</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
+
 			<dependency>
 				<groupId>com.google.cloud</groupId>
 				<artifactId>spring-cloud-gcp-dependencies</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,6 @@
 
 	<dependencyManagement>
 		<dependencies>
-
 			<dependency>
 				<groupId>com.google.cloud</groupId>
 				<artifactId>spring-cloud-gcp-dependencies</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,7 @@
 	<properties>
 
 		<!-- Dependency versions -->
-		<spring-cloud-dependencies.version>2021.0.0</spring-cloud-dependencies.version>
+		<spring-cloud-dependencies.version>2021.0.1</spring-cloud-dependencies.version>
 		<spring-boot-dependencies.version>2.6.6</spring-boot-dependencies.version>
 		<zipkin-gcp.version>1.0.4</zipkin-gcp.version>
 		<java-cfenv.version>2.4.0</java-cfenv.version>
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -31,6 +31,15 @@
 	<dependencyManagement>
 		<dependencies>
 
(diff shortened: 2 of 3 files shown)


spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -2,12 +2,6 @@
 <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
 		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
-	<parent>
```
meltsufin, line 5: Why this change?

pom.xml
```diff
@@ -79,6 +79,15 @@
 
 	<dependencyManagement>
 		<dependencies>
+
+			<dependency>
```
meltsufin, line 83: Can we add a comment as to why this is there? Also, if the application imports both, Spring Cloud GCP BOM and Spring Cloud BOM, would this still work? I think the version specified in Spring Cloud BOM will take precedence if the BOM is imported first.

spring-cloud-gcp-dependencies/pom.xml
elefeint: It confusingly brings in a completely different version that may or may not match what's in the 2021 BOM. I can back the change out, but then we should update the version. `spring-cloud-build` version in 2021.0.1 BOM is 3.1.1, and we here are still on 3.0.4.

It's also irrelevant because [`spring-cloud-dependencies-parent`](https://github.com/spring-cloud/spring-cloud-build/blob/v3.0.4/spring-cloud-dependencies-parent/pom.xml) does not contain anything except for license/scm/distribution and the snapsot repositories.

meltsufin: Maybe let's just update the version for now, and do the removal in a different PR if needed?

pom.xml
meltsufin: What about my question about the BOMs order?

elefeint: Missed the question, sorry. In Maven BOM import, the first version encountered wins. That's why I put `spring-cloud-function-dependencies` first. From [maven docs](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#transitive-dependencies): "Note that if two dependency versions are at the same depth in the dependency tree, the first declaration wins."

Do you mean if a customer puts Spring Cloud BOM in their application? The value of Spring Cloud GCP upgrading and releasing is for the customers who use our BOM to manage Spring Boot dependencies. If a customer application puts a different BOM first, there is nothing we can do about it.

I've actually realized that putting the version overrides in the Spring Cloud GCP parent is insufficient -- we have to add it to the `spring-cloud-gcp-dependencies` BOM, too. Otherwise the closest version ends up the one derived from spring-cloud-stream.

elefeint: Yes, I've tested one of the samples, and if the user puts Spring Cloud BOM above Spring Cloud GCP BOM, they will get the old version of functions:

```
    <dependencyManagement>

        <dependencies>

            <dependency>
                <groupId>org.springframework.cloud</groupId>
                <artifactId>spring-cloud-dependencies</artifactId>
                <version>2021.0.1</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>

            <dependency>
                <groupId>com.google.cloud</groupId>
                <artifactId>spring-cloud-gcp-dependencies</artifactId>
                <version>${project.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>
```
 Running `mvn dependency:tree -Dincludes=org.springframework.cloud:spring-cloud-function-context`:

```
[INFO] --- maven-dependency-plugin:3.2.0:tree (default-cli) @ spring-cloud-gcp-pubsub-stream-sample ---
[INFO] com.google.cloud:spring-cloud-gcp-pubsub-stream-sample:jar:3.3.0-SNAPSHOT
[INFO] \- com.google.cloud:spring-cloud-gcp-pubsub-stream-binder:jar:3.3.0-SNAPSHOT:compile
[INFO]    \- org.springframework.cloud:spring-cloud-stream:jar:3.2.2:compile
[INFO]       \- org.springframework.cloud:spring-cloud-function-context:jar:3.2.2:compile

```

But if you have Spring Cloud GCP BOM first, the behavior is correct:

```
    <dependencyManagement>

        <dependencies>

            <dependency>
                <groupId>com.google.cloud</groupId>
                <artifactId>spring-cloud-gcp-dependencies</artifactId>
                <version>${project.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>

            <dependency>
                <groupId>org.springframework.cloud</groupId>
                <artifactId>spring-cloud-dependencies</artifactId>
                <version>2021.0.1</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>

        </dependencies>
    </dependencyManagement>
```

```
[INFO] --- maven-dependency-plugin:3.2.0:tree (default-cli) @ spring-cloud-gcp-pubsub-stream-sample ---
[INFO] com.google.cloud:spring-cloud-gcp-pubsub-stream-sample:jar:3.3.0-SNAPSHOT
[INFO] \- com.google.cloud:spring-cloud-gcp-pubsub-stream-binder:jar:3.3.0-SNAPSHOT:compile
[INFO]    \- org.springframework.cloud:spring-cloud-stream:jar:3.2.2:compile
[INFO]       \- org.springframework.cloud:spring-cloud-function-context:jar:3.2.3:compile

```

meltsufin: Unfortunately, it's common to have the Spring Cloud BOM first. For instance, the Spring Initializr does that by default. So, I'm not sure if we should add this temp fix as it may give folks a false sense of security.
Instead, it might be better to advise them to add the spring-cloud-function-dependencies directly in their application pom.xml.

meltsufin (approved):

PR #1068, opened 2022-04-12
Open 3.3.0-SNAPSHOT for development

elefeint:
Since the previous release was 3.2.1, we get to open 3.3.0-SNAPSHOT up again.


docs/src/main/asciidoc/migration-guide-1.x.adoc
```diff
@@ -32,17 +32,10 @@ To use the newly unbundled libraries, add the `spring-cloud-gcp-dependencies` bi
 ----
 <dependencyManagement>
   <dependencies>
-    <dependency>
```
elefeint, line 35: Pulling in Spring Cloud BOM is not something we recommend in the main docs, so we should not do it for migration, either.

meltsufin (approved):

PR #1110, opened 2022-05-03
Fully qualified subscription configuration support + immutable configuration in Pub/Sub

### LINKED ISSUE #1005: Wrong mechanism for creating subscription specific FlowControlSettings in GcpPubSubAutoConfiguration
In class `GcpPubSubAutoConfiguration` in method `createAndRegisterSelectiveFlowControlSettings` you are iterating over entries in subscriber map:
`for (Map.Entry<String, PubSubConfiguration.Subscriber> subscription :
        subscriberMap.entrySet())`
The issue here is that within this loop you are also calling:
``` 
this.gcpPubSubProperties.computeSubscriberFlowControlSettings(subscriptionName, this.finalProjectIdProvider.getProjectId());
```
Which underneath executes method `getSubscriber` from class `com.google.cloud.spring.pubsub.core.PubSubConfiguration` which internally modifies entry set over which you are iterating (one mentioned above). Code responsible for that:
```if (this.subscription.containsKey(subscriptionName)
        && projectIdFromFullName.equals(projectId)) {
      this.subscription.putIfAbsent(
          fullyQualifiedSubscriptionKey, this.subscription.get(subscriptionName));
      this.subscription.remove(subscriptionName);
      return this.subscription.get(fullyQualifiedSubscriptionKey);
    }
```
Obviously such a modification of a set should be illegal since there is no guarantee in the order of processing elements. Such a situation leads to errors like `BeanDefinitionOverrideException` - if subscriber is processed twice during iteration.

elefeint:
TLDR: 
- immutable maps
- type-safe and unambiguous `ProjectSubscriptionName` map keys
- switched map types to generic `Map` interface

Details:
- In the past, settings for subscriptions not explicitly specified by the user were cached in the map as a reference to global settings. Now they are not; instead global subscriber is returned on-demand whenever properties are requested for an heretofore unknown subscription. This caused a couple of tests to change in behavior -- instead of testing for 2 entries in the map, they are now testing for only 1.
- An instance of `PubSubConfiguration` effectively can't be used until `.initialize()` is called, allowing tying settings to fully qualified subscription names, using the passed-in project ID as a default.
-  `PubSubConfiguration` no longer allows retrieval of user-provided properties in the `subscription` map. Instead, an immutable map containing equivalent fully-qualified subscription keys can be retrieved with `.getFullyQualifiedSubscriberProperties()`. Both Spring bean wiring and our unit tests now use the `setSubscription()` setter to inject "user provided" properties.
- NEW FEATURE: users can now configure the fully-qualified subscription name with `spring.cloud.gcp.pubsub.subscription.fake-short-name.fully-qualified-name` property.

What did not change?
- I did not update many of the `compute*` methods to accept a fully qualified `ProjectSubscriptionName` because they are called from a place where only a string subscription name (either short or fully qualified) is available.
- Almost no new tests; the old tests covered all cases. 

Fixes #1005.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -55,10 +55,10 @@ import com.google.cloud.spring.pubsub.support.converter.PubSubMessageConverter;
 import com.google.pubsub.v1.ProjectSubscriptionName;
 import java.io.IOException;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
@@ -98,17 +98,17 @@ public class GcpPubSubAutoConfiguration {
 
   private final HeaderProvider headerProvider = new UserAgentHeaderProvider(this.getClass());
 
-  private final ConcurrentHashMap<String, ThreadPoolTaskScheduler> threadPoolTaskSchedulerMap =
-      new ConcurrentHashMap<>();
+  private final Map<ProjectSubscriptionName, ThreadPoolTaskScheduler> threadPoolTaskSchedulerMap =
+      new HashMap<>();
 
-  private final ConcurrentHashMap<String, FlowControlSettings> subscriberFlowControlSettingsMap =
-      new ConcurrentHashMap<>();
+  private final Map<ProjectSubscriptionName, FlowControlSettings> subscriberFlowControlSettingsMap =
+      new HashMap<>();
 
-  private final ConcurrentHashMap<String, RetrySettings> subscriberRetrySettingsMap =
-      new ConcurrentHashMap<>();
+  private final Map<ProjectSubscriptionName, RetrySettings> subscriberRetrySettingsMap =
+      new HashMap<>();
 
-  private final ConcurrentHashMap<String, ExecutorProvider> executorProviderMap =
-      new ConcurrentHashMap<>();
+  private final Map<ProjectSubscriptionName, ExecutorProvider> executorProviderMap =
+      new HashMap<>();
 
   private final ApplicationContext applicationContext;
 
@@ -511,21 +511,18 @@ public class GcpPubSubAutoConfiguration {
    * @param context application context
    */
   private void registerSelectiveSchedulerBeans(GenericApplicationContext context) {
-    Map<ProjectSubscriptionName, PubSubConfiguration.Subscriber> propertyMap =
+    Map<ProjectSubscriptionName, PubSubConfiguration.Subscriber> subscriberMap =
         this.gcpPubSubProperties.getFullyQualifiedSubscriberProperties();
-    // TODO halfway done, try to use fq where possiblel
-    for (ProjectSubscriptionName fullSubscriptionName : propertyMap.keySet()) {
-      //String subscriptionName = fullSubscriptionName.getSubscription();
-      PubSubConfiguration.Subscriber subscriberProps = propertyMap.get(fullSubscriptionName);
-      Integer selectiveExecutorThreads = subscriberProps.getExecutorThreads();
+    for (ProjectSubscriptionName psn : subscriberMap.keySet()) {
+      PubSubConfiguration.Subscriber selectiveSubscriber = subscriberMap.get(psn);
+      Integer selectiveExecutorThreads = selectiveSubscriber.getExecutorThreads();
       if (selectiveExecutorThreads != null) {
-        String suffix = fullSubscriptionName.toString(); // will include slashes
-        String threadName = "gcp-pubsub-subscriber-" + suffix;
-        String beanName = "threadPoolScheduler_" + suffix;
+        String qualifiedName = psn.toString(); // will include slashes
+        String threadName = "gcp-pubsub-subscriber-" + qualifiedName;
+        String beanName = "threadPoolScheduler_" + qualifiedName;
         ThreadPoolTaskScheduler selectiveScheduler =
             createAndRegisterSchedulerBean(selectiveExecutorThreads, threadName, beanName, context);
-        this.threadPoolTaskSchedulerMap.putIfAbsent(
-            fullSubscriptionName.toString(), selectiveScheduler);
+        this.threadPoolTaskSchedulerMap.putIfAbsent(psn, selectiveScheduler);
       }
     }
   }
@@ -569,20 +566,19 @@ public class GcpPubSubAutoConfiguration {
   }
 
   private void createAndRegisterSelectiveFlowControlSettings(GenericApplicationContext context) {
-    Map<ProjectSubscriptionName, PubSubConfiguration.Subscriber> propertyMap =
+    Map<ProjectSubscriptionName, PubSubConfiguration.Subscriber> subscriberMap =
         this.gcpPubSubProperties.getFullyQualifiedSubscriberProperties();
 
-    for (ProjectSubscriptionName fullSubscriptionName : propertyMap.keySet()) {
-      String fullSubscriptionNameString = fullSubscriptionName.toString();
+    for (ProjectSubscriptionName psn : subscriberMap.keySet()) {
+      String qualifiedName = psn.toString();
       PubSubConfiguration.FlowControl flowControl =
-          this.gcpPubSubProperties.computeSubscriberFlowControlSettings(
-              fullSubscriptionNameString, this.finalProjectIdProvider.getProjectId());
+          this.gcpPubSubProperties.computeSubscriberFlowControlSettings(psn);
       FlowControlSettings flowControlSettings = buildFlowControlSettings(flowControl);
       if (flowControlSettings != null
           && !flowControlSettings.equals(this.globalFlowControlSettings)) {
         this.subscriberFlowControlSettingsMap.putIfAbsent(
-            fullSubscriptionNameString, flowControlSettings);
-        String beanName = "subscriberFlowControlSettings-" + fullSubscriptionNameString;
+            psn, flowControlSettings);
+        String beanName = "subscriberFlowControlSettings-" + qualifiedName;
         context.registerBeanDefinition(
             beanName,
             BeanDefinitionBuilder.genericBeanDefinition(
@@ -593,18 +589,16 @@ public class GcpPubSubAutoConfiguration {
   }
 
   private void createAndRegisterSelectiveExecutorProvider(GenericApplicationContext context) {
-    for (Map.Entry<String, ThreadPoolTaskScheduler> schedulerSet :
-        this.threadPoolTaskSchedulerMap.entrySet()) {
-      ProjectSubscriptionName fullSubscriptionName = getFullSubscriptionName(schedulerSet.getKey());
-      String fullyQualifiedName = fullSubscriptionName.toString();
-      if (!this.executorProviderMap.containsKey(fullyQualifiedName)) {
-        ThreadPoolTaskScheduler scheduler = schedulerSet.getValue();
+    for (ProjectSubscriptionName psn : this.threadPoolTaskSchedulerMap.keySet()) {
+      String fullyQualifiedName = psn.toString();
+      if (!this.executorProviderMap.containsKey(psn)) {
+        ThreadPoolTaskScheduler scheduler = this.threadPoolTaskSchedulerMap.get(psn);
         ExecutorProvider executorProvider =
             createAndRegisterExecutorProvider(
                 "subscriberExecutorProvider-" + fullyQualifiedName,
                 scheduler,
                 context);
-        this.executorProviderMap.putIfAbsent(fullyQualifiedName, executorProvider);
+        this.executorProviderMap.putIfAbsent(psn, executorProvider);
       }
     }
   }
@@ -621,19 +615,19 @@ public class GcpPubSubAutoConfiguration {
   }
 
   private void createAndRegisterSelectiveRetrySettings(GenericApplicationContext context) {
-    Map<ProjectSubscriptionName, PubSubConfiguration.Subscriber> propertyMap =
+    Map<ProjectSubscriptionName, PubSubConfiguration.Subscriber> subscriberMap =
         this.gcpPubSubProperties.getFullyQualifiedSubscriberProperties();
 
-    for (ProjectSubscriptionName fullSubscriptionName : propertyMap.keySet()) {
-      String fullSubscriptionNameString = fullSubscriptionName.toString();
+    for (ProjectSubscriptionName psn : subscriberMap.keySet()) {
+      String qualifiedName = psn.toString();
 
       PubSubConfiguration.Retry retry =
           this.gcpPubSubProperties.computeSubscriberRetrySettings(
-              fullSubscriptionNameString, this.finalProjectIdProvider.getProjectId());
+              qualifiedName, this.finalProjectIdProvider.getProjectId());
       RetrySettings retrySettings = buildRetrySettings(retry);
       if (retrySettings != null && !retrySettings.equals(this.globalRetrySettings)) {
-        this.subscriberRetrySettingsMap.putIfAbsent(fullSubscriptionNameString, retrySettings);
-        String beanName = "subscriberRetrySettings-" + fullSubscriptionNameString;
+        this.subscriberRetrySettingsMap.putIfAbsent(psn, retrySettings);
+        String beanName = "subscriberRetrySettings-" + qualifiedName;
         context.registerBeanDefinition(
             beanName,
             BeanDefinitionBuilder.genericBeanDefinition(RetrySettings.class, () -> retrySettings)
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -596,12 +596,12 @@ public class GcpPubSubAutoConfiguration {
     for (Map.Entry<ProjectSubscriptionName, ThreadPoolTaskScheduler> schedulerSet :
         this.threadPoolTaskSchedulerMap.entrySet()) {
       ProjectSubscriptionName fullSubscriptionName = schedulerSet.getKey();
-      String fullyQualifiedName = fullSubscriptionName.toString();
+      String qualifiedName = fullSubscriptionName.toString();
       if (!this.executorProviderMap.containsKey(fullSubscriptionName)) {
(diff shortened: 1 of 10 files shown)


spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -92,11 +147,27 @@ public Subscriber getSubscriber(String name, String projectId) {
...
+
+  /**
+   * Computes flow control settings to use. The subscription-specific property takes precedence if
+   * both global and subscription-specific properties are set. If subscription-specific settings are
+   * not set then global settings are picked.
+   *
+   * @param psn Fully qualified subscription name
+   * @return flow control settings defaulting to global where not provided
+   */
+  public FlowControl computeSubscriberFlowControlSettings(ProjectSubscriptionName psn) {
```
mpeddada1, line 166: a minor suggestion here: 
```suggestion
  public FlowControl computeSubscriberFlowControlSettings(ProjectSubscriptionName projectSubscriptionName) {
```

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -444,10 +454,10 @@ public ExecutorProvider getExecutorProvider(String subscriptionName) {
     if (this.executorProvider != null) {
       return this.executorProvider;
     }
-    String fullyQualifiedName =
-        PubSubSubscriptionUtils.toProjectSubscriptionName(subscriptionName, projectId).toString();
-    if (this.executorProviderMap.containsKey(fullyQualifiedName)) {
-      return this.executorProviderMap.get(fullyQualifiedName);
+    ProjectSubscriptionName psn =
```
mpeddada1, line 457:
```suggestion
    ProjectSubscriptionName projectSubscriptionName =
```

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubConfigurationTests.java
```diff
@@ -122,18 +135,22 @@ void testSubscriberFlowControlSettings() {
...
 
     selectiveFlowControl.setLimitExceededBehavior(FlowController.LimitExceededBehavior.Ignore);
     selectiveFlowControl.setMaxOutstandingElementCount(1L);
     selectiveFlowControl.setMaxOutstandingRequestBytes(2L);
 
-    pubSubConfiguration
-        .getSubscription()
-        .put("projects/projectId/subscriptions/subscription-name", subscriber);
+    ProjectSubscriptionName psn =
+        ProjectSubscriptionName.parse("projects/projectId/subscriptions/subscription-name");
```
mpeddada1, line 147: How about we put this string in a variable?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -54,34 +69,74 @@ public Health getHealth() {
...
+  public Map<ProjectSubscriptionName, Subscriber> getFullyQualifiedSubscriberProperties() {
+    Assert.notNull(this.fullyQualifiedSubscriptionProperties, "Please call initialize() prior to retrieving properties.");
+    return this.fullyQualifiedSubscriptionProperties;
+  }
 
-    if (this.subscription.containsKey(fullyQualifiedSubscriptionKey)) {
-      return this.subscription.get(fullyQualifiedSubscriptionKey);
-    }
+  /**
+   * Standardizes all subscription properties to be fully qualified. Not thread-safe.
```
mpeddada1, line 91: It could be helpful to mention in the javadoc that the `fullyQualifiedName` property takes precedence over the generated one.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -20,18 +20,33 @@
...
-  private final ConcurrentHashMap<String, Subscriber> subscription = new ConcurrentHashMap<>();
+  /** Automatically extracted user-provided properties. Contains only short subscription keys
+   *  user-provided properties, therefore do not use except in initialize().
+   */
+  private Map<String, Subscriber> subscription = new HashMap<>();
+
+  /** Properties keyed by fully qualified subscription name.
+   * Initialized once; effectively a singleton.
+   */
+  private Map<ProjectSubscriptionName, Subscriber> fullyQualifiedSubscriptionProperties;
```
mpeddada1, line 49: Maybe we can consider using an `ImmutableMap` here?

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -54,34 +69,74 @@ public Health getHealth() {
...
+   * @deprecated use {@link #getSubscriptionProperties(ProjectSubscriptionName)} instead.
+   */
+  @Deprecated
+  public Subscriber getSubscriber(String name, String projectId) {
+    return getSubscriptionProperties(PubSubSubscriptionUtils.toProjectSubscriptionName(name, projectId));
+  }
 
-    String subscriptionName = fullyQualifiedName.getSubscription();
-    String projectIdFromFullName = fullyQualifiedName.getProject();
+  public Subscriber getSubscriptionProperties(ProjectSubscriptionName psn) {
```
mpeddada1, line 132: Nice! This really simplifies all the logic. Perhaps we could add a javadoc here?

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubConfigurationTests.java
```diff
@@ -349,31 +386,29 @@ void testComputeRetryableCodes_returnCustom() {
...
 
     assertThat(pubSubConfiguration.computeRetryableCodes("subscription-name", "projectId"))
         .containsExactly(Code.INTERNAL);
   }
 
+  // ATTENTION: this test shows a change in behavior. In the past, specific subscription properties
+  // would get derived from global properties if the subscription itself had none defined by user.
+  // Now the fully-qualified subscription map only contains user-specified properties.
   @Test
   void testSubscriberMapProperties_defaultOrGlobal_addToMap() {
```
mpeddada1, line 401: Let's rename the test to represent this change.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubAutoConfigurationIntegrationTests.java
```diff
@@ -53,15 +54,16 @@ class PubSubAutoConfigurationIntegrationTests {
...
-              "spring.cloud.gcp.pubsub.subscription.test-sub-1.retry.total-timeout-seconds=600",
-              "spring.cloud.gcp.pubsub.subscription.test-sub-1.retry.initial-retry-delay-seconds=100",
-              "spring.cloud.gcp.pubsub.subscription.test-sub-1.retry.retry-delay-multiplier=1.3",
-              "spring.cloud.gcp.pubsub.subscription.test-sub-1.retry.max-retry-delay-seconds=600",
-              "spring.cloud.gcp.pubsub.subscription.test-sub-1.retry.max-attempts=1",
-              "spring.cloud.gcp.pubsub.subscription.test-sub-1.retry.initial-rpc-timeout-seconds=600",
-              "spring.cloud.gcp.pubsub.subscription.test-sub-1.retry.rpc-timeout-multiplier=1",
-              "spring.cloud.gcp.pubsub.subscription.test-sub-1.retry.max-rpc-timeout-seconds=600",
+              "spring.cloud.gcp.pubsub.subscription.fully-qualified-test-sub-1-with-project-abc.executor-threads=3",
+              "spring.cloud.gcp.pubsub.subscription.fully-qualified-test-sub-1-with-project-abc.fully-qualified-name=projects/" + projectIdProvider.getProjectId() + "/subscriptions/test-sub-1",
```
mpeddada1, line 58: Perhaps we could put `"projects/ + projectIdProvider.getProjectId() + ..."` in a constant?

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubConfigurationTests.java
```diff
@@ -349,31 +386,29 @@ void testComputeRetryableCodes_returnCustom() {
...
-        .put("projects/projectId/subscriptions/subscription-name", subscriber);
+    pubSubConfiguration.setSubscription(
+        Collections.singletonMap("projects/projectId/subscriptions/subscription-name", subscriber));
+    pubSubConfiguration.initialize("projectId");
 
     assertThat(pubSubConfiguration.computeRetryableCodes("subscription-name", "projectId"))
         .containsExactly(Code.INTERNAL);
   }
 
+  // ATTENTION: this test shows a change in behavior. In the past, specific subscription properties
```
mpeddada1, line 397: To verify, I'm guessing that this change in behavior is in terms of implementation details? The actual behavior of how the properties are picked up still remains the same (i.e. user provided properties come first and then global). Is that an accurate understanding?

mpeddada1 (commented):
Looks great! This is a really nice fix.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -592,18 +593,18 @@ private void createAndRegisterSelectiveFlowControlSettings(GenericApplicationCon
   }
 
   private void createAndRegisterSelectiveExecutorProvider(GenericApplicationContext context) {
-    for (Map.Entry<String, ThreadPoolTaskScheduler> schedulerSet :
+    for (Map.Entry<ProjectSubscriptionName, ThreadPoolTaskScheduler> schedulerSet :
         this.threadPoolTaskSchedulerMap.entrySet()) {
-      ProjectSubscriptionName fullSubscriptionName = getFullSubscriptionName(schedulerSet.getKey());
+      ProjectSubscriptionName fullSubscriptionName = schedulerSet.getKey();
       String fullyQualifiedName = fullSubscriptionName.toString();
```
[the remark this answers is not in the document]
elefeint: yep. I was trying to keep changes to minimum to make review easier, but happy to change for consistency.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -509,21 +510,20 @@ private void registerSubscriberRetrySettingsBeans(GenericApplicationContext cont
...
-      String subscriptionName = fullSubscriptionName.getSubscription();
-      PubSubConfiguration.Subscriber selectiveSubscriber = subscriberMap.get(subscriptionName);
+      ProjectSubscriptionName fullSubscriptionName = subscription.getKey();
+      PubSubConfiguration.Subscriber selectiveSubscriber = subscription.getValue();
       Integer selectiveExecutorThreads = selectiveSubscriber.getExecutorThreads();
       if (selectiveExecutorThreads != null) {
-        String threadName = "gcp-pubsub-subscriber-" + subscriptionName;
-        String beanName = "threadPoolScheduler_" + subscriptionName;
+        String qualifiedName = fullSubscriptionName.toString(); // will include slashes
+        String threadName = "gcp-pubsub-subscriber-" + qualifiedName;
```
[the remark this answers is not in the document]
elefeint: Empirically, it does not. The name is passed around as a String key in many Java maps. There are naming conventions (camelcase), but nothing explicitly preventing non-compliant names.

When I was making this change, it did feel iffy, so I went to look for restrictions on what "name" could contain. [The `Bean` javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html#name--) does not impose restrictions. The [original XML schema](http://www.springframework.org/schema/beans/spring-beans.xsd) (which was the most common way to define bean relationships prior to Spring Boot) specified the type as [`xsd:string`](https://www.w3.org/TR/xmlschema11-2/#string), which is a sequence of characters, which in XML is [basically any Unicode character](https://www.w3.org/TR/xml11/#NT-Char).

If this ends up causing problems, we can set up a naming strategy that either converts slashes into camelcase and/or ads a numeric suffix. 

Another consideration is that having suffixes that long makes using them as thread prefixes basically useless in logs. But still useful for debugging.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -54,34 +69,74 @@ public Health getHealth() {
...
+        "Pub/Sub properties have already been initialized; cannot update subscription properties");
+
+    this.subscription = map;
   }
 
-  public Subscriber getSubscriber(String name, String projectId) {
-    ProjectSubscriptionName fullyQualifiedName =
-        PubSubSubscriptionUtils.toProjectSubscriptionName(name, projectId);
-    String fullyQualifiedSubscriptionKey = fullyQualifiedName.toString();
+  public Map<ProjectSubscriptionName, Subscriber> getFullyQualifiedSubscriberProperties() {
```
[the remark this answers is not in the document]
elefeint: yep

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -54,34 +69,74 @@ public Health getHealth() {
...
+    if (this.fullyQualifiedSubscriptionProperties != null) {
+      logger.warn("Pub/Sub configuration can only be initialized once; ignoring request.");
+      return;
+    }
+
+    Map<ProjectSubscriptionName, Subscriber> fullyQualifiedProps = new HashMap<>();
+    for (Entry<String, Subscriber> entry : this.subscription.entrySet()) {
+      // Subscription name is either a valid short name, or a made-up name with fully-qualified provided as a property
+      Subscriber subProps = entry.getValue();
+      String qualifiedName = subProps.fullyQualifiedName == null ? entry.getKey() : subProps.fullyQualifiedName;
```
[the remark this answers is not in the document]
elefeint: Will do. Bit of a tradeoff between double negatives vs unexpected order, but probably worth it.

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubConfigurationTests.java
elefeint: extracting as global constant `QUALIFIED_SUBSCRIPTION_NAME`

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
elefeint: The implementation is of an immutable map -- it's instantiated in `initialize()`, and never assigned anywhere else. What kind of immutable map is more of an implementation detail -- it could have been a Guava ImmutableMap. In this case, I went with built-in Java `UnmodifiableMap`.

I added a mention that the map is immutable to the `getFullyQualifiedSubscriberProperties()` javadoc.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/it/PubSubAutoConfigurationIntegrationTests.java
elefeint: yes, extracted `fullSubscriptionNameSub1` and `fullSubscriptionNameSub2`.

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubConfigurationTests.java
elefeint: Right; the difference is that the default properties used to get inserted into the map under multiple subscription keys (for all subscription where there was no dedicated configuration), but now only user-provided properties for a specific subscriptions cause that subscription to be present in the map. Global settings are still returned for subscriptions lacking custom configuration, but they are returned right away, without being placed in the map first.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
mpeddada1: Thanks for the explanation. That makes sense! Using a UnmodifiableMap is also a valid approach.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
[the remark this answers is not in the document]
mpeddada1: Thanks for researching this! Sg.

> If this ends up causing problems, we can set up a naming strategy that either converts slashes into camelcase and/or ads a numeric suffix.

Yup, this is a good plan.

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/PubSubConfigurationTests.java
mpeddada1: Nice!

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java
```diff
@@ -54,34 +69,74 @@ public Health getHealth() {
...
+  public void initialize(String defaultProjectId) {
+    if (this.fullyQualifiedSubscriptionProperties != null) {
+      logger.warn("Pub/Sub configuration can only be initialized once; ignoring request.");
+      return;
+    }
+
+    Map<ProjectSubscriptionName, Subscriber> fullyQualifiedProps = new HashMap<>();
+    for (Entry<String, Subscriber> entry : this.subscription.entrySet()) {
+      // Subscription name is either a valid short name, or a made-up name with fully-qualified provided as a property
+      Subscriber subProps = entry.getValue();
```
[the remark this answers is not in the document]
mpeddada1: lol!

mpeddada1 (approved):
LGTM!

PR #1120, opened 2022-05-13
Remove 2.0.x release notes from 3.x branch changelog.

zhumin8:
Clean up changelog: 
Move 2.0.x release notes since 2.0.8 to 2.x branch changelog. These releases are after split with 3.x, and should be in 2.x branch.


elefeint (changes requested):
Could you add a link to the 2.x changelog on top? Something like "For 2.x changelog, see here"

zhumin8:
Good idea! just added.

elefeint (approved):

PR #1142, opened 2022-05-25
Backport: Replace PublisherInterface with Publisher

elefeint:
Backport of #900. I had to frankenstein a unit test -- it was upgraded to JUnit 5 in the main branch, but is still on JUnit4 in 2.x.

----- original PR description ---

After the discussion in googleapis/java-pubsub#949, reverting back to using `Publisher` for all API return types. I've ported the trace header injection functionality from `TracingPublisher` into `TracingPublisherFactory`.

The Trace information will now get into a standard `Publisher` through a builder callback `PubsubMessage.Builder.setTransform()`.


elefeint:
After arguing with myself, decided not to backport new features. Spring Boot 2.5 is out of support now, so while this branch needs security fixes and routine version bumps, it should not introduce new functionality.

PR #1144, opened 2022-05-26
Open 3.4.0-SNAPSHOT for development

ddixit14:


elefeint (approved):
And you can see that the Github Pages reference documentation got published automatically, and there is [a section there now](https://googlecloudplatform.github.io/spring-cloud-gcp/3.3.0/reference/html/index.html#header-mapping-2) that wasn't there in version 3.2.1!

PR #1152, opened 2022-05-30
Customize thread name prefixes for subscription specific TaskSchedulers

poznachowski:
Currently thread names for subscription specific Task Schedulers are lengthy as they comprise of 
`gcp-pubsub-subscriber-` + fully qualified subscription name, ex:
`gcp-pubsub-subscriber-projects/fake project/subscriptions/subscription-name1`

This may result in a lengthy log statement, if thread name is a part of it.

This MR allows customization of subscription specific Task Scheduler's thread names.

### DIFF
--- a/docs/src/main/asciidoc/pubsub.adoc
+++ b/docs/src/main/asciidoc/pubsub.adoc
@@ -109,6 +109,9 @@ Maximum number of outstanding bytes to keep in memory before enforcing flow cont
 The behavior when the specified limits are exceeded. | No | Block
 |===
 
+NOTE: By default, subscription-specific threads are named after fully-qualified subscription name, ex: `gcp-pubsub-subscriber-projects/project-id/subscriptions/subscription-name`.
+This can be customized, by registering a `SelectiveSchedulerThreadNameProvider` bean.
+
 ==== GRPC Connection Settings
 
 The Pub/Sub API uses the https://cloud.google.com/pubsub/docs/reference/service_apis_overview#grpc_api[GRPC] protocol to send API requests to the Pub/Sub service.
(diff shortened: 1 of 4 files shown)


elefeint (changes requested):
It's a nice improvement, but unfortunately accepting a bean in `GcpPubSubAutoConfiguration` constructor that is defined in the same autoconfiguration class leads to  a circular reference. You can test this by starting up the [sample app](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/v3.3.0/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/main/java/com/example/PubSubApplication.java). 

I see two ways around this:
1) creating `defaultSelectiveSchedulerThreadNameProvider` in a new autoconfiguration class, 
2) not providing a default bean at all and instead accepting `Optional<SelectiveSchedulerThreadNameProvider> selectiveSchedulerThreadNameProvider` in the constructor.  Then in `registerSelectiveSchedulerBeans`, the default would end up hardcoded as before:
```
        String threadName = selectiveSchedulerThreadNameProvider.isPresent() ? 
            selectiveSchedulerThreadNameProvider.get().getThreadName(fullSubscriptionName) : 
            "gcp-pubsub-subscriber-" + qualifiedName;
```

The second way seems simpler to me, but it's your call.

Also, could you turn on [google style](https://github.com/google/styleguide) and reformat? `mvn validate` will run style checks to make sure it's good.

elefeint (changes requested):
Looks great!
Would you mind adding a line or two to the _Subscription-specific Configurations_ section in the [reference documentation](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/docs/src/main/asciidoc/pubsub.adoc)?

poznachowski:
Sure, reference added. Please check if that concludes the change.

poznachowski:
Hey @elefeint ! I don't know where to ask this question elsewhere - Do you have any ETA for publishing new version of spring-cloud-gcp ?

elefeint:
We'll release in the next 2 weeks. Need to get Spring Boot 2.7 compatibility in.

PR #1157, opened 2022-06-03
feat: add support for ARRAY<JSON> type in Spring Data Spanner

### LINKED ISSUE #651: support for ARRAY<JSON> in spring data spanner
As an enhancement and followup to spring data spanner JSON support - #635, spring data spanner should also support Array of JSON type fields.
* Note: also handle case where List<T> of simple types get annotated with JSON.

zhumin8:
support ARRAY<JSON> spanner type. 
- read: `StructAccessor.java` and `StructPropertyValueProvider.java`
- write: `ConverterAwareMappingSpannerEntityWriter.java`
- sql query when return type is ARRAY<JSON> field.
    - `SqlSpannerQuery.java` and `SpannerPersistentEntityImpl.java`
- test and sample

closes #651.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
@@ -326,7 +326,7 @@ public class ConverterAwareMappingSpannerEntityWriter implements SpannerEntityWr
 
     if (spannerPersistentProperty.getAnnotatedColumnItemType() == Type.Code.JSON) {
       // if column annotated with JSON, convert directly
-      valueBinder.toJsonArray(this.covertIterableJsonToValue(value));
+      valueBinder.toJsonArray(this.convertIterableJsonToValue(value));
       valueSet = true;
     } else if (spannerPersistentProperty.getAnnotatedColumnItemType() != null) {
       // use the annotated column type if possible.
@@ -376,7 +376,7 @@ public class ConverterAwareMappingSpannerEntityWriter implements SpannerEntityWr
     return Value.json(jsonString);
   }
 
-  private Iterable<String> covertIterableJsonToValue(Iterable<Object> value) {
+  private Iterable<String> convertIterableJsonToValue(Iterable<Object> value) {
     List<String> result = new ArrayList<>();
     if (value == null) {
       return null;
--- a/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
@@ -30,10 +30,12 @@ import com.google.cloud.spring.data.spanner.core.mapping.SpannerMappingContext;
 import com.google.cloud.spring.data.spanner.core.mapping.SpannerPersistentEntity;
 import com.google.cloud.spring.data.spanner.core.mapping.SpannerPersistentProperty;
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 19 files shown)


spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -361,14 +368,23 @@ private static <T> boolean attemptSetSingleItemValue(
...
-  private Value covertJsonToValue(Object value) {
+  private Value convertJsonToValue(Object value) {
     if (value == null) {
       return Value.json(null);
     }
     String jsonString = this.spannerMappingContext.getGson().toJson(value);
     return Value.json(jsonString);
   }
 
+  private Iterable<String> covertIterableJsonToValue(Iterable<Object> value) {
```
elefeint, line 379: You fixed the covert conversion typo but not before copying it over :)

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
```diff
@@ -99,6 +100,7 @@ public class StructAccessor {
           .build();
 
   // @formatter:on
+  private static final String EXCEPTION_NARRATIVE_COL_NOT_ARRAY = "Column is not an ARRAY type: ";
```
elefeint, line 103: This is very minor, but we don't have a constant naming convention with `NARRATIVE` in this code base, so might as well skip it and make the constant name shorter.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
```diff
@@ -145,14 +147,43 @@ public Object getSingleValue(int colIndex) {
...
+  private  <T> List<T> getListJsonValue(int colIndex, Class<T> colType) {
+    if (this.struct.getColumnType(colIndex).getCode() != Code.ARRAY) {
+      throw new SpannerDataException(EXCEPTION_NARRATIVE_COL_NOT_ARRAY + colIndex);
+    }
+    List<String> jsonStringList = this.struct.getJsonList(colIndex);
+    List<T> result = new ArrayList<>();
+    jsonStringList.forEach(item ->
+        result.add(gson.fromJson(item, colType)));
+    return result;
+  }
```
elefeint, line 185: You might be able to reuse the same code for the two `getListJsonValue` definitions by passing `struct -> struct.getJsonList(colIndex);` and `struct->struct.getJsonList(colName)` as lambdas.

elefeint (approved):
Looks amazing. I only have a few optional nits.

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
zhumin8: ah, good catch!!

ansh0l (approved):
Looks good to me.

QQ> I couldn't see a test for the behaviour around a malformed array of Json. Would there be any behavioural difference in what was happening prior to this PR (say it used to throw something like an UnimplementedException) vs after this PR (say it throws something like InvalidDataException)?

rajatbhatta (approved):

spring-cloud-gcp-data-spanner/src/main/java/com/google/cloud/spring/data/spanner/core/convert/StructAccessor.java
```diff
@@ -184,7 +215,8 @@ <T> T getSingleJsonValue(String colName, Class<T> colType) {
     return gson.fromJson(jsonString, colType);
   }
 
-  public <T> T getSingleJsonValue(int colIndex, Class<T> colType) {
+  //TODO: change this to private in next major release
```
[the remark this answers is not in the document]
zhumin8: Yes, `getJsonValue` is public and `getSingleJsonValue` and `getListJsonValue` does not need to be exposed. In fact, this method is mostly for internal usage. Users don’t need to interact with neither of these methods, as shown in the samples, user only need to add annotation of the dedicated field and use save()/find() as usual. But this method needs to be public because it is used outside of its package.

zhumin8:
@ansh0l To your question:
Before this implementation: there is no specific logic around `ARRAY<JSON>` to throw an `UnimplementedException` , it would fall through logic to process regular ARRAY and fail on conversion because it does not know what to convert to. For example on write, it would fail with `Caused by: java.lang.IllegalArgumentException: Target type to convert to cannot be null`

About malformed array of Json: User is expected to interact with entities (data model), with `ARRAY<JSON>` field as a member of `List<some object class>`. Between write and read, conversion is done by Spring Data registered converters where this value is parsed by Gson to and from json format. This step is where malformed Json may occur, and it is expected to be captured by Gson with exception like `JsonParseException`.

PR #1177, opened 2022-06-27
Remove exclamation points

elefeint:
To fix the build. Backport of #1176.


zhumin8 (commented):
This actually breaks the integration tests for 2.x. The spring data change is introduced in [2.6.5](https://github.com/spring-projects/spring-data-commons/releases/tag/2.6.5) release, and I don't see a 2.5.x release around the same time. 

If we want to backport this in case of a future release, maybe use `.hasMessageContaining("Result must not be null");` instead so it satisfies both formats?

elefeint:
Oh! The 2.x build failed for a random unrelated reason. I did not look, just assumed it was the same.

PR #1185, opened 2022-07-12
Upgrade to Spring Boot 2.7.x

elefeint:
With Spring Boot 2.7,  the old R2DBC drivers for MySQL and Postgres were removed from the BOM ([Spring Boot 2.7 release notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.7-Release-Notes#r2dbc-driver-changes)).

Spring Cloud GCP typically follows Spring Cloud, which supports both Spring Boot 2.6 and 2.7 in the same major version of their release trains. Additionally, we are not planning on releasing a major version, so the main goal is to not break existing users. This means keeping the old versions of drivers in Spring Cloud GCP R2DBC starters and managing the two drivers' version in Spring Cloud GCP BOM (`spring-cloud-gcp-dependencies`).

**Postgres**:
- Changed Maven coordinates from `io.r2dbc` to `org.postgres`. Package named stayed the same, so the drivers are technically mutually compatible. However, the new driver uses a constant introduced in R2DBC SPI 0.9 (`LOCK_WAIT_TIMEOUT`), and so is not compatible with applications still on Spring Boot 2.6.x.
- Using the old driver (`0.8.12.RELEASE`) is safe for both Spring Boot 2.6 and 2.7. 
- Spring Boot 2.7 users who want the new functionality will need to exclude the old driver dependency and include the new one. Documentation included in PR.

**MySQL**
- old driver is safe for both Spring Boot 2.6 and 2.7. 
- MariaDB driver is recommended, however there is work to be done in both the [socket factory](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/990) and [MariaDB driver](https://github.com/mariadb-corporation/mariadb-connector-r2dbc/issues/44) to allow its use.

I've tested 5 combinations:
- MySQL with Spring Boot 2.6
- Postgres with Spring Boot 2.6
- MySQL with Spring Boot 2.7
- Postgres with Spring Boot 2.7 -- as is, using old `0.8.12.RELEASE` driver that's included in the starter.
- Postgres with Spring Boot 2.7 -- with customer app excluding the old driver and including `r2dbc-postgresql-0.9.1.RELEASE instead.

Addresses a part of #1200.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ on GitHub.
 ## 3.4.0-SNAPSHOT
 
 **This release officially introduces Spring Boot 2.7 compatibility.**
-Note that the previous releases of Spring Cloud GCP 3.x are also compatible with Spring Boot 2.7. The one exception is that if you use Cloud SQL with R2DBC, you'd have to manage the driver versions in your own application dependencies.
+Note that the previous releases of Spring Cloud GCP 3.x are also compatible with Spring Boot 2.7. The one exception is that if you use Cloud SQL with R2DBC, you'd have to manage the driver versions in your own application dependencies (see [refdoc](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/docs/src/main/asciidoc/sql.adoc#r2dbc-support) for details).
 
 ### Cloud SQL
   - Add version management for the older MySQL and Postgres R2DBC drivers (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1185).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ on GitHub.
 
 ## 3.4.0-SNAPSHOT
 
+**This release officially introduces Spring Boot 2.7 compatibility.**
+Note that the previous releases of Spring Cloud GCP 3.x are also compatible with Spring Boot 2.7. The one exception is that if you use Cloud SQL with R2DBC, you'd have to manage the driver versions in your own application dependencies.
+
+### Cloud SQL
+  - Add version management for the older MySQL and Postgres R2DBC drivers (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1185).
+
 ### Pub/Sub
 
   - Allow Publishers shutdown gracefully
--- a/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessorTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessorTests.java
@@ -105,7 +105,7 @@ class R2dbcCloudSqlEnvironmentPostProcessorTests {
         .withPropertyValues(
             "spring.cloud.gcp.sql.databaseName=my-database",
             "spring.cloud.gcp.sql.instanceConnectionName=my-project:region:my-instance")
-        .withClassLoader(new FilteredClassLoader("dev.miku.r2dbc.mysql"))
+        .withClassLoader(new FilteredClassLoader("dev.miku.r2dbc.mysql", "org.mariadb.r2dbc"))
         .run(
             context -> {
               assertThat(context.getEnvironment().getProperty("spring.r2dbc.url"))
--- a/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessorTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessorTests.java
@@ -19,9 +19,7 @@ package com.google.cloud.spring.autoconfigure.sql;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 import org.junit.jupiter.api.Test;
 import org.springframework.boot.SpringApplication;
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ on GitHub.
 
 ## 3.4.0-SNAPSHOT
 
+**This release officially introduces Spring Boot 2.7 compatibility.**
+Note that the previous releases of Spring Cloud GCP 3.x are also compatible with Spring Boot 2.7. The one exception is that if you use Cloud SQL with R2DBC, you'd have to manage the driver versions in your own application dependencies.
+
```
mpeddada1, line 21: optional: Do you think it will be helpful to reference the 2.7 part of the `sql.adoc` here?

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessorTests.java
```diff
@@ -100,18 +103,44 @@ void testSetR2dbcProperty_mySql_urlProvidedByUserIgnored() {
...
+
+  /**
+   * Accepts a list of packages to _keep_, and tests that correct properties for the database type
+   * got injected into context.
+   * The input is used to calculate the inverse list of packages to _remove_ that is passed to
+   * {@code FilteredClassLoader}
+   *
+   * @param includePackages a list of driver packages to keep on the classpath
+   */
+  private void validateDatabaseProperties(String[] includePackages, String username, String url) {
```
mpeddada1, line 127: How about we name this to `verifyhatCorrectUrlAndUsernameSet`? Something that tells us what the expected result is.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessorTests.java
```diff
@@ -100,18 +103,44 @@ void testSetR2dbcProperty_mySql_urlProvidedByUserIgnored() {
...
+    validateDatabaseProperties(new String[] {"dev.miku.r2dbc.mysql"},
+        "root",
+        "r2dbc:gcp:mysql://my-project:region:my-instance/my-database");
+  }
+
+  /**
+   * Accepts a list of packages to _keep_, and tests that correct properties for the database type
+   * got injected into context.
+   * The input is used to calculate the inverse list of packages to _remove_ that is passed to
+   * {@code FilteredClassLoader}
```
mpeddada1, line 123: Very helpful documentation! Very much optional but is it possible to move this line to an inline comment inside the method and perhaps add a line in the javadoc about what the method is verifying?

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -274,13 +276,32 @@
...
 			</dependency>
 
+			<!-- Spring Boot 2.7 no longer manages this dependency.
+     			 Replace with org.postgres group ID once Spring Boot 2.6 is out of support.
+ 			-->
+			<dependency>
+				<groupId>io.r2dbc</groupId>
+				<artifactId>r2dbc-postgresql</artifactId>
+				<version>${r2dbc-postgres-driver.version}</version>
+			</dependency>
```
mpeddada1, line 303: Will our tests still continue to verify Spring Boot 2.6?

elefeint: I thought about introducing a new GHA job, like we had with 2.5 support, but 2.6 is [going out of support in a month](https://spring.io/projects/spring-boot#support)...

elefeint: Still might be worth keeping around for a few months. I'll look into it in a separate PR.

elefeint: #1292 to track.

CHANGELOG.md
mpeddada1: Ah definitely didn't think this through. Makes sense!

spring-cloud-gcp-dependencies/pom.xml
mpeddada1: I see! It is worth it for the sake of making sure there aren't any breaking changes, if doesn't take too much time. Most importantly, I think it is also worth testing Spring Boot 2.7 with the new postgresql driver (with the `org.postgres` coordinates) since we document this.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessorTests.java
elefeint: done -- I think I've made the description clearer. Take a look?

mpeddada1: Looks great! Thank you

CHANGELOG.md
elefeint: Decided to made it a reference to source code; seems less confusing. At least that will be available once this PR is merged.

spring-cloud-gcp-dependencies/pom.xml
elefeint: As discussed; it's either 2.6 testing or the exclude/include-new-driver flow. Chose 2.6 integration testing.

mpeddada1 (approved):
LGTM!

breun:
I see [Spring Cloud GCP 3.4.0 has been released](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/releases/tag/v3.4.0) with support for Spring Boot 2.7, which is great news! I guess https://github.com/GoogleCloudPlatform/spring-cloud-gcp#compatibility-with-spring-project-versions should be updated to indicate support for Spring Boot 2.7 is now available in version 3.4.0?

elefeint:
I'll send a pull request; waiting on the last prerequisite.

Issue #231, opened 2021-01-27
Entity returned by FirestoreTemplate.save has id == null if id is auto generated


Pojahn-Fortum:
When saving an entity, you may choose to not to set the ID to let Firestore generate it for you. Similar functionality exists in nearly all databases.

So if you create a new user:

```java
User usr = new User(null, "first_name", "last_name); 
User returnedUsr = firestoreTemplate.save(user).block(); 
```

If we try to print `returnedUsr.getId()`, we will see that it's null. The only way to currently get the ID is to either do a findAll() or query by some other property. 

Looking at the source code of FirestoreTemplate.saveAll, I can see that the ID is present and there is no need to do an extra request to fetch it. Would be very convenient if the returned entity have its `DocumentId` set.

meltsufin:
@Pojahn-Fortum This makes sense. Would you want the same `usr` object returned, or a new object created? Currently the same object is returned, which is more efficient. As always, you are welcome to create a PR for this. Thanks!

PojahnM:
Yes it's efficient but have very little to no use. In a real life application, we would likely never be able to find out which entity we just created because of the absents of the id. 

Returning the same object with the id field populated could be tricky. It may be immutable or something?

Maybe run the `Document` through the registered class mapper? I think that's most consistent.

I have sadly little time to contribute.

meltsufin:
No problem, we'll try to get to it or maybe someone else will pick it up.

abhijeetshuklaoist:
@meltsufin Hey, I would like to work on this. But I am new to community so might need little help.

meltsufin:
Hi @abhijeetshuklaoist, thanks for you interest in helping out. We would love to help you get started. Just let us know what you need. One way to start would be to open a draft PR, and we can guide you on it.

abhijeetshuklaoist:
Thanks a lot @meltsufin 
I have started working on this. I tried to reproduce the issue using test where I removed the id and then made a call to saveAll and in response I could see the id. 
Any idea what I am doing wrong or any hint would be really helpful as I have started OpenSource contribution just a few weeks back. 
Thanks
@Pojahn-Fortum Could you please help me reproduce the issue.  

[0] https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/258/files

PojahnM:
Here is my test:

```kotlin
	@Test
	fun `returned entity is populated with id`() {
[...]

PR #1196, opened 2022-07-24
Validates that issue #231 works as intended

bijukunjummen:
- Validates that #231 works as intended already by adding tests for checking the behavior - if an entity is added without an id, an id is auto-assigned.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
+++ b/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.spring.data.firestore;
 
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doAnswer;
@@ -112,6 +114,37 @@ public class FirestoreTemplateTests {
 
     builder.addWrites(Write.newBuilder().setUpdate(buildDocument("e1", 100L)).build());
     builder.addWrites(Write.newBuilder().setUpdate(buildDocument("e2", 200L)).build());
+
+    verify(this.firestoreStub, times(1)).commit(eq(builder.build()), any());
+  }
+
+  @Test
+  void saveNoId() {
+    mockCommitMethod();
+
+    TestEntity entity1 = new TestEntity(null, 100L);
+    TestEntity entity2 = new TestEntity(null, 200L);
+    assertNull(entity1.getIdField());
+    assertNull(entity2.getIdField());
+
+    StepVerifier.create(
+            this.firestoreTemplate.saveAll(
+                Flux.just(entity1, entity2)))
+        .assertNext(e1 -> assertNotNull(e1.getIdField()))
+        .assertNext(e2 -> assertNotNull(e2.getIdField()))
+        .verifyComplete();
+
+    CommitRequest.Builder builder =
+        CommitRequest.newBuilder().setDatabase("projects/my-project/databases/(default)");
+
+    builder.addWrites(Write.newBuilder()
+        .setUpdate(buildDocument(entity1.getIdField(), 100L))
+        .setCurrentDocument(Precondition.newBuilder().setExists(false).build())
+        .build());
+    builder.addWrites(Write.newBuilder()
+        .setUpdate(buildDocument(entity2.getIdField(), 200L))
+        .setCurrentDocument(Precondition.newBuilder().setExists(false).build())
+        .build());
 
     verify(this.firestoreStub, times(1)).commit(eq(builder.build()), any());
   }
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/main/java/com/example/UserController.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/main/java/com/example/UserController.java
@@ -86,11 +86,15 @@ public class UserController {
                       formData.getFirst("name"),
                       Integer.parseInt(formData.getFirst("age")),
                       createPets(formData.getFirst("pets")));
-              FirestoreReactiveOperations userTemplate = this.firestoreTemplate.withParent(user);
               List<PhoneNumber> phones = getPhones(formData.getFirst("phones"));
-              return userTemplate
-                  .saveAll(Flux.fromIterable(phones))
-                  .then(userRepository.save(user));
+
+              return userRepository.save(user)
+                  .flatMap(savedUser -> {
+                    FirestoreReactiveOperations userTemplate = this.firestoreTemplate.withParent(
+                        savedUser);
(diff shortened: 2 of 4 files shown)


spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
```diff
@@ -245,31 +278,31 @@ void deleteByIdTest() {
 
   private void mockCommitMethod() {
     doAnswer(
-            invocation -> {
```
emmileaf, line 248: Could you help remove the formatting changes introduced in this PR that are outside of the tests edited/added? (For consistency with the rest of the repository, which conforms to [Google Java Style](https://google.github.io/styleguide/javaguide.html))

spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/test/java/com/example/FirestoreSampleApplicationNativeIntegrationTests.java
```diff
@@ -88,48 +106,45 @@ static void beforeAll() {
...
+    testUserClient.saveUser(ALPHA_USER, ALPHA_PHONE_NUMBERS);
+    verifyUserPersisted(ALPHA_USER);
+
+    testUserClient.saveUser(BETA_USER, emptyList());
+    verifyUserPersisted(BETA_USER);
+
+    testUserClient.saveUser(DELTA_USER, emptyList());
+    verifyUserPersisted(DELTA_USER);
+
+    User savedUser = testUserClient.saveUser(NULL_NAME_USER, emptyList());
```
emmileaf, line 124: Would it be possible to just have minimal changes for the `NULL_USER` case and assertion added here in [`FirestoreSampleNativeApplicationIntegrationTests`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/test/java/com/example/FirestoreSampleNativeApplicationIntegrationTests.java) (maybe also in [`FirestoreSampleApplicationIntegrationTests`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/test/java/com/example/FirestoreSampleApplicationIntegrationTests.java) if it’s not too much effort to do), without removing the `MyAppNativeDockerContainer` aspect of this sample test? 

Please let me know if I am missing any context for why the other changes were made here; if this ends up not being a simple fix, then feel free to leave these tests for the samples as-is, since I think validating for this behaviour in `FirestoreTemplateTests` should be sufficient.

emmileaf (changes requested):

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
bijukunjummen: Done @emmileaf , however I am finding that if I use the Google Intellij Style xml available here - https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml, then it tends to reformat the code the way I had originally submitted. Is it possible to get hold of a intellij style that conforms to the style here.

spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/test/java/com/example/FirestoreSampleApplicationNativeIntegrationTests.java
bijukunjummen: Thanks @emmileaf , I could not execute the original `FirestoreSampleApplicationNativeIntegrationTest` in my local machine as the test depends on a `spring-cloud-gcp-data-firestore-sample:test` image which is not accessible to me. I realized that I could simplify the test to depend instead on the codebase available with the spring-cloud-gcp-data-firestore-sample instead of an image packaged with the binary of the codebase. I also added the null check as part of it as then this feature gets integration tested as part of the change - the issue had mentioned the scenario of emulator, so that is one reason to fix this test. Let me know, I can remove this if it still feels not relevant.

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
elefeint: Because we've moved from Spring style to Google style over time, it's best to avoid reformatting a whole file as part of a feature PR. Try selecting a block of code you are working on and formatting just the block instead. Sorry about that!

spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/test/java/com/example/FirestoreSampleApplicationNativeIntegrationTests.java
elefeint: This one is our bad. `FirestoreSampleApplicationNativeIntegrationTests` has not worked in some time; we should have removed it since we'll be rewriting native support across Spring Cloud GCP.

You might notice that this test does not run with `-Dit.firestore=true` -- could you port your changes to `FirestoreSampleApplicationIntegrationTests`, so they are part of our normal CI?

elefeint (changes requested):

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
bijukunjummen: Ah, okay, that makes sense @elefeint. I will ensure that only my changes are made with the Google style.

spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/test/java/com/example/FirestoreSampleApplicationNativeIntegrationTests.java
bijukunjummen: Done @elefeint, I have now reverted the changes to `FirestoreSampleApplicationNativeIntegrationTests` and instead added the test changes to `FirestoreSampleApplicationIntegrationTests`. I will let the CI verify it however as it depends on a ci project to execute the test. I am assuming Test container based tests will not run in the CI env?

elefeint (approved):

spring-cloud-gcp-samples/spring-cloud-gcp-data-firestore-sample/src/test/java/com/example/FirestoreSampleApplicationNativeIntegrationTests.java
emmileaf: Thank you for porting the changes to `FirestoreSampleApplicationIntegrationTests`, and apologies for missing this earlier!

I ran the Firestore integration tests locally and verified that they pass –  though integration tests were not triggered as part of the presubmit checks for this PR, this test will be included as part of Firestore integration tests in our CI workflow.

emmileaf (approved):

PR #1213, opened 2022-08-05
Add refdoc for short vs fully qualified topics/subscriptions

### LINKED ISSUE #581: Publishing to and Subscribing from multiple Pub/Sub projects is extremely difficult

elefeint:
Several documentation changes:
1) Added notes to Spring Integration adapters that subscription/topic names can be fully qualified.
2) Replaced javadoc `<property>` with `[property]` because IntelliJ does not render what it thinks are HTML tags (`<...>`).
3) Replaced "canonical" with "short", because "canonical" reads as ambiguous -- are short names canonical, or are the fully qualified names canonical? It's kind of both -- short names are historically canonical to Spring Cloud GCP, but fully qualified ones are canonical to the client library.

Fixes #581.

### DIFF
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -43,6 +43,8 @@ The `PubSubInboundChannelAdapter` delegates the conversion to the desired payloa
 
 To use the inbound channel adapter, a `PubSubInboundChannelAdapter` must be provided and configured on the user application side.
 
+NOTE: The subscription name could either be a short subscription name within the current project, or the fully-qualified name referring to a subscription in a different project using the `projects/[project_name]/subscriptions/<subscription_name>` format.
+
 [source,java]
 ----
 @Bean
@@ -202,6 +204,8 @@ This can be overridden by using `setBlockOnPull()` method to wait for at least o
 By default, `PubSubMessageSource` pulls from the subscription one message at a time.
 To pull a batch of messages on each request, use the `setMaxFetchSize()` method to set the batch size.
 
+NOTE: The subscription name could either be a short subscription name within the current project, or the fully-qualified name referring to a subscription in a different project using the `projects/[project_name]/subscriptions/<subscription_name>` format.
+
 [source,java]
 ----
 @Bean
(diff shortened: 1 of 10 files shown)


docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -43,6 +43,8 @@ The `PubSubInboundChannelAdapter` delegates the conversion to the desired payloa
 
 To use the inbound channel adapter, a `PubSubInboundChannelAdapter` must be provided and configured on the user application side.
 
+NOTE: The subscription name could either be a short subscription name within the current project, or the fully-qualified name referring to a subscription in a different project using the `projects/[project_name]/subscriptions/[subscription_name]` format.
+
 [source,java]
```
zhumin8, line 48: Do you mind also adding these updates to `src/main/md/spring-integration.md`, so we don't need to regenerate the md files?

elefeint: Done; changed the format a bit.

zhumin8 (approved):

PR #1215, opened 2022-08-10
Add testcontainers reference

eddumelendez:


elefeint (changes requested):
Thanks! Could you also make the same changes to the `adoc` versions of those files? We are still publishing refdoc from `*.adoc` files for now, but already split off `*.md` files in preparation for the new publishing process.

elefeint (approved):

PR #1219, opened 2022-08-11
feat: BigQuery Storage Write API integration

prash-mi:
Added `BigQueryTemplate.writeJsonStream(...)` method with it's supporting classes and implementations for writing JSON File's inputStream to BigQuery Storage using the storage write API.  This uses exactly once delivery semantics which are supported by the write API, this will guarantee that the data doesn't get duplicated.


spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
```diff
@@ -0,0 +1,148 @@
...
+/** Helper class for using BigQuery storage write API in exactly once delivery mode */
+public class BigQueryJsonDataWriter {
+
+  private JsonStreamWriter streamWriter;
+  // Track the number of in-flight requests to wait for all responses before shutting down.
+  private final Phaser inflightRequestCount = new Phaser(1);
+
+  private final Object lock = new Object();
+
+  private final Logger logger = Logger.getLogger(this.getClass().getName());
```
elefeint, line 50: Unllike with client libraries, the Spring modules use SLF4J abstraction for logging (JUL can still be used as one of the implementations). So this would be `org.slf4j.Logger` static field initialized as `LoggerFactory.getLogger(BigQueryJsonDataWriter.class)`.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -57,6 +75,10 @@ public class BigQueryTemplate implements BigQueryOperations {
 
   private Duration jobPollInterval = Duration.ofSeconds(2);
 
+  private static final int JSON_STREAM_WRITER_BATCH_SIZE = 1000; // write records in batches of 1000
```
elefeint, line 78: It would be good to make this configurable by the user -- could you add a property to `GcpBigQueryProperties`, and inject it with a new setter in this class?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -159,16 +181,152 @@ public ListenableFuture<Job> writeDataToTable(
...
+   *
+   * @param tableName name of the table to write to
+   * @param jsonInputStream input stream of the json file to be written
+   * @return {@link ListenableFuture} containing the WriteApiResponse indicating completion of
+   *     operation
    */
+  @Override
+  public ListenableFuture<WriteApiResponse> writeJsonStream(
+      String tableName, InputStream jsonInputStream, Schema schema) {
+    createTable(tableName, schema); // create table if it's not already created
```
elefeint, line 195: This sounds like a relatively lengthy operation -- are there timeout configurations we can expose to the users later?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -159,16 +181,152 @@ public ListenableFuture<Job> writeDataToTable(
...
+   *     operation
+   */
+  @Override
+  public ListenableFuture<WriteApiResponse> writeJsonStream(
+      String tableName, InputStream jsonInputStream) {
+
+    SettableListenableFuture<WriteApiResponse> writeApiFutureResponse =
+        new SettableListenableFuture<>();
+    CompletableFuture<Void> completableFuture =
+        CompletableFuture.runAsync(
```
elefeint, line 227: Non-actionable, just a note for future -- we may want to allow users to inject a custom threadpool here, like we do for Pub/Sub. Although with BigQuery writing, I doubt there is as high of parallelism as with Pub/Sub.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -159,16 +181,152 @@ public ListenableFuture<Job> writeDataToTable(
...
+                writeApiFutureResponse.setException(e);
+                // Restore interrupted state in case of an InterruptedException
+                Thread.currentThread().interrupt();
+              }
+            });
+
+    // register success and failure callback
+    writeApiFutureResponse.addCallback(
+        (res) -> {
+          completableFuture.cancel(true); // interrupt the completableFuture
```
elefeint, line 245: Is this "just in case"? The success callback ideally will happen after the future value is set. Completable future javadoc says that calling `cancel` is a no-op if already completed, but is there an actual scenario in which canceling an already-completed or alread-failed future is needed?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -159,16 +181,152 @@ public ListenableFuture<Job> writeDataToTable(
...
+        });
+
+    return writeApiFutureResponse;
+  }
+
+  private WriteApiResponse getWriteApiResponse(String tableName, InputStream jsonInputStream)
+      throws DescriptorValidationException, IOException, InterruptedException {
+    WriteApiResponse apiResponse = new WriteApiResponse();
+    BigQueryWriteClient client = BigQueryWriteClient.create();
```
elefeint, line 257: This should be done as part of autoconfiguration, similar to [`BigQuery` object](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/e04f755f75c98619c11fde42106cff81e95f1540/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java#L71) for two reasons -- 
1) to allow users to customize their own `BigQueryWriteClient` and provide it as part of user configuration.
2) to allow us to provide properties for customization that will automatically be used to configure the object in autoconfiguration.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
```diff
@@ -0,0 +1,148 @@
...
+
+  @GuardedBy("lock")
+  private RuntimeException error = null;
+
+  void initialize(TableName parentTable, BigQueryWriteClient client)
```
elefeint, line 55: Is this object generally meant to be re-used between different table? From the usage in `BigQueryTemplate` it does not seem so. Would it make sense to make this object immutable, meant for one-time use against a single table? Right now, some very strange things would happen if `initialize()` is called the second time while the first write is in progress.

elefeint (changes requested):
First general pass. @burkedavison will review in more detail from Java perspective.

spring-cloud-gcp-samples/spring-cloud-gcp-bigquery-sample/src/main/java/com/example/WebController.java
```diff
@@ -49,6 +54,99 @@ public ModelAndView renderIndex(ModelMap map) {
...
+  }
+
+  private ModelAndView getWriteApiResponse(
+      ListenableFuture<WriteApiResponse> writeApiFuture, String tableName) {
+    String message = null;
+    try {
+      WriteApiResponse apiResponse = writeApiFuture.get();
+      if (apiResponse.isSuccessful()) {
+        message = "Successfully loaded data to " + tableName;
+      } else if (apiResponse.getErrors() != null && apiResponse.getErrors().size() > 0) {
```
burkedavison, line 134: Following Effective Java recommendations, could the contract of `getErrors()` be modified to always return a collection -- even if that collection is empty?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
```diff
@@ -0,0 +1,148 @@
...
+          String.format("\nAppend %d success", response.getAppendResult().getOffset().getValue()));
+      done();
+    }
+
+    public void onFailure(Throwable throwable) {
+      synchronized (this.parent.lock) {
+        if (this.parent.error == null) {
+          StorageException storageException = Exceptions.toStorageException(throwable);
+          this.parent.error =
+              (storageException != null) ? storageException : new RuntimeException(throwable);
```
burkedavison, line 136: I'm coming in to this a little late in the game, but has there been discussion around the error handling design for this class?

Wrapping this Throwable in a RuntimeException and only saving the first one seems like it might create some limitations for the user of this class to get the full context of the problem -- particularly if they're performing a `catch` on a particular exception that may now be hidden one level down within the wrapper.

Should the caller of the BigQueryJsonDataWriter provide an onFailure callback? Or perhaps the BigQueryJsonDataWriter.append method should take the AppendCompleteCallback as a parameter? (In both cases: the goal being to delegate the error handling decisions one level higher.)

I'm not sure. Just for consideration.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: sure, I have added `spring.cloud.gcp.bigquery.jsonWriterBatchSize` property. Also we are checking if it's set too low (or not set at all), in that case it will be defaulted to `JSON_STREAM_WRITER_BATCH_SIZE`

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: Definitely, I have added it as a feedback @ go/bq-write-api-sc-gcp-integration. We will look into it later.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: Good idea, captured it as a feedback @  go/bq-write-api-sc-gcp-integration

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: I think calling cancel isn't needed in the `success` callback, as we would have already set the value and the `runAsync` task would have gotten completed. Is it okay if I remove the `success` callback and just keep the failure callback with the `cancel` call?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: This makes sense, I have made a note of it and will update the code accordingly

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
prash-mi: Thanks for pointing this out, This is supposed to be initialised just once for the given table. I have added a parameterised constructor @ `BigQueryJsonDataWriter` to initialise class and have removed the `initialize()` method. Also, I have marked `streamWriter` as `final`

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -159,16 +181,152 @@ public ListenableFuture<Job> writeDataToTable(
...
+    TableName parentTable =
+        TableName.of(bigQuery.getOptions().getProjectId(), datasetName, tableName);
+
+    BigQueryJsonDataWriter writer = new BigQueryJsonDataWriter();
+    // Initialize a write stream for the specified table.
+    writer.initialize(parentTable, client);
```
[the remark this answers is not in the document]
prash-mi: Thanks for having a look @burkedavison , I have refactored the class and have implemented `AutoClosable` as well.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
```diff
@@ -0,0 +1,148 @@
...
+  public String getStreamName() {
+    return streamWriter.getStreamName();
+  }
+
+  static class AppendCompleteCallback implements ApiFutureCallback<AppendRowsResponse> {
+    private final Logger logger = Logger.getLogger(this.getClass().getName());
+    private final BigQueryJsonDataWriter parent;
+
+    public AppendCompleteCallback(BigQueryJsonDataWriter parent) {
+      this.parent = parent;
```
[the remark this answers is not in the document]
prash-mi: sure. Refactored it accordingly

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
```diff
@@ -0,0 +1,148 @@
...
+    WriteStream writeStream = client.createWriteStream(createWriteStreamRequest);
+
+    // Use the JSON stream writer to send records in JSON format.
+    // For more information about JsonStreamWriter, see:
+    // https://googleapis.dev/java/google-cloud-bigquerystorage/latest/com/google/cloud/bigquery/storage/v1beta2/JsonStreamWriter.html
+    streamWriter =
+        JsonStreamWriter.newBuilder(writeStream.getName(), writeStream.getTableSchema()).build();
+  }
+
+  public void append(JSONArray data, long offset)
```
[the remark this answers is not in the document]
prash-mi: Sure, added javadoc

spring-cloud-gcp-samples/spring-cloud-gcp-bigquery-sample/src/main/java/com/example/WebController.java
prash-mi: sure. it will always return a `List<StorageError>` , even if empty

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -159,16 +181,152 @@ public ListenableFuture<Job> writeDataToTable(
...
+    // If the response does not have a commit time, it means the commit operation failed.
+    if (commitResponse.hasCommitTime() == false) {
+      for (StorageError err : commitResponse.getStreamErrorsList()) {
+        apiResponse.addError(err); // this object is returned to the user
+      }
+    }
+
+    // set isSucccessful flag to true of there were no errors
+    if (apiResponse.getErrors().size() == 0) {
+      apiResponse.setSuccessful(true);
```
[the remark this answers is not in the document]
prash-mi: sure, I have captured this as an AI @ go/bq-write-api-sc-gcp-integration and I will address it in a followup PR if it's okay

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
prash-mi: Makes sense and thanks for pointing this. I think we will need to improve the error handling design here. Say a user is writing data in bulk, like billions of records and say append failed for a few thousand. Then we should be having a mechanism to pass this message back to the user without interrupting the whole operation (as we might be still good as far as analytics is concerned).This same applies while committing the records. Let me take it back to the drawing board and address this in a followup PR.  Let me have a word with @elefeint  as well on this.

prash-mi: captured as an AI @ go/bq-write-api-sc-gcp-integration

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * @param data JSONArray to be appended
+   * @param offset offset at which data has to be added
+   */
+  public void append(JSONArray data, long offset)
+      throws DescriptorValidationException, IOException, ExecutionException {
+    synchronized (this.lock) {
+      // If earlier appends have failed, we need to reset before continuing.
+      if (this.error != null) {
+        throw this.error;
```
elefeint, line 89: Should the error get reset based on comment above? It seems to be unrecoverable based on usage in the template.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
```diff
@@ -0,0 +1,156 @@
...
+    // Increase the count of in-flight requests.
+    inflightRequestCount.register();
+  }
+
+  /**
+   * Call this method before committing the stream
+   *
+   * @param client BigQueryWriteClient
+   */
+  public void finalizeWriteStream(BigQueryWriteClient client) {
```
elefeint, line 104: Since the client got passed in the constructor, would it make sense to persist it there as a field, and have `finalizeWriteStream()` method be no-argument?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
```diff
@@ -0,0 +1,156 @@
...
+    // Wait for all in-flight requests to complete.
+    inflightRequestCount.arriveAndAwaitAdvance();
+
+    // Verify that no error occurred in the stream.
+    synchronized (this.lock) {
+      if (this.error != null) {
+        throw this.error;
```
elefeint, line 111: When errors get thrown in different places, does `close()` method need to be called to clean up the writer?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
```diff
@@ -0,0 +1,156 @@
...
+    // Close the connection to the server.
+    streamWriter.close();
+  }
+
+  class AppendCompleteCallback implements ApiFutureCallback<AppendRowsResponse> {
+    private final Logger logger = LoggerFactory.getLogger(AppendCompleteCallback.class);
+
+    public void onSuccess(AppendRowsResponse response) {
+      logger.info(
+          String.format("\nAppend %d success", response.getAppendResult().getOffset().getValue()));
```
elefeint, line 136: String.format() has a cost, but as it happens, slf4j `logger` has overloaded method variants that take a string pattern and parameters. This way, the string substitution will only be done if `INFO` level is needed.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
elefeint: I guess I am not sure why the callback is needed at all. Invocation of success callback means that ` writeApiFutureResponse.set(apiResponse);` got called successfully, which will lead to `completableFuture` completing. Invocation of error callback means that one of the `writeApiFutureResponse.setException(e);` calls was reached, again leading to `completableFuture`  completing correctly.
What is the state of `completableFuture` if the logic in the runnable completes successfully?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
elefeint: One more benefit is that it will make the code more testable -- if a `BigQueryWriteClient` object can come from the outside of the template, it can be mocked.

One thing I was not sure about -- is `BigQueryWriteClient` object meant to be reused between different write attempts (possibly to different tables)?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -159,16 +188,164 @@ public ListenableFuture<Job> writeDataToTable(
...
+
+    try {
+      // Write data in batches. Ref: https://cloud.google.com/bigquery/quotas#write-api-limits
+      long offset = 0;
+      int currentBatchSize = 0;
+
+      BufferedReader jsonReader = new BufferedReader(new InputStreamReader(jsonInputStream));
+      String jsonLine = null;
+      JSONArray jsonBatch = new JSONArray();
+      while ((jsonLine = jsonReader.readLine()) != null) { // read the input stream line by line
```
elefeint, line 279: Will every line of input be required to be a valid JSON object? What about fixed-length formatted input, or minimized input in which all objects are on a single line?

elefeint (changes requested):
The bad news about why there are no unit tests in the BigQuery integration is that [it was too difficult to unit test](https://github.com/spring-attic/spring-cloud-gcp/pull/1916#discussion_r327685714).

However, a lot of the code you are adding lends itself better to being unit tested -- for example, with `jsonWriterBatchSize`, setting it and making sure that the correct batch size is passed to the client object. Do what you can for the unit testing of `BigQueryTemplate` and `GcpBigQueryAutoConfiguration`, even if it won't reach the normal 80% threshold.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: Thanks @elefeint . I did some standalone testing for this and it appears that we do not need a success callback because of the reason you mentioned. However we might need a failure callback as the developer might call ` writeApiFutureResponse.cancel(true);` asynchronously and then we may want to interrupt the async operation which is writing the data. Please let me know if it looks good to you, we can later execute `asyncTask` on the developer's thread pool.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: Sure @elefeint , I have modified the code

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -57,6 +75,17 @@ public class BigQueryTemplate implements BigQueryOperations {
...
+  private static final int DEFAULT_JSON_STREAM_WRITER_BATCH_SIZE =
+      1000; // write records in batches of 1000
+
+  private static final int MIN_JSON_STREAM_WRITER_BATCH_SIZE = 10; // minimum batch size
+
+  private final Logger logger = LoggerFactory.getLogger(BigQueryTemplate.class);
+
+  // @Value("${spring.cloud.gcp.bigquery.jsonWriterBatchSize}") TODO(prasmish) debug exception while
+  // autowiring
+  private int jsonWriterBatchSize;
```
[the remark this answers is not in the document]
prash-mi: sure. I have now autowired accordingly

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
prash-mi: sure. updated the code

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
prash-mi: done

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: Do we want to restrict to newline delimited JSON format and then we can explore fixed-length formatted input, or minimised input as per the user's feedback. BQ Write API supports newLine Separated JSON out of the box (https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json) and we may have to do additional processing at the client library layer for any additional formats we add, which we can plan to add as per the user's demand. WDYT?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryJsonDataWriter.java
prash-mi: We get to know about the failure asynchronously @ `onFailure` and we throw this exception when the developer tries to append @ `BigQueryTemplate`. I have captured a task in the design document to better handle the errors (on the lines of what we had discussed, like returning the unique set of exceptions and the count of exceptions). Is it okay if we plan to address this in a follow-up PR ?

prash-mi:
> The bad news about why there are no unit tests in the BigQuery integration is that [it was too difficult to unit test](https://github.com/spring-attic/spring-cloud-gcp/pull/1916#discussion_r327685714).
> 
> However, a lot of the code you are adding lends itself better to being unit tested -- for example, with `jsonWriterBatchSize`, setting it and making sure that the correct batch size is passed to the client object. Do what you can for the unit testing of `BigQueryTemplate` and `GcpBigQueryAutoConfiguration`, even if it won't reach the normal 80% threshold.

Sure. Let me update you after adding unit tests.

prash-mi:
> > The bad news about why there are no unit tests in the BigQuery integration is that [it was too difficult to unit test](https://github.com/spring-attic/spring-cloud-gcp/pull/1916#discussion_r327685714).
> > However, a lot of the code you are adding lends itself better to being unit tested -- for example, with `jsonWriterBatchSize`, setting it and making sure that the correct batch size is passed to the client object. Do what you can for the unit testing of `BigQueryTemplate` and `GcpBigQueryAutoConfiguration`, even if it won't reach the normal 80% threshold.
> 
> Sure. Let me update you after adding unit tests.

@elefeint JFYI I have added UT for the new methods @ `BigQueryTemplate` and `GcpBigQueryAutoConfiguration`. Currently getting ~ 40% coverage

elefeint (changes requested):
Broadly the PR is looking great -- thank you for adding the tests! I've run through the sample, and that's very good, too.

I have some more minor comments inlined. 

Also, could you update [CHANGELOG](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/CHANGELOG.md) and take a look at Sonar code smells?

prash-mi:
> Broadly the PR is looking great -- thank you for adding the tests! I've run through the sample, and that's very good, too.
> 
> I have some more minor comments inlined.
> 
> Also, could you update [CHANGELOG](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/CHANGELOG.md) and take a look at Sonar code smells?

@elefeint . Sure, I see the last released version was `3.3.0`, I have added v `3.3.1` in the change log. Please let me know if it looks good. Also I have addressed the Sonar code smells

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
```diff
@@ -80,7 +89,24 @@ public BigQuery bigQuery() throws IOException {
 
   @Bean
   @ConditionalOnMissingBean
-  public BigQueryTemplate bigQueryTemplate(BigQuery bigQuery) {
-    return new BigQueryTemplate(bigQuery, this.datasetName);
+  public BigQueryWriteClient bigQueryWriteClient() throws IOException {
+    BigQueryWriteSettings bigQueryWriteSettings =
+        BigQueryWriteSettings.newBuilder()
+            .setCredentialsProvider(
+                FixedCredentialsProvider.create(this.credentialsProvider.getCredentials()))
```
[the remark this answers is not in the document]
prash-mi: sure, Done

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -159,16 +222,194 @@ public ListenableFuture<Job> writeDataToTable(
...
+  public WriteApiResponse getWriteApiResponse(String tableName, InputStream jsonInputStream)
+      throws DescriptorValidationException, IOException, InterruptedException {
+    WriteApiResponse apiResponse = new WriteApiResponse();
+    TableName parentTable =
+        TableName.of(bigQuery.getOptions().getProjectId(), datasetName, tableName);
+
+    // Initialize a write stream for the specified table.
+    BigQueryJsonDataWriter writer = getBigQueryJsonDataWriter(parentTable);
+
+    // new BigQueryJsonDataWriter(parentTable, bigQueryWriteClient);
```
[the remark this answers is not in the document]
prash-mi: sure, done

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/WriteApiResponse.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.spring.bigquery.core;
+
+import com.google.cloud.bigquery.storage.v1.StorageError;
+import java.util.ArrayList;
+import java.util.List;
+
+public class WriteApiResponse {
```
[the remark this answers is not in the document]
prash-mi: But we will be required to access it outside the package, for example our `com.example` package uses `WriteApiResponse`.  Also, the developers might use the public `ListenableFuture<WriteApiResponse> writeJsonStream` method in `BigQueryTemplate` and hence they will also need to access `WriteApiResponse` in their own packages. Please let me know if I am missing anything here.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -51,38 +70,82 @@ public class BigQueryTemplate implements BigQueryOperations {
...
    * Creates the {@link BigQuery} template.
    *
    * @param bigQuery the underlying client object used to interface with BigQuery
+   * @param bigQueryWriteClient the underlying BigQueryWriteClient reference use to connect with
+   *     BigQuery Storage Write Client
    * @param datasetName the name of the dataset in which all operations will take place
    */
-  public BigQueryTemplate(BigQuery bigQuery, String datasetName) {
-    this(bigQuery, datasetName, new DefaultManagedTaskScheduler());
+  public BigQueryTemplate(
```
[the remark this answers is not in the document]
prash-mi: This makes sense @elefeint , I have marked the two constructors ` @Deprecated` and we can use  `BigQueryTemplate(BigQuery bigQuery, BigQueryWriteClient bigQueryWriteClient, Map<String, Object> bqInitSettings, TaskScheduler taskScheduler)` .

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -51,38 +70,82 @@ public class BigQueryTemplate implements BigQueryOperations {
...
+   * @param bqInitSettings Properties required for initialisation of this class
+   */
+  public BigQueryTemplate(
+      BigQuery bigQuery,
+      BigQueryWriteClient bigQueryWriteClient,
+      Map<String, Object> bqInitSettings) {
+    this(
+        bigQuery,
+        bigQueryWriteClient,
+        (String) bqInitSettings.get("DATASET_NAME"),
```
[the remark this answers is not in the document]
prash-mi: done

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/WriteApiResponse.java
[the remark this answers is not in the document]
elefeint: That makes sense; it was meant to be part of public API. Thanks!

elefeint (approved):
Thank you very much! I'm sorry it took so long to review.

I'll merge the PR in, and move the changelog from 3.3.1 to 3.4.0-SNAPSHOT (the next release will be 3.4.0).

It would also be good to add documentation in a separate PR -- unfortunately right now we have to duplicate documentation between `bigquery.adoc` and `bigquery.md` in preparation for doc publishing migration.

prash-mi:
@elefeint Thanks for merging the PR. I will add documentation in both the files (`bigquery.adoc` and `bigquery.md`)

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -159,16 +226,187 @@ public ListenableFuture<Job> writeDataToTable(
...
    */
+  @Override
+  public ListenableFuture<WriteApiResponse> writeJsonStream(
+      String tableName, InputStream jsonInputStream) {
+
+    SettableListenableFuture<WriteApiResponse> writeApiFutureResponse =
+        new SettableListenableFuture<>();
+
+    Thread asyncTask =
+        new Thread(
```
meltsufin, line 276: It seems dangerous to create an unbounded number of threads without a thread pool here. It might be causing the OutOfMemoryError reported in #1599.

cc: @prash-mi, @burkedavison

burkedavison: +1 : Stack trace in the issue points to [asyncTask.start()](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/v3.4.0/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java#L310)

Good stackoverflow on the issue here: https://stackoverflow.com/a/16789621

PR #1241, opened 2022-08-26
fix: remove gson bean creation in spanner auto-configuration. 

zhumin8:
Discovered from Spring Boot Doc that auto-configuration for Gson is already provided ([doc](https://docs.spring.io/spring-boot/docs/2.6.x/reference/html/features.html#features.json.gson), [`GsonAutoConfiguration`](https://github.com/spring-projects/spring-boot/blob/ed42823bd2e7a526572374a0d25ba4e455ce4aeb/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/gson/GsonAutoConfiguration.java)). Configuration properties are also provided for customizing.

Removing the Gson bean creation in Spanner auto-configuration in favor of this one. This bean was original created in #937 for the conversion when dealing with JSON/ARRAY<JSON> Spanner type.

- [x] review documentation if changes needed.


elefeint (approved):
Nice, I did not know either.

elefeint (approved):

PR #1242, opened 2022-08-30
Modify spanner sample docs to instruct user running spanner using emulator

JoeWang1127:
Property `spring.cloud.gcp.spanner.emulator.enabled` is missing in spanner sample docs. It is required per [Spring Cloud GCP Spanner doc](https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#configuration-5) if the user want to run an instance on emulator.

### DIFF
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -1507,7 +1507,12 @@ Once the Spanner emulator is running, ensure that the following properties are s
 
 ----
 spring.cloud.gcp.spanner.emulator.enabled=true
-spring.cloud.gcp.spanner.emulator-host=${EMULATOR_HOSTPORT}
+----
+
+Note that the default emulator hostname and port (i.e., localhost:9010) is used. If you prefer a customized value, ensure the following property is set in your `application.properties` of your Spring application:
+
+----
+spring.cloud.gcp.spanner.emulator-host=ip:port
 ----
 
 === Sample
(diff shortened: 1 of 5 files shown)


elefeint (changes requested):
`export SPANNER_EMULATOR_HOST=localhost:9010` is using the client library mechanism for pointing at the emulator directly. I think that either setting `spring.cloud.gcp.spanner.emulator.enabled` or setting `SPANNER_EMULATOR_HOST` will work. But I also think we should be instructing users to use Spring Cloud GCP conveniences, so this PR is useful.

Could you test both ways (with only `spring.cloud.gcp.spanner.emulator.enabled` , and then with only `SPANNER_EMULATOR_HOST`, and then if both work the same way, remove the `SPANNER_EMULATOR_HOST` part of the instruction?

JoeWang1127:
> `export SPANNER_EMULATOR_HOST=localhost:9010` is using the client library mechanism for pointing at the emulator directly. I think that either setting `spring.cloud.gcp.spanner.emulator.enabled` or setting `SPANNER_EMULATOR_HOST` will work. But I also think we should be instructing users to use Spring Cloud GCP conveniences, so this PR is useful.
> 
> Could you test both ways (with only `spring.cloud.gcp.spanner.emulator.enabled` , and then with only `SPANNER_EMULATOR_HOST`, and then if both work the same way, remove the `SPANNER_EMULATOR_HOST` part of the instruction?

`spring.cloud.gcp.spanner.emulator.enabled` is required to be present in `application.properties` and set to `true` in order for the application connects to spanner emulator, otherwise `GcpSpannerEmulatorAutoConfiguration` won't get registered.

However, `SPANNER_EMULATOR_HOST` is not needed as the application will connect to the default emulator host and post (localhost:9010). I've modified the docs to reflect this.

elefeint (approved):
Thanks!

PR #1260, opened 2022-09-12
Support publisher shutdown

JoeWang1127:
See https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/864.

Add `@PreDestroy` method to shutdown cached `Publisher` in `CachingPublisherFactory`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,12 @@ refer to the [commit
 history](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/commits/main)
 on GitHub.
 
-## 3.4.0.BUILD-SNAPSHOT
+## 3.4.0-SNAPSHOT
+
+### Pub/Sub
+
+  - Allow Publishers shutdown gracefully
+    (<https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1260>)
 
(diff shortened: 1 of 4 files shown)


elefeint (changes requested):
Could you also try to add an integration test for this? I think the easiest way would be to take an existing integration test, get a `Publisher` instance from the template's `publish()` call, then initiate application shutdown (by autowiring an ApplicationContext). Then to try to publish a message on the already-shutdown `Publisher` should throw an exception from the client library - "Cannot publish on a shut-down publisher.".

JoeWang1127:
> Could you also try to add an integration test for this? I think the easiest way would be to take an existing integration test, get a `Publisher` instance from the template's `publish()` call, then initiate application shutdown (by autowiring an ApplicationContext). Then to try to publish a message on the already-shutdown `Publisher` should throw an exception from the client library - "Cannot publish on a shut-down publisher.".

Thanks for the advice, will do.

elefeint (approved):
Thank you! Can you add this to the CHANGELOG, either in this PR or in a follow-up?

JoeWang1127:
> Thank you! Can you add this to the CHANGELOG, either in this PR or in a follow-up?

I'll make the change in this PR, thanks.

PR #1284, opened 2022-10-10
Fix bigquery integration tests

elefeint:
Each test runs in isolation, but when run together, the different schema requirement in `testJsonFileLoadWithSchema` causes it to fail.

Creating a clean table for each test in bigquery to allow each test to specify its own schema fixes the immediate issue.

This testing issue also uncovered a real problem (tests are good!!) -- if writing fails with anything but `DescriptorValidationException`, `IOException` or `InterruptedException`, the thread failure is not passed into the returned future. I'll send a PR for this separately.


prash-mi (approved):
Thanks for making the changes @elefeint

burkedavison (approved):

PR #1285, opened 2022-10-10
Catch generic throwable and pass exception to returned future

elefeint:
This PR adds a catch-all for any unforeseen errors and passes the error to the returned future.

In normal code, catching `Throwable` is a bad idea -- the error should be allowed to propagate and stop the app. However, when running in a separate thread, an uncaught `Throwable` will simply disappear without a trace, so we have to catch it and hand the error over to the returned future, so that the failure may be acted upon.

Unrelated changes:
* lightly refactored to have the template get created before each test, since every test will need it.
* somehow this test was still on JUnit4. Upgraded to JUnit5.

### DIFF
--- a/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
+++ b/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
@@ -285,6 +285,9 @@ public class BigQueryTemplate implements BigQueryOperations {
                 writeApiFutureResponse.setException(e);
                 // Restore interrupted state in case of an InterruptedException
                 Thread.currentThread().interrupt();
+              } catch (Throwable t) {
+                writeApiFutureResponse.setException(t);
+                Thread.currentThread().interrupt();
               }
             });
     asyncTask
--- a/spring-cloud-gcp-bigquery/src/test/java/com/google/cloud/spring/bigquery/BigQueryTemplateTest.java
+++ b/spring-cloud-gcp-bigquery/src/test/java/com/google/cloud/spring/bigquery/BigQueryTemplateTest.java
@@ -16,10 +16,12 @@
 
 package com.google.cloud.spring.bigquery;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -285,6 +285,9 @@ public ListenableFuture<WriteApiResponse> writeJsonStream(
                 writeApiFutureResponse.setException(e);
                 // Restore interrupted state in case of an InterruptedException
                 Thread.currentThread().interrupt();
+              } catch (Throwable t) {
```
prash-mi, line 288: @elefeint Catching a wider exception is a good idea here and thanks for making the changes. However shall we catch `Exception` instead? Catching `Throwable` results in this Sonar issue: https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=1285&id=GoogleCloudPlatform_spring-cloud-gcp&open=AYPCUFuggImngqXU1xdj

meltsufin (approved):
I'm not sure what would be a better way to handle exceptions from the asynchronous task. In this case, catching `Throwable` or `Exception` seems appropriate.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
elefeint: Sonar is right, in general, however the user experience will be significantly worse if we follow this recommendation -- for example, if an unexpected Exception (or even `Throwable` such as linkage errors) arises during the separate thread execution, the error will just disappear without leaving a trace.

elefeint: That causes 2 issues:
1) The error disappears without any user-visible evidence.
2) `CompletableFuture` will be left hanging forever.

prash-mi: Makes sense @elefeint

PR #1286, opened 2022-10-11
Make CreateDisposition configurable on BigQueryTemplate

seakayone:
This  change is adding the possibility to be able to change the `CreateDisposition` (`CREATE_IF_NEEDED` or `CREATE_NEVER`) on the `BigQueryTemplate` ackin to how you would currently configure the `WriteDisposition`.

In our use case we have to use `CREATE_NEVER` because we want to prevent even the attempt to create a table in BigQuery. Currently this is only be possible by using a `BigQuery` instance directly, because the `CreateDisposition` is not passed through by the template to the `WriteChannelConfiguration` and eventually the writer falls back to the `CREATE_IF_NEEDED` default.


elefeint (approved):
Thank you!

@prash-mi FYI -- eventually, perhaps both WriteDisposition and CreateDisposition could become properties.

PR #1307, opened 2022-10-24
feat(codegen): starter script to generate one module

zhumin8:
A naive starter script to start working on the code generation.

- from `/generator` run with `./generate-one.sh -c vision -v 3.1.2 -i google-cloud-vision -g com.google.cloud`
- this script will generate `/vision` under `/generated` folder with spring code and pom.xml, also adds module to `generated-parent`.


elefeint (approved):

zhumin8:
force-pushed after rebase to main.

elefeint (changes requested):
The `generated/pom.xml` needs to have a packaging specified -- otherwise when building with Maven, you get:
`'packaging' with value 'jar' is invalid. Aggregator projects require 'pom' as packaging.`.

generator/generate-one.sh
```diff
@@ -0,0 +1,82 @@
...
+# get into gapic and checkout branch to use
+cd gapic-generator-java
+git checkout autoconfig-gen-draft2
+# go back to previous folder
+cd -
+
+
+cd googleapis
+# todo: change to local repo --> gapic
+# very not stable change - todo: change to search and replace.
```
emmileaf, line 49: To future-proof for the time being, maybe we can check out googleapis on a specific branch or commit in case this file gets updated?

elefeint (changes requested):
The starters' pom files either need to have a parent or need to set the target Java version. Right now, you get this when building, say, `com-google-cloud-vision-v1-spring-starter`: 

```
[ERROR] Source option 5 is no longer supported. Use 7 or later.
[ERROR] Target option 5 is no longer supported. Use 7 or later.
```

elefeint (approved):

PR #1308, opened 2022-10-25
feat(autoconfig): autoconfig module

diegomarquezp:
has shared properties to be used by generated spring autoconfigured client beans


generator/shared-autoconfig-properties/src/main/java/com/google/cloud/spring/shared/SharedProperties.java
```diff
@@ -0,0 +1,19 @@
...
+import com.google.cloud.spring.core.CredentialsSupplier;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
+
+@ConfigurationProperties("spring.auto.shared")
+public class SharedProperties implements CredentialsSupplier {
+
+    // Overrides the GCP OAuth2 credentials specified in the Core module.
+    @NestedConfigurationProperty
+    private final Credentials credentials = new Credentials("https://www.googleapis.com/auth/cloud-language");
```
zhumin8, line 13: hmm, I missed this part in my poc. 
In generated modules' code, this scope comes from service, and I think the origin is [here](https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1/vision_v1.yaml#L39-L40).
What should we have for shared credentials though? 
- Is "https://www.googleapis.com/auth/cloud-platform" enough, since we are aiming for clients under `google/cloud`? 
- Or do we need to add service specific scopes too?
- spring-cloud-gcp core shared credentials has its scopes as empty [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/core/GcpProperties.java#L34)

generator/shared-autoconfig-properties/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+            <artifactId>gax</artifactId>
+            <version>2.18.2</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-language</artifactId>
+            <version>2.3.0</version>
+            <scope>compile</scope>
+        </dependency>
```
zhumin8, line 40:
```suggestion

```

generator/shared-autoconfig-properties/pom.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <!-- Does it make sense to add spring-cloud-gcp as parent? -->
```
zhumin8, line 7: Although this is manual-written module, I feel it makes sense to have it in `generated` folder alongside with generated code:
- they can be in same released schedule, and `generator` will only contain internal process code.
- they can also share same parent pom and dependencies

What do you all think? @elefeint @diegomarquezp @emmileaf

zhumin8 (changes requested):

generator/shared-autoconfig-properties/src/main/java/com/google/cloud/spring/shared/SharedProperties.java
emmileaf: Spring Cloud GCP documentation mentions that default credentials [contain scopes for every supported service](https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#scopes).

Wrapping my head around this, I think the top-level `GcpProperties` have scopes as [empty](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/93a3eb540e14d3c1befb2b9a0e6ebc1b1e5dbd16/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/core/GcpProperties.java#L34) if not overridden by service-specific properties, but is later passed into a `DefaultCredentialsProvider` that further processes the scopes list.

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/93a3eb540e14d3c1befb2b9a0e6ebc1b1e5dbd16/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/core/GcpContextAutoConfiguration.java#L56

When the list is empty, scopes for all of Spring Cloud GCP’s supported services seem to be added here:

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/93a3eb540e14d3c1befb2b9a0e6ebc1b1e5dbd16/spring-cloud-gcp-core/src/main/java/com/google/cloud/spring/core/DefaultCredentialsProvider.java#L132

---

Looking at the oauth scopes [docs](https://developers.google.com/identity/protocols/oauth2/scopes), just having the `cloud-platform` probably isn't enough, but I wonder if we can have the service-specific scopes list be picked up from a config file instead of the exhaustive list approach?

generator/shared-autoconfig-properties/pom.xml
emmileaf: Having this module in the same place as the generated modules makes sense to me!

diegomarquezp: Makes sense to me as well. I will move it to `generated`

generated/shared-autoconfig-properties/src/main/java/com/google/cloud/spring/shared/SharedProperties.java
```diff
@@ -0,0 +1,19 @@
+package com.google.cloud.spring.shared;
+
+import com.google.cloud.spring.core.Credentials;
+import com.google.cloud.spring.core.CredentialsSupplier;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
+
+@ConfigurationProperties("spring.auto.shared")
```
elefeint, line 8: End users are likely to have other modules that are not GCP in their applications. Typically, prefix starting with `spring` is for properties that are managed by Spring; it would be good if this prefix included "gcp" or equivalent "google.cloud" somewhere.

elefeint (changes requested):

generator/shared-autoconfig-properties/src/main/java/com/google/cloud/spring/shared/SharedProperties.java
diegomarquezp: I added a `Credentials()` call without arguments to use all scopes

generated/shared-autoconfig-properties/src/main/java/com/google/cloud/spring/shared/SharedProperties.java
diegomarquezp: Changed to `com.google.cloud.global`

generated/global-autoconfig-properties/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter</artifactId>
+            <version>2.7.3</version>
```
diegomarquezp, line 26: Should versions be inherited from `generated-parent`?

zhumin8: Yes! You could either decide to do it in this pr or we can do a separate refactor pr along with the same thing in the [module poms](https://github.com/googleapis/gapic-generator-java/blob/autoconfig-gen-draft2/src/main/java/com/google/api/generator/spring/SpringWriter.java#L243-L254). 
btw, I see some unused dependencies in this file that can be removed.

diegomarquezp: Alright! Just removed the dependencies.

generated/global-autoconfig-properties/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+        <version>3.5.0-SNAPSHOT</version>
+    </parent>
+
+
+    <artifactId>global-spring-autoconfig-properties</artifactId>
+    <name>Spring Cloud GCP - Global Autoconfig Properties</name>
+    <description>Common properties for generated spring autoconfigurations</description>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
```
blakeli0, line 19: Does Spring Cloud GCP support Java 8? If yes, we need to make this `1.8`.

diegomarquezp: @zhumin8

generated/global-autoconfig-properties/src/main/java/com/google/cloud/spring/global/GlobalProperties.java
```diff
@@ -0,0 +1,19 @@
+package com.google.cloud.spring.global;
```
emmileaf, line 1: Could you add a license header to this file (similar to the other files we are generating), and run `google-java-format` (see [repo's style guide](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/CONTRIBUTING.md#code-style)) to fix the formatting here? 

Just realized that `mvn verify` (through the integration tests workflow) runs checkstyle in this repo, so the check probably slipped away when we disabled presubmit ITs for codegen PRs.

emmileaf (changes requested):

generated/global-autoconfig-properties/pom.xml
emmileaf: `spring-cloud-gcp` has unit tests run against Java 8, 11, and 17 - so we probably should align with the parent pom here: 
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/631a1af70a8f22cb2913ffba094284f707540437/pom.xml#L57-L58

Actually, given the parent structures, I wonder if these properties are already visible and inherited by this child pom? (Haven't verified this locally, but if that's the case we can remove this properties block as well.)

generated/global-autoconfig-properties/src/main/java/com/google/cloud/spring/global/GlobalProperties.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.google.cloud.spring.core.Credentials;
+import com.google.cloud.spring.core.CredentialsSupplier;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
+
+@ConfigurationProperties("com.google.cloud.global")
```
emmileaf, line 8:
```suggestion
@ConfigurationProperties("com.google.cloud.spring.global")
```

generated/global-autoconfig-properties/pom.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>generated-parent</artifactId>
```
emmileaf, line 9: We'll also need to add `global-autoconfig-properties` as a child module into [spring-cloud-gcp/generated/pom](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/489874079dafc07f4d6810c097f08b2e1db83693/generated/pom.xml#L15-L16).

generated/global-autoconfig-properties/pom.xml
diegomarquezp: Yes, I saw it's defined in the root pom, so it should be inherited. Removing!

generated/global-autoconfig-properties/src/main/java/com/google/cloud/spring/global/GlobalProperties.java
diegomarquezp: Fixed format and added header

generated/global-autoconfig-properties/pom.xml
diegomarquezp: I'm adding both `spring-boot-starter` and `spring-cloud-gcp-core` to `generated-parent`

generated/pom.xml
```diff
@@ -12,6 +12,20 @@
   <artifactId>generated-parent</artifactId>
   <packaging>pom</packaging>
```
emmileaf, line 13: Ah - possibly the unit test checks are failing now because we made this a multi-module maven project, but did not specify this line `<packaging>pom</packaging>` also in global-autoconfig-properties/pom. Could you try if this is the fix? 

We also probably don't need the dependency block below, since they should be [inherited from parent](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1353#discussion_r1030764468).

diegomarquezp: This did it. Thanks Emily!

emmileaf (approved):

generated/global-autoconfig-properties/pom.xml
```diff
@@ -0,0 +1,18 @@
...
+        <version>3.5.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>global-spring-autoconfig-properties</artifactId>
+    <packaging>pom</packaging>
```
zhumin8, line 14: I think only the parent pom needs to be marked as `<packaging>pom</packaging>` so maven won't produce further artifacts. Submodules should have the default packaging as jar. For reference, checkout submodules under`spring-cloud-gcp-samples`.

generated/pom.xml
diegomarquezp: Makes sense - thanks, Min. 
I changed `global-autoconf...` to look like the datastore pom. Checks are passing locally.

generated/global-autoconfig-properties/src/main/java/com/google/cloud/spring/global/GlobalProperties.java
emmileaf: @zhumin8 you are right - I just saw that checkstyle is actually bound to the [validate](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/0489cf907df695379b9d1d4a41a0c9b3241ccfb0/pom.xml#L260) phase, which is covered by unit tests. I had the wrong understanding here earlier when trying to reason why the checks didn't catch this file.

In that case, at the time of this comment, this was probably not picked up because the global-autoconfig-properties wasn't bound to the generated-parent pom yet (which has now been updated).

diegomarquezp: Checkstyle currently [passing](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/3577646038/jobs/6016883951#step:7:217)

generator/shared-autoconfig-properties/src/main/java/com/google/cloud/spring/shared/SharedProperties.java
zhumin8: Revisiting the scope issue. I am okay if you want to split and fix in a separate PR.
I don't think `Credentials()` call without arguments includes all scopes. as @emmileaf pointed out above, scopes listed in [`CREDENTIALS_SCOPES_LIST`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-core/src/main/java/com/google/cloud/spring/core/GcpScope.java) will be added if non specified. 
I feel comfortable adding all scopes for the generated libraries to this or a similar list. However, I am thinking, since our approach now is creating the credential provider in the autoconfig code:
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/9d22b0222be796f67291991d81573f92d8667157/generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java#L81-L86
We can probably add additional scopes when using the `globalProperties`. see POC here: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1366

What do you all think?

zhumin8:
Perhaps get some test coverage with a simple test with fake credential key? 
For reference: https://www.baeldung.com/spring-boot-testing-configurationproperties#binding-POJOs

generator/shared-autoconfig-properties/src/main/java/com/google/cloud/spring/shared/SharedProperties.java
meltsufin: I'm not sure, but I think `https://www.googleapis.com/auth/cloud-platform` should be enough for shared credentials.

diegomarquezp: Added `https://www.googleapis.com/auth/cloud-platform` as default scope.

meltsufin (commented):
Why are these global properties specific to the auto-generated modules? Shouldn't they be in autoconfiguration core?

elefeint (approved):
LGTM as far as I'm concerned, but I've added Mike's question to the design doc's section.

diegomarquezp:
Closing as not planned

PR #1310, opened 2022-10-31
feat(cicd): ignore IT when changes on generat[ed|or]/

diegomarquezp:
Modified the IT workflow to perform a check for changes in the `generated/` and `generator/` folder. If any changed file belongs to these folders, the ITs will not be run and a label will be added to indicate this behavior (the label does not trigger anything, it is purely informative)

- [X] add `generator/` folder to triggers as well
- [X] add label to indicate ITs were skipped
- [X] remove label if case new commits omit the changes in `generat[ed|or]`


diegomarquezp:
Pending 
- add `generator/` folder to triggers as well
- add label to indicate ITs were skipped

.github/workflows/integrationTests.yaml
```diff
@@ -10,8 +10,52 @@ on:
...
+          while IFS= read -r file
+          do
+            echo $file
+            if [[ $file != generated/* ]] && [[ $file != generator/* ]]; then
+              echo "This modified file is not under the 'generated' nor 'generator' folders."
+              echo "::set-output name=run_job::true"
+            else
+              echo "found a file that belongs in generated/ or generator/"
+              echo "::set-output name=run_job::false"
+              break
```
zhumin8, line 40: If I am not mistaken, the current logic says if there is at least one file altered belonging to `generat[ed|or]/`, then skip it-tests. As a result, it-test is skipped in your test settings with `generator/dummy-it-trigger.txt` present, even though we also have alterations in this yaml file, which should technically trigger integration tests.

zhumin8 (commented):
> If any changed file belongs to these folders
I think we want "if **all** changes belong to these folders" to skip IT.

.github/workflows/integrationTests.yaml
zhumin8: btw, from workflow outputs, `set-output` seems to be newly deprecated. 
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

diegomarquezp: Makes sense, I will trigger ITs when at least one file is outside `generat[ed|or]/`

.github/workflows/integrationTests.yaml
```diff
@@ -10,8 +10,52 @@ on:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: check modified files
+        id: check_files
+        run: |
+          echo "=============== list modified files ==============="
+          git diff --name-status HEAD^ HEAD
```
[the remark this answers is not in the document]
diegomarquezp: It was mostly intended for debugging. I'm removing the other unnecessary `echo` statements

emmileaf (approved):
Ah I think when I made the review suggestion from web UI it registered my personal email, hence the [co-authored commit](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1310/commits/9c850763d30965e5da50f7471962b77255d1d95e) breaking the CLA check. Sorry and feel free to wipe over that commit!

.github/workflows/integrationTests.yaml
```diff
@@ -10,8 +10,43 @@ on:
...
+        id: check_files
+        run: |
+          git diff --name-only HEAD^ HEAD > files.txt
+          while IFS= read -r file
+          do
+            if [[ $file != generated/* ]] && [[ $file != generator/* ]]; then
+              echo "run_job=true" >> $GITHUB_OUTPUT
+              break
+            else
+              echo "run_job=false" >> $GITHUB_OUTPUT
```
elefeint, line 33: You could skip the `else` by setting `run_job=false` before the loop. Then instead of getting written on each "excluded" file, it defaults to false and only gets changed once if a non-excluded file is found.

elefeint (approved):

PR #1313, opened 2022-11-01
feat(codegen): script to generate all files

diegomarquezp:
modularizes repo download in case generate-one.sh needs to be called independently. generate-all.sh reads the library csv and calls generate-one.sh for each library. repo download is used only once before the generation jobs and then removed

pending:
- [x] specify comittish in `gapic-generator-java` to prevent bazel rule separation issues mentioned in the [main PR](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1307/files)


generator/generate-one.sh
```diff
@@ -30,32 +33,13 @@ echo "Client Library ArtifactId: $client_lib_artifactid";
...
-cd googleapis
-# fix googleapis committish for test/dev purpose
-git checkout f88ca86
-# todo: change to local repo --> gapic
-# very not stable change - todo: change to search and replace.
-sed -i '274,278d' WORKSPACE
-
-sed -i '274 i local_repository(\n    name = "gapic_generator_java",\n    path = "../gapic-generator-java/",\n)' WORKSPACE
+if [[ $download_repos -eq 1 ]]; then
+  bash download-repos.sh
```
zhumin8, line 37: I suppose we can move the download and delete repo logic into the `generate-all.sh` so it's more efficient. 
Then in future, when we have the metadata as discussed today, we can checkout specific commitish here for each client library potentially.

generator/download-repos.sh
```diff
@@ -0,0 +1,27 @@
...
+git checkout c51d51bc495ea1284087b66c369dedf741dd4824
+# go back to previous folder
+cd -
+
+
+cd googleapis
+# fix googleapis committish for test/dev purpose
+git checkout f88ca86
+# todo: change to local repo --> gapic
+# very not stable change - todo: change to search and replace.
```
emmileaf, line 21: Dropping in a quick note here for the update merged in #1315, since it might not get sync'd or show up as conflict after the file split.

generator/generate-one.sh
diegomarquezp: Oh, I'm actually downloading the repos _once_ in `generate-all.sh:8`. This piece of code was intended in case we want to call `generate-one.sh` directly and it's not reached when this script is called from `generate-all.sh`.

diegomarquezp:
Force pushed a rebase in main to have changes from #1315 and #1307

zhumin8 (approved):
Looks like `generator/pom.xml` fall through the cracks somehow. We have `generated/pom.xml` and don't need another one in `generator/`.

Otherwise LGTM!

emmileaf:
@diegomarquezp Ah, sorry I didn't follow-up on [this comment](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1313#discussion_r1013282373), but looks like the changes from #1315 got updated into `generate-one.sh` when this PR intended to extract it into `download-repos.sh`?

diegomarquezp:
@emmileaf True! The rebase was successful but I oversaw that the new replacement change should actually be moved `download-repos.sh` (so it's only ran once). I will submit a small fix PR for this!

PR #1315, opened 2022-11-02
fix(codegen): use find and replace to switch to local gapic-generator-java repo

emmileaf:
To replace `http_archive()` rule with `local_repository()` rule for using local repo of `gapic-generator-java`, this PR updates the current approach in generator script of replacement by line number to search and replace with pattern matching.
- Some additional notes for the perl one-liner logic, not sure if they'd be helpful to include as inline comments?
```
# Using perl for multi-line replace
# -0777 slurps file instead of per-line
# -p for assuming input loop around program, like sed
# -i to edit file in-place
# -e to enter one line of program
# "s/{find_pattern}/{replace_text}/s": substitute first occurrence of text matching {find_pattern} with {replace_text}, with dot matching all characters including new lines
```

**Potential conflicts:** with #1313, will just need to update location of change if merged after.

### DIFF
--- a/generator/generate-one.sh
+++ b/generator/generate-one.sh
@@ -48,10 +48,19 @@ cd googleapis
 # fix googleapis committish for test/dev purpose
 git checkout f88ca86
 # todo: change to local repo --> gapic
-# very not stable change - todo: change to search and replace.
-sed -i '274,278d' WORKSPACE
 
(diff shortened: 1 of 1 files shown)


diegomarquezp (approved):
LGTM - safer to have str replacement than exact line numbers!

elefeint (approved):
Y'all might enjoy AWK.

zhumin8 (approved):
Thank you! - the line numbers were super hacky.

PR #1317, opened 2022-11-03
scripts to publish javadoc and doc to cloud-rad

zhumin8:
Adding kokoro scripts and related changes needed to assemble and upload to be published javadoc and documentation to cloud-rad process. Intend to get this in for some final test run before:
- changing from dev to prod bucket
- removing debug info from script.

Also in this pr: table format changes in `getting-started.md` and `pubsub.md`  to fix rendered html format issues.

update: switch to use a requirements.txt file with hash checking to securely install dependencies with pip install.


zhumin8:
force-pushed after rebase.

.kokoro/publish_javadoc11.cfg
```diff
@@ -0,0 +1,21 @@
...
+}
+
+before_action {
+  fetch_keystore {
+    keystore_resource {
+      keystore_config_id: 73713
+      keyname: "docuploader_service_account"
+    }
+  }
+}
```
elefeint, line 18: I think this config should live in g3.

.kokoro/publish_javadoc11.sh
```diff
@@ -0,0 +1,133 @@
...
+pyenv global 3.7.2
+
+# install pandoc for combining md files
+sudo apt install -q -y pandoc
+
+# Install docuploader package
+python3 -m pip install --upgrade six
+python3 -m pip install --upgrade protobuf
+python3 -m pip install gcp-docuploader==0.6.2
+
```
mpeddada1, line 28: The public_javadoc scripts in googleapis/ were recently migrated to use a requirements.txt file with hash checking to securely install dependencies. Maybe we could do the same here? Here's an example: https://github.com/googleapis/synthtool/pull/1588

zhumin8: updated with changes. Test run succeeded.
image

.kokoro/publish_javadoc11.cfg
zhumin8: copied from discussion: `before_action` is only configurable in `build.proto` not `job.proto`.

.kokoro/publish_javadoc11.sh
mpeddada1: Nice!

mpeddada1 (approved):

elefeint (approved):

PR #1318, opened 2022-11-03
fix(actions): suppress IT only on PRs

diegomarquezp:
Switching trigger logic to use `on`.
* PRs will trigger ITs only if
  * They have at least 1 change in a folder other than `generat[ed|or]/`. [details](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths)
  * The PR has been opened or [`synchronized`](https://github.com/orgs/community/discussions/25161#discussioncomment-3246673)
* Pushes to `main` will also trigger the ITs (e.g. merge a PR)

The logic for excluding ITs from dependabot was moved to the job's `if` entry

### DIFF
--- a/.github/workflows/integrationTests.yaml
+++ b/.github/workflows/integrationTests.yaml
@@ -34,7 +34,12 @@ jobs:
           done < files.txt
       - name: Add skip IT label
         run: |
-          cat $GITHUB_EVENT_PATH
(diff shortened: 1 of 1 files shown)


elefeint (approved):
Very nice!

emmileaf (approved):

PR #1322, opened 2022-11-04
draft(spring codegen): generate in-scope library list from monorepo

zhumin8:
The added script tries to generate `library_list.txt` from filtering info from [`google-cloud-java`](https://github.com/googleapis/google-cloud-java) for each client library.
This list is intended to use as a static data source to generate modules from. It can be replaced when `googleapis-folder` can be accessed from metadata.

- a commitish from `google-cloud-java` is needed to generate the list. 
- information from `.repo-metadata.json` and `versions.txt` are used.
-  criteria to be included in list: 
    - `library_type` is GAPIC_AUTO
    - `group_id` is com.google.cloud
    - `release_level` is stable
    - `transport` has "grpc" (grpc/grpc+rest)
 - output list includes: `api_shortname, distribution_name:version` and a few debug items at the tail.

A few discussion items: 
-  made assumption: when `group_id = com.google.cloud`, it is located under [googleapis/google/cloud](https://github.com/googleapis/googleapis/tree/master/google/cloud). 
- `.repo-metadata.json` is populated at time of library enrollment. 
- is `api_shortname` reliable to look for folder location under [googleapis/google/cloud](https://github.com/googleapis/googleapis/tree/master/google/cloud)? - No

Update:
- aligned format in `library_list.txt` to be used in `generate-one.sh`
- added heuristic approach to compute googleapis path borrowed from #1371.
  - Able to get 71 out of 91 in-scope repo list.
  - known omits from logic: if `v1` does not exists.
- Intent to merge this in to start testing on full list.


generator/generate-library-list.sh
```diff
@@ -0,0 +1,79 @@
...
+echo "# api_shortname, googleapis-folder, distribution_name:version" > $filename
+
+count=0
+for d in ./google-cloud-java/*java-*/; do
+  # parse variables from .repo-metadata.json
+  language=$(cat $d/.repo-metadata.json | jq -r .language)
+  api_shortname=$(cat $d/.repo-metadata.json | jq -r .api_shortname)
+  distribution_name=$(cat $d/.repo-metadata.json | jq -r .distribution_name)
+  library_type=$(cat $d/.repo-metadata.json | jq -r .library_type)
+  transport=$(cat $d/.repo-metadata.json | jq -r .transport)
```
blakeli0, line 40: The transport attribute in `repo-metadata.json` is not the source of truth, the source of truth is the bazel config in googleapis, like [here](https://github.com/googleapis/googleapis/blob/1051d7d7ffe66fdbdc14cbd2ec2eed3b8741cb4a/google/cloud/asset/v1/BUILD.bazel#L80), we've seen a lot of discrepancies before. If all we want to do is excluding rest only clients, then we can include everything in the mono repo except for `compute`

zhumin8:
Added logic to figure out the correct corresponding path in `googleapis`, with some known flaws with this approach. But since the intent for this static list is temporary, I would like to get it merged in and unblock testing generating full list of libraries.
note: #1371 and #1370 both had this change, removed from #1370 and @diegomarquezp you can keep #1371 as your testing ground or close when you are done.

generator/generate-library-list.sh
```diff
@@ -0,0 +1,95 @@
...
+  fi
+
+  # get version from versions.txt
+  # this changed in https://github.com/googleapis/google-cloud-java/pull/8755/
+  module_versions=$(sed -n "/^${artifact_id}:/p" ./google-cloud-java/versions.txt)
+
+  released_version=$(echo "$module_versions" | cut -f2 -d:)
+
+  # heuristic approach to compute googleapis path: uses first directory in
+  # googleapis matching "**/api_shortname/v[0-9]" (may bring any version)
```
emmileaf, line 71: Let's update the comments for this tweak from #1371 as well 
```suggestion
  # googleapis matching "**/api_shortname/v1"
```

emmileaf (approved):

diegomarquezp (approved):

PR #1336, opened 2022-11-11
codegen: adding documentation items for generated modules

zhumin8:
As discussed earlier, adding these items in prep for documentation:
- a `README.md` to `generated/` with header for  a list of modules generated.
- list item generation line into script.
- link from documentation page.

### DIFF
--- a/generator/generate-one.sh
+++ b/generator/generate-one.sh
@@ -73,6 +73,7 @@ else
   # also write to generated/README.md
   # format |name|distribution name|
   echo -e "|$client_lib_name|com.google.cloud:$starter_artifactid|" >> README.md
+  echo "$({(cat README.md | grep -vw ".*:.*");(cat README.md | grep ".*:.*" | sort | uniq)})" > README.md
 
 fi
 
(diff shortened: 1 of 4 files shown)


generator/generate-one.sh
```diff
@@ -68,6 +68,9 @@ if [[ found_library_in_pom -eq 0 ]]; then
 else
   echo "adding module $client_lib_name to pom"
   sed -i "/^  <modules>/a\ \ \ \ <module>"$client_lib_name"</module>" pom.xml
+  # also write to generated/README.md
+  # format |name|distribution name|
+  echo -e "|$client_lib_name|$client_lib_groupid:$client_lib_artifactid|" >> README.md
```
elefeint, line 73: We'll probably end up with duplicates over time. Might have to add some shell script like `cat README.md | grep .*:.* | sort | uniq >> README.md`

elefeint (approved):

zhumin8:
I missed the fact that we are adding our generated starter coordinates, not the corresponding client library's in the README list. Updated the changes. 
This needs to work with changes in https://github.com/googleapis/gapic-generator-java/pull/1092 to use `[client_lib_artifactid]-spring-starter` instead of `[package_name]-spring-starter` for starter artifact ids.

generator/generate-one.sh
zhumin8: The `if` logic checks for if this module is already added to the `generated-parent` pom. I am assuming here that a new module == not in `generated-parent` pom == not in `README.md`.
But there is no harm adding this de-dup logic just in case.

elefeint (approved):

PR #1341, opened 2022-11-11
chore(codegen): manual GH action to generate autoconfigs

diegomarquezp:


.github/workflows/generateAutoConfigs.yaml
```diff
@@ -0,0 +1,83 @@
+name: Generate Spring Auto-Configurations
+
+on:
+  push:
+    branches:
+      - 'main'
+  # what triggers this action is yet TBD - leaving PRs and `main` pushes for now
```
emmileaf, line 7: We can maybe switch to [`on: workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) for now to just [manually trigger](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow?tool=webui) this workflow.

diegomarquezp: Thanks @emmileaf - I modified the trigger to _only_ use `workflow_dispatch`

diegomarquezp:
Merged #1436, #1437, #1438 to have the latest changes to be merged in main. Please ignore changes other than `.github/workflows/generateAutoConfigs.yaml`

.github/workflows/generateAutoConfigs.yaml
```diff
@@ -0,0 +1,127 @@
...
+        run: |
+          set -x
+          set -e
+          bash generate-all.sh
+      - name: Check for generation errors
+        continue-on-error: false
+        working-directory: spring-cloud-generator
+        run: |
+          set -x
+          if [[ $(ls *failed-library-generations | wc -l) -eq 0 ]]; then
```
meltsufin, line 64: Should we check something got generated at all?

meltsufin (approved):

.github/workflows/generateAutoConfigs.yaml
diegomarquezp: I think the `Detect changes` step would stop the workflow if there were no generated files

.github/workflows/generateAutoConfigs.yaml
```diff
@@ -0,0 +1,127 @@
...
+          git stash push -- spring-cloud-previews/
+          git reset --hard
+          git remote update
+          git fetch
+
+          # pops the changes in a new branch from main
+          git checkout main
+          git checkout -b $NEW_BRANCH_NAME
+          #git checkout stash -- .
+          git stash pop
```
emmileaf, line 109: Could there be possible stash conflicts here that may cause complication? Or do we not expect any because the changes are scoped to `spring-cloud-previews`

emmileaf (approved):
This is so neat! Added some non-blocking comments (mostly just me trying to understand the script better), but my main curiosity from reviewing this is whether the workflow works as expected when `spring-cloud-previews` is already populated. 

Would it be straightforward to test out this setup against a branch that already contains the generated modules (e.g. #1449 or #1465?). If not, we could make it a todo to test again after merging the very first PR with these modules into main.

.github/workflows/generateAutoConfigs.yaml
emmileaf: If nothing got generated in the run (and the modules already exist on main branch), then it might still show up as file changes, just all deletions. That said, I think the `failed-library-generations` check should be catching these, @diegomarquezp  correct me if I’m wrong.

.github/workflows/generateAutoConfigs.yaml
diegomarquezp: This workflow would stash the results of the generation done in the `HEAD` of the `main` branch (assuming the workflow is triggered in the main branch), create a new branch from this `HEAD` and pop them there. Actually, this logic was intended to generate autoconfigs from a newly created PR/pushes to that PR. I think I can simplify this logic to only generate autoconfigs using the main branch (i.e. simply running `git checkout -b ... && git commit`), putting off the need for this stash method. It was useful to test generations from a branch other than `main` though (i.e. this branch), so I think we can leave it as a cleanup once we merge.

.github/workflows/generateAutoConfigs.yaml
diegomarquezp: Oh, now I see the point @emmileaf and @meltsufin. We have two cases:
1. `generate-library-list.sh` produces an empty list: depending on the branch the WF is ran on, it would either leave an empty `spring-cloud-previews` or whichever modules the source branch already had. Both cases would be catched by `Detect Changes`, which basically counts the files changed in the `spring-cloud-previews` folder (no changes would be made).
2. Every call to `generate-one.sh` fails: If any generation fails, it would be catched in the `failed-library-generations` folder. However, this made me realize that the replacement logic in `generate-one.sh` does not really replace the whole folder (it [uses](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/02267b865b146d9948b33666c72ec6e6d4f16796/spring-cloud-generator/generate-one.sh#L87) the `-o` option in `unzip`), which leaves the possibility of a new version that removes an unnecesary file, for example, but that it's not removed by the script (`unzip -o` will only replace existing files in a folder and will not delete those which are not in the `.zip`). 
That said, adding folder replacement logic to `generate-one.sh` would have a deleted folder still catched since the error detection is basically an exit code capture of `generate-one.sh`, which runs with `set -e` (i.e. the `bazel` call, folder removal and jar extraction are points of failure within the `generate-one.sh` script that are [catched](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/02267b865b146d9948b33666c72ec6e6d4f16796/spring-cloud-generator/generate-all.sh#L36) in the loop of `generate-all.sh`).

.github/workflows/generateAutoConfigs.yaml
```diff
@@ -0,0 +1,127 @@
+name: Generate Spring Auto-Configurations
+
+on:
+  workflow_dispatch:
+    inputs:
+      google_cloud_java_commitish:
```
danielgazineu, line 6: Spring Cloud GCP already depends on Libraries BOM. Can this action infer the rest of the information it needs from that instead of requiring a (potentially wrong or incompatible) google-cloud-java tag?

.github/workflows/generateAutoConfigs.yaml
meltsufin: If `generate-library-list.sh` and/or `generate-all.sh` did absolutely nothing due to a bug, would we catch that as a failure in the CI?

.github/workflows/generateAutoConfigs.yaml
diegomarquezp: Yes, we could query the pom by extracting the libraries version and using it, for example, in `https://raw.githubusercontent.com/googleapis/java-cloud-bom/libraries-bom-v26.4.0/google-cloud-bom/pom.xml -> <dependencies> -> <artifactId>gapic-libraries-bom</artifactId>`
I will add this version extraction logic to the workflow

.github/workflows/generateAutoConfigs.yaml
```diff
@@ -0,0 +1,127 @@
...
+        run: |
+          set -x
+          bash generate-library-list.sh -c "$GOOGLE_CLOUD_JAVA_COMMITISH"
+        env:
+          GOOGLE_CLOUD_JAVA_COMMITISH: ${{ inputs.google_cloud_java_commitish }}
+      - name: Compile non-autogen libraries
+        continue-on-error: false
+        run: |
+          set -e
+          mvn install -Dcheckstyle.skip -DskipTests
```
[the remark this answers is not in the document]
diegomarquezp: Modified the action to use mvnw and use the arguments of the unitTests action

.github/workflows/generateAutoConfigs.yaml
diegomarquezp: Added `google-cloud-java` tag inference logic

.github/workflows/generateAutoConfigs.yaml
diegomarquezp: I did add a sanity check script/step to confirm start and exit of `generate-all.sh`, `generate-library-list.sh` and `compute-monorepo-tag.sh` (new script).
Basically `touch`es two files at start and exit of each script. The check confirms that all these files have been created

spring-cloud-generator/generate-one.sh
```diff
@@ -106,9 +110,12 @@ add_module_to_pom () {
     echo "adding module $starter_artifactid to pom"
     sed -i "/$2/a\ \ \ \ <module>"$starter_artifactid"</module>" $1
     if [[ ${3:-1} -eq 1 ]]; then
+      # we need to remove any previously existing entries so the entry points to the
+      # right $monorepo_commitish's readme
+      sed -i "/$starter_artifactid/d" README.md
       # also write to spring-cloud-previews/README.md
       # format |name|distribution name|
-      echo -e "|$client_lib_name|com.google.cloud:$starter_artifactid|" >> README.md
+      echo -e "|$client_lib_name|com.google.cloud:$starter_artifactid|[README.md](https://github.com/googleapis/google-cloud-java/blob/$monorepo_commitish/$monorepo_folder/README.md)|" >> README.md
```
emmileaf, line 118: @diegomarquezp feel free to move this discussion to a different PR when you separate out the changes, but what do you think of something like this? 
```suggestion
      echo -e "|[$monorepo_folder](https://github.com/googleapis/google-cloud-java/blob/$monorepo_commitish/$monorepo_folder/README.md)|com.google.cloud:$starter_artifactid|" >> README.md
```
| Client Library | Starter Maven Artifact |
|-|-|
| [java-language](https://github.com/googleapis/google-cloud-java/blob/v1.2.0/java-language/README.md)| com.google.cloud:google-cloud-language-spring-starter|

emmileaf: Moving this to #1501.

emmileaf:
@diegomarquezp After syncing the changes from main, would the only expected file change in this PR be `.github/workflows/generateAutoConfigs.yaml`?

(Also dropping a side note for when you merge to update prefix to something like `chore(codegen):` - thanks!)

diegomarquezp:
@emmileaf yes, the only other altered files were some `chmod`s I did and a `set -e` I added. Both are optional changes I can include in the sanity check PR (or maybe a different one for the file modifiers). The PR now only has the GH action

emmileaf:
> @emmileaf yes, the only other altered files besides the gh script were some `chmod`s I did and a `set -e` I added. Both are optional changes I can include in the sanity check PR (or maybe a different one for the file modifiers). The PR now only has the GH action

Gotcha, thanks & LGTM!

emmileaf:
Apologies - just going to quickly edit the PR prefix before auto-squash-merging this! (Since this is more procedural and we don't intend it to be picked up as a releasable unit)

PR #1353, opened 2022-11-21
fix(codegen): remove version replacement for modules.

zhumin8:
- Remove version replacement for generated modules, these versions should be inferred from parent pom.
- set parent version to project version.
- [x] Also need to adjust relevant placeholder from gapic-generator-java side. https://github.com/googleapis/gapic-generator-java/pull/1099

Using placeholder for parent version for unblocking generation workflow setup. For future improvements:
- Use `revision` to get versions from spring-cloud-gcp parent pom. This will involve changes for all modules with `spring-cloud-gcp` as parent.

### DIFF
--- a/generated/pom.xml
+++ b/generated/pom.xml
@@ -12,22 +12,6 @@
   <artifactId>generated-parent</artifactId>
   <packaging>pom</packaging>
 
-  <dependencyManagement>
-    <dependencies>
-      <dependency>
-        <groupId>com.google.cloud</groupId>
-        <artifactId>spring-cloud-gcp-core</artifactId>
-        <version>${project.version}</version>
-      </dependency>
-
-      <dependency>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-starter</artifactId>
-        <version>${spring-boot-dependencies.version}</version>
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
-
   <modules>
   </modules>
 </project>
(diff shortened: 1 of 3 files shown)


diegomarquezp (approved):
LGTM

generated/pom.xml
```diff
@@ -12,6 +12,22 @@
   <artifactId>generated-parent</artifactId>
   <packaging>pom</packaging>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
```
emmileaf, line 18: What is the context for adding `spring-cloud-gcp-core` as a dependency here?

emmileaf (approved):

generated/pom.xml
emmileaf: Ah just realized we need it for `Credentials`,`DefaultCredentialsProvider`, etc. - got confused with something else here! Resolving.

zhumin8: Great catch! I missed the fact that we have `spring-cloud-gcp` as parent pom now. Versions for both these dependencies are already specified by BOMs in [`spring-cloud-gcp/pom`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/1c5b83dc8d039b8a71c9aa185f4bb1a328d746e9/pom.xml#L82-L96). I'll go ahead and remove the `dependencyManagement` section.

emmileaf: Makes sense!

emmileaf (approved):

PR #1355, opened 2022-11-21
test(codegen): add unit tests for one generated starter module

emmileaf:
**This PR:**
*  Adds `LanguageAutoConfigurationTests.java` (and resource files), which contains handwritten unit tests (adapted from the [POC](https://github.com/zhumin8/language_poc1/blob/shared-property/src/test/java/com/sample/autoconfig/LanguageAutoConfigTest.java)) for the corresponding generated `google-cloud-language-spring-starter` module. 
- [x] Updated file location of these tests given https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1394 updated repo structure.

**Notes for review:**
* Would appreciate feedback on any missing feature coverage! 
* I have been testing this branch against a local copy of generated `google-cloud-language-spring-starter` module, which I could also temporarily commit if it makes reviewing easier?

**Current test setup depends on the following changes in generator code:**
- CredentialsProvider updates in https://github.com/googleapis/gapic-generator-java/pull/1123
- Service settings bean configuration in https://github.com/googleapis/gapic-generator-java/pull/1110
- RetrySettings updates in https://github.com/googleapis/gapic-generator-java/pull/1143
- Properties prefix update in https://github.com/googleapis/gapic-generator-java/pull/1176


generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
```diff
@@ -0,0 +1,310 @@
...
+                  .isEqualTo(SERVICE_CREDENTIAL_CLIENT_ID);
+            });
+  }
+
+  @Test
+  void testShouldUseCredentialsBeanWhenPropertiesNotProvided() {
+    this.contextRunner.run(
+        ctx -> {
+          LanguageServiceClient client = ctx.getBean(LanguageServiceClient.class);
+          assertThat(client.getSettings().getCredentialsProvider()).isNotNull();
```
meltsufin, line 92: Wouldn't this be true even when properties were provided, like in the test above?

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
```diff
@@ -0,0 +1,310 @@
...
+        });
+  }
+
+  @Test
+  void testShouldUseDefaultGrpcTransport() {
```
meltsufin, line 97: Is there a test for the opposite, or we only allow `grpc` at this point?

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
```diff
@@ -0,0 +1,310 @@
...
+              String quotaProjectId = client.getSettings().getQuotaProjectId();
+              assertThat(quotaProjectId).isEqualTo(SERVICE_OVERRIDE_CLIENT_ID);
+            });
+  }
+
+  @Test
+  void testExecutorThreadCountFromProperties() {
+    this.contextRunner
+        .withPropertyValues(
+            "com.google.cloud.language.v1.spring.auto.language-service.executor-thread-count=3")
```
meltsufin, line 125: I would maybe define a variable for `3`.

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
```diff
@@ -0,0 +1,310 @@
...
+            .build();
+    contextRunner
+        .withBean(
+            "languageServiceSettings",
+            LanguageServiceSettings.class,
+            () -> customLanguageServiceSettings)
+        .run(
+            ctx -> {
+              LanguageServiceClient client = ctx.getBean(LanguageServiceClient.class);
+              assertThat(client.getSettings().getCredentialsProvider())
```
meltsufin, line 174: Can you just assert that `client.getSettings()` is the same as `customLanguageServiceSettings`?

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
```diff
@@ -0,0 +1,310 @@
...
+          LanguageServiceClient client = ctx.getBean(LanguageServiceClient.class);
+          assertThat(client).isNotNull();
+        });
+  }
+
+  @Test
+  void testServiceCredentialsFromProperties() {
+    this.contextRunner
+        .withPropertyValues(
+            "com.google.cloud.language.v1.spring.auto.language-service.credentials.location=file:"
```
meltsufin, line 75: This part `v1.spring.auto` seems odd. Do we have separate settings for each version? Also, `spring.auto` should not be in the properties.

meltsufin (commented):
I assume you intend these tests to be autogenerated. If that being the case, I'm curious what @blakeli0 and @burkedavison think about that because we've recently had discussions about pros and cons of auto-generating unit tests.

emmileaf:
> I assume you intend these tests to be autogenerated. If that being the case, I'm curious what @blakeli0 and @burkedavison think about that because we've recently had discussions about pros and cons of auto-generating unit tests.

@meltsufin Thank you for the review! I'll follow-up on the in-line comments individually but just wanted to quickly respond to this note - these are intended to be manually-written tests as an initial testing strategy we've decided for spring codegen, given the discussion on moving away from autogenerated unit tests as you've mentioned. 

(We’ve also considered the showcase module for unit testing, but decided to go with an example module for now given we are testing an integration layer and not for any client library functionalities that showcase provides power of coverage for.)

burkedavison:
> ... these are intended to be manually-written tests (for one example module) as an initial testing strategy we've decided for spring codegen, given the discussion on moving away from autogenerated unit tests as you've mentioned.
> 
> (We’ve also considered the showcase module for unit testing, but decided to go with an example module for now given we are testing an integration layer and not for any client library functionalities that showcase provides power of coverage for.)

I agree this is consistent with our general strategy. Handwritten unit tests should continue to be implemented whenever appropriate. Autogenerated unit tests should not.

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
zhumin8: For context: `com.google.cloud.language.v1` is the package name for `LanguageServiceClient`. We are using [package-name](.[product prefix]).[service-name] as prefix pattern as of now.

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
emmileaf: I somehow missed this, thanks for catching! Updated.

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
emmileaf: Discussion routed to design doc - will update these tests along with corresponding changes to generated code.

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
emmileaf: We are only generating for grpc client libraries in the initial list. But on a second thought, this test could have been set up in a better way. I've update it to check instead that the client library default `TransportChannelProvider` is used when no overrides are present.

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
```diff
@@ -0,0 +1,350 @@
...
+              LanguageServiceClient client = ctx.getBean(LanguageServiceClient.class);
+              Credentials credentials =
+                  client.getSettings().getCredentialsProvider().getCredentials();
+              assertThat(((ServiceAccountCredentials) credentials).getClientId())
+                  .isEqualTo(TOP_LEVEL_CREDENTIAL_CLIENT_ID);
+            });
+  }
+
+  @Test
+  void testShouldUseDefaultTransportChannelProvider() {
```
zhumin8, line 108: For the sake of coverage, you might also want to add a `testCustomTransportChannelProviderSetToRest`? 
when property `"com.google.cloud.language.v1.language-service.useRest=true"` present,
we should expect the `transportChannelProvider` set to 
```
LanguageServiceSettings.defaultHttpJsonTransportProviderBuilder().build();
```

zhumin8 (approved):
LGTM.

generated/language/src/test/java/com/google/cloud/language/v1/spring/LanguageAutoConfigurationTests.java
emmileaf: Thank you for this suggestion and added! I think it also addresses what I missed earlier in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1355#discussion_r1052341540.

PR #1356, opened 2022-11-23
fix(codegen): run formatter on generated modules

emmileaf:
**This PR:** adds `fmt-maven-plugin` in generated modules to run `google-java-format` auto-formatter as a post-processing step.

- Using `fmt-maven-plugin` version 2.9.1 with `google-java-format` at version 1.7, since they are the latest versions still [compatible with Java 8](https://github.com/spotify/fmt-maven-plugin#using-with-java-8).

**Pending:** end-to-end run depends on generated pom.xml fixes in #1353.

### DIFF
--- a/generated/pom.xml
+++ b/generated/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <artifactId>spring-cloud-gcp</artifactId>
     <groupId>com.google.cloud</groupId>
-    <version>3.5.0-SNAPSHOT</version>
+    <version>${revision}</version>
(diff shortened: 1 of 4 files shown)


zhumin8 (approved):

zhumin8:
Perhaps worth making a comment for future references with the version chosen to be compatible with JAVA 8 and why we are not adding to the `spring-cloud-gcp/pom`?

diegomarquezp (approved):

PR #1364, opened 2022-11-29
[DONOT MERGE] showing generated modules

zhumin8:
This is a draft showing the generated modules language and vision.
- `global-autoconfig-properties` module pending on #1308

From live session:

- [x] - review folder/pom structure decisions: do not leak "generated" into maven coordinates #1394

    - match folder name with artifact id by conventrion
    - consider adding to `spring-cloud-gcp-dependencies`, possibly as a separate BOM

- [x] - Add Javadoc to each properties (so Spring can pick up as hints) @emmileaf in https://github.com/googleapis/gapic-generator-java/pull/1171
- [X] - Change “@Generated by (gapic-generator-java)” to something more Spring specific https://github.com/googleapis/gapic-generator-java/pull/1103
- [x] - Revisit logging strategy in generated code @diegomarquezp 
- [x] - RetrySetting object - refactored in https://github.com/googleapis/gapic-generator-java/pull/1143 
    - Explore possibility and workload vs. current flatten setup.
    - Nice to have, readability is not the top priority of generated code


generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java
```diff
@@ -0,0 +1,471 @@
...
+@BetaApi("Autogenerated Spring autoconfiguration is not yet stable")
+@AutoConfiguration
+@ConditionalOnClass(LanguageServiceClient.class)
+@ConditionalOnProperty(
+    value = "com.google.cloud.language.v1.spring.auto.language-service.enabled",
+    matchIfMissing = false)
+@EnableConfigurationProperties({LanguageServiceSpringProperties.class, GlobalProperties.class})
+public class LanguageServiceSpringAutoConfiguration {
+  private final LanguageServiceSpringProperties clientProperties;
+  private final GlobalProperties globalProperties;
```
meltsufin, line 66: What are these global properties? I can't find this class.

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java
```diff
@@ -0,0 +1,471 @@
...
+   * default retry settings when they are not specified in LanguageServiceSpringProperties.
+   */
+  @Bean
+  @ConditionalOnMissingBean
+  public LanguageServiceClient languageServiceClient(
+      @Qualifier("languageServiceCredentials") CredentialsProvider credentialsProvider,
+      @Qualifier("defaultLanguageServiceTransportChannelProvider")
+          TransportChannelProvider defaultTransportChannelProvider)
+      throws IOException {
+    LanguageServiceSettings.Builder clientSettingsBuilder =
```
meltsufin, line 115: Why not make `LanguageServiceSettings` its own bean so that it can be overwritten, and then have a separate simple method that takes in `LanguageServiceSettings` and creates the `LanguageServiceClient`?

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java
```diff
@@ -0,0 +1,471 @@
...
+      clientSettingsBuilder.setBackgroundExecutorProvider(executorProvider);
+      LOGGER.info(
+          "Background executor thread count is " + this.clientProperties.getExecutorThreadCount());
+    }
+    if (this.clientProperties.getUseRest()) {
+      clientSettingsBuilder.setTransportChannelProvider(
+          LanguageServiceSettings.defaultHttpJsonTransportProviderBuilder().build());
+      LOGGER.info("Using HTTP transport channel");
+    }
+    RetrySettings.Builder analyzeSentimentRetrySettingBuilder =
```
meltsufin, line 141: `analyzeSentimentRetrySettings` could also be its own bean to allow customization. One thing to remember is that Spring Boot allows you to configure the app via properties, and also by overwriting beans. For the overwriting beans route, you want to make the beans as granular as possible, for maximum flexiblity.

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringProperties.java
```diff
@@ -0,0 +1,469 @@
...
+  @NestedConfigurationProperty
+  private final Credentials credentials =
+      new Credentials(
+          "https://www.googleapis.com/auth/cloud-language",
+          "https://www.googleapis.com/auth/cloud-platform");
+
+  private String quotaProjectId;
+  private Integer executorThreadCount;
+  private boolean useRest = false;
+  private Duration analyzeSentimentInitialRetryDelay;
```
meltsufin, line 42: As mentioned offline, a lot of duplication here can probably be avoided by creating an object for retry settings.

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java
blakeli0: If we are going to have a handwritten shared class, it makes more sense for us to move away from the using the folder name `generated`.

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java
blakeli0: Are you suggesting to have an RetrySetting bean for each RPC? If it has additional benefit other than readability, I think we should do it too.

meltsufin: Yes. For example, it also allows you to use the same retry settings for all RPCs.

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java
```diff
@@ -0,0 +1,471 @@
...
+ *   <li>Retries are configured for idempotent methods but not for non-idempotent methods.
+ * </ul>
+ */
+@Generated("by gapic-generator-java")
+@BetaApi("Autogenerated Spring autoconfiguration is not yet stable")
+@AutoConfiguration
+@ConditionalOnClass(LanguageServiceClient.class)
+@ConditionalOnProperty(
+    value = "com.google.cloud.language.v1.spring.auto.language-service.enabled",
+    matchIfMissing = false)
```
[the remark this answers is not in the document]
zhumin8: This is something that fall through the cracks. I'll put in a fix on the generator side.

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java
zhumin8: This is a good idea. I'll add to our todo list.

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringProperties.java
elefeint: A helper object sounds like a good idea.

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java
```diff
@@ -0,0 +1,471 @@
...
+          this.clientProperties.getClassifyTextTotalTimeout());
+      LOGGER.info(
+          "ClassifyTextTotalTimeout set to " + this.clientProperties.getClassifyTextTotalTimeout());
+    }
+    clientSettingsBuilder
+        .classifyTextSettings()
+        .setRetrySettings(classifyTextRetrySettingBuilder.build());
+    RetrySettings.Builder annotateTextRetrySettingBuilder =
+        clientSettingsBuilder.annotateTextSettings().getRetrySettings().toBuilder();
+    if (this.clientProperties.getAnnotateTextInitialRetryDelay() != null) {
```
elefeint, line 412: @meltsufin this is probably the type of code for which the duplication discussion came up. 
IMHO, there are two reasons people read code: one is to modify the code, and the other is to troubleshoot. For the first one one (code modification), there is an obvious benefit to avoiding code duplication ("keeping it DRY") -- if one piece of duplicated code is changed and the other is not, it will lead to bugs. However, this scenario does not apply to autogenerated code, per @blakeli0 's point.
The second scenario -- troubleshooting -- actually benefits from this type of code duplication, as the users can identify the exact set of properties that caused them issues. If duplication were avoided, a loop is needed, so users will then need a conditional breakpoint, which is a bit harder.

generated/language/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,10 @@
+{
+  "properties": [
+    {
+      "name": "com.google.cloud.language.v1.spring.auto.language-service.enabled",
+      "type": "java.lang.Boolean",
+      "description": "Auto-configure Google Cloud language/LanguageService components.",
+      "defaultValue": true
+    }
+  ]
```
elefeint, line 9: There are probably more options you'd want to enable IDE autocompletion for.

elefeint (approved):
LGTM, although you were not planning to merge it in :)

generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringAutoConfiguration.java
```diff
@@ -0,0 +1,471 @@
...
+    if (this.clientProperties.getExecutorThreadCount() != null) {
+      ExecutorProvider executorProvider =
+          LanguageServiceSettings.defaultExecutorProviderBuilder()
+              .setExecutorThreadCount(this.clientProperties.getExecutorThreadCount())
+              .build();
+      clientSettingsBuilder.setBackgroundExecutorProvider(executorProvider);
+      LOGGER.info(
+          "Background executor thread count is " + this.clientProperties.getExecutorThreadCount());
+    }
+    if (this.clientProperties.getUseRest()) {
```
blakeli0, line 136: I probably missed this earlier, we should use `newHttpJsonBuilder` in [here](https://github.com/googleapis/google-cloud-java/blob/c14dc312650865ba032459152cbd3901f27826ed/java-language/google-cloud-language/src/main/java/com/google/cloud/language/v1/LanguageServiceSettings.java#L158) instead of only setting the transport, `newHttpJsonBuilder` does a few other things like setting http specific headers. We can move this logic to the top of this method to something like 
```LanguageServiceSettings.Builder clientSettingsBuilder;
    if (this.clientProperties.getUseRest()) {
      clientSettingsBuilder = LanguageServiceSettings.newHttpJsonBuilder();
    } else {
      clientSettingsBuilder = LanguageServiceSettings.newBuilder();
    }
```

emmileaf: Thank you for catching this! Put together [a quick PoC](https://github.com/zhumin8/language_poc1/pull/9) to note the changes needed, and will be updating the generator code correspondingly.

emmileaf: Updated generator code in https://github.com/googleapis/gapic-generator-java/pull/1227.

zhumin8:
Please head over to #1449 for a newer version to review. Closing this one.

PR #1370, opened 2022-12-06
fix(codegen): use googleapis_folder where applicable.

zhumin8:
Fix to use `googleapis_folder` parsed from `library_list.txt` in generation process where applicable.

Moved logic to get the correct `googleapis_folder` to #1322.


generator/generate-one.sh
```diff
@@ -43,37 +45,62 @@ fi
 
 cd googleapis
 
+## If $googleapis_folder does not exist,
+# try "google/$client_lib_name/v1"
+#     "google/devtools/$client_lib_name/v1"
+#     "google/cloud/($client_lib_name without 'cloud' prefix)/v1"
+# exit if none exists
+if [ ! -d "$googleapis_folder" ]
```
diegomarquezp, line 53: We may need some logic to handle libraries that have other versions than 1. For example: `//google/devtools/clouddebugger/v2`

zhumin8: I am moving the "guess the folder name" part to #1322

emmileaf (approved):

diegomarquezp (approved):

PR #1371, opened 2022-12-06
fix: compute googleapis path heuristically

diegomarquezp:
Temporary workaround to provide valid paths to googleapis BUILD files. The actual approach is to obtain the exact path from repo-metadata.json in `google-cloud-java` (a feature yet to be added).


zhumin8:
close as #1322 is merged in.

PR #1373, opened 2022-12-07
fix(codegen): module and folder name changes

zhumin8:
Some changes to align with naming conventions:

- [x] for generated modules, keep folder name same as artifact id.
- [x] change 'generator' folder name to 'spring-cloud-generator'
- [ ] change 'generated' folder and artifact id to 'spring-cloud-generated-output'
Potential task that can be split into separate pr
- [ ] change 'spring-cloud-generated-output' from parent pom to aggregator only, use `spring-cloud-gcp/pom` as parent.


zhumin8:
close in preference of #1394

PR #1383, opened 2022-12-13
[draft] generated modules repo structure PoC

zhumin8:
Creating this draft PR to gather feedback on the repo structure with generated modules addition. Brought in `language` and `vision` module for demonstration and testing purposes. 
**Please only comment on structure and pom files.** (disregard generation scripts tailored for this test.)

As discussed, 
- Let `spring-cloud-gcp-starters` be parent for generated modules.
- Keep generated modules in a `spring-cloud-previews` folder.
- Keep an aggregator for generated modules under `spring-cloud-previews` folder.
  - This is mostly to run formatter at once (needed to fix generator results). 
  - Currently `fmt-maven-plugin` seems to be in conflict of `maven-checkstyle-plugin` in spring-cloud-gcp/pom. (this could be resolved separately. See code comment below.) 
- versions: 
  - Added `.preview` suffix to generated modules.
  - A fix for `spring-cloud-gcp-dependencies` -- any better ideas?


pom.xml
```diff
@@ -36,9 +36,11 @@
 	</distributionManagement>
 
 	<properties>
+		<project.parent.version>${project.version}</project.parent.version>
 
 		<!-- Dependency versions -->
 		<spring-cloud-dependencies.version>2021.0.5</spring-cloud-dependencies.version>
+		<spring-cloud-gcp-dependencies.version>${project.parent.version}</spring-cloud-gcp-dependencies.version>
```
zhumin8, line 43: This combination works in building either from root or submodules (spring-cloud-gcp-starters, spring-cloud-previews aggregator) but still looks kind of sketchy. Any better ideas?

spring-cloud-previews/google-cloud-vision-spring-starter/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.google.cloud</groupId>
+    <artifactId>spring-cloud-gcp-starters</artifactId>
+    <version>3.5.0-SNAPSHOT</version>
+    <relativePath>../../spring-cloud-gcp-starters/pom.xml</relativePath>
+  </parent>
+  <artifactId>google-cloud-vision-spring-starter</artifactId>
+  <version>${project.parent.version}.preview</version>
```
zhumin8, line 13: tried setting version as xx.preview only for this vision-starter module

spring-cloud-previews/pom.xml
```diff
@@ -9,10 +9,12 @@
...
-  <artifactId>generated-parent</artifactId>
+  <artifactId>spring-cloud-previews</artifactId>
   <packaging>pom</packaging>
 
   <modules>
+    <module>google-cloud-language-spring-starter</module>
+    <module>google-cloud-vision-spring-starter</module>
   </modules>
 
   <build>
```
zhumin8, line 20: In current setup, build would fail on checkstyle because `maven-checkstyle-plugin` from spring-cloud-gcp/pom runs on `validate` phase but  `fmt-maven-plugin` runs at [a later phase](https://github.com/spotify/fmt-maven-plugin#overriding-the-default-lifecycle-phase). I can think of a few options here:
- @emmileaf based on your work earlier introducing `fmt-maven-plugin`, by any chance we can use only one of the two formatter?
- Adjust running phase so `fmt-maven-plugin` (validate) runs before `maven-checkstyle-plugin`(compile). Drawback is that repo-wide checkstyle would run at a later lifecycle phase.
- Perhaps create separate “generate” profile in `spring-cloud-gcp/pom` and exclude checkstyle. So  when building with aggregator, use this profile so only `fmt-maven-plugin` is triggered.

What are your thoughts on this?

zhumin8: > On your proposal for adjusting the running phase, would it work if we configure the phase of fmt-maven-plugin so that it runs during validate?

checkstyle is also on validate, so it won't run before.

>  invoking mvnw fmt:format at the end of the generate-all.sh script is also an option.

Inspired by this, I think I have a better understanding of the issue now. `fmt-maven-plugin` was not correctly configured for each of the generated modules because it's parent are set to starter and `fmt-maven-plugin` configs in aggregator does not pass down. 
I've added the plugin to `spring-cloud-gcp/pom` with default set to skip, so it won't affect build/release for the project. Then each starter module will inherit and we are able to trigger the autoformat with `mvnw com.coveo:fmt-maven-plugin:format -Dfmt.skip=false`

emmileaf: Gotcha, I see the issue with both plugins being bound to validate now. This approach to suppress and trigger later in generated modules LGTM!

zhumin8:
proposed changes implemented.

PR #1386, opened 2022-12-13
chore: update metadata name and projectname for doc generation.

zhumin8:
Changes to docuploader metadata and java-docfx-doclet argument to stay consistent with updated names.


mpeddada1 (approved):

meltsufin (approved):
Does this actually show up in the public documentation, or is it just an internal name?

PR #1409, opened 2022-12-21
codegen: fix for bazel upgrade issue

zhumin8:
Based on `gapic-generator-java` and `googleapis` both have bazel version fixed:
- https://github.com/googleapis/gapic-generator-java/pull/1164/files
- https://github.com/googleapis/googleapis/blob/4834b4c97ed2577db072d645df393ee246328ee3/.bazeliskrc#L2

This change should be enough for our script to pickup these versions.

### DIFF
--- a/generator/generate-one.sh
+++ b/generator/generate-one.sh
@@ -68,7 +68,7 @@ perl -0777 -pi -e "s/(java_gapic_spring_library\((.*?)(\n    rest_numeric_enums 
 
 echo "CALL BAZEL TARGET"
 # call bazel target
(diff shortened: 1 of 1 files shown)


emmileaf (approved):
LGTM! 

IIUC bazelisk is able to pick up the googleapis bazel version because it’s the working directory at that point in the script (and the gapic-generator-java version fix is not used here) - so this patch should be good as long as we don't change the directory to invoke bazel build from.

zhumin8:
I am merging this in for now to make testing easier.

PR #1418, opened 2023-01-03
test(codegen): add test to verify getting the intended bean with qualifier name.

zhumin8:
This is a followup on #1355, the additional test verifies in case of multiple bean of type `TransportChannelProvider` the one with correct qualifier name is used. 
related fix in generator: https://github.com/googleapis/gapic-generator-java/pull/1207

PR #1421, opened 2023-01-04
cicd(codegen): Fix sonar ignore

zhumin8:
**Goal**: to skip sonar checks on `spring-cloud-previews` modules. Do not affect other check on other modules.
**Previous attempt**: #1416 - added to ignore path; 
**Issue**: sonar ignores any changes in `spring-cloud-previews/` folder, but sonar check is set as required to merge into main branch. So PRs with changes in `spring-cloud-previews` modules (manual unit test changes, generated code) won’t be able to merge. (e.g.#1418)
**Solution**: 
- revert `paths-ignore` changes in #1416
- add list of modules to include for `codecov` (copied from `default` profile except `spring-cloud-previews`), and alter `sonar.yaml` to deactivate `default` profile when running.

**Concern**: If new changes existing modules occurs, needs to change both modules in `default` profile and `codecov` profile.

### DIFF
--- a/.github/workflows/sonar.yaml
+++ b/.github/workflows/sonar.yaml
@@ -49,7 +49,7 @@ jobs:
             --batch-mode \
             --show-version \
             --threads 1.5C \
-            --activate-profiles codecov \
+            --activate-profiles codecov,-default \
             clean \
             install
       - name: Retry Install on Failure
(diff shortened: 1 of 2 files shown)


meltsufin:
What kind of failures does Sonar report on the auto-generated starters?

zhumin8:
@meltsufin I expect Sonar will fail on test coverage since we only have unit test for one module.

meltsufin:
Maybe we can just exclude it from code coverage calculation then. I think jacoco has setting for that.

PR #1425, opened 2023-01-04
feat: use googleapis commitish

diegomarquezp:
For each library, a hard reset is performed towards the specified commitish. The repo setup previously done in download-repos.sh was moved to generate-one.sh. The hard reset is also due to WORKSPACE and repository_rules changing between commits.

### DIFF
--- a/spring-cloud-generator/generate-one.sh
+++ b/spring-cloud-generator/generate-one.sh
@@ -7,6 +7,8 @@ set -e
 # poc with one specified repo - vision
 #cmd line:: ./generate-one.sh -c vision -v 3.1.2 -i google-cloud-vision -g com.google.cloud -p 3.5.0-SNAPSHOT -d 1
 
+#set -x
+
 # by default, do not download repos
 download_repos=0
 while getopts c:v:i:g:d:p:f: flag
@@ -17,6 +19,7 @@ do
         i) client_lib_artifactid=${OPTARG};;
         g) client_lib_groupid=${OPTARG};;
         p) parent_version=${OPTARG};;
+        x) googleapis_commitish=${OPTARG};;
         f) googleapis_folder=${OPTARG};;
         d) download_repos=1;;
     esac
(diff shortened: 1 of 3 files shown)


spring-cloud-generator/generate-one.sh
```diff
@@ -7,6 +7,8 @@ set -e
 # poc with one specified repo - vision
 #cmd line:: ./generate-one.sh -c vision -v 3.1.2 -i google-cloud-vision -g com.google.cloud -p 3.5.0-SNAPSHOT -d 1
 
+#set -x
+
```
zhumin8, line 11:
```suggestion
```

spring-cloud-generator/generate-one.sh
```diff
@@ -17,6 +19,7 @@ do
         i) client_lib_artifactid=${OPTARG};;
```
zhumin8, line 19: Cannot comment on unchanged code, but L18 `v) version=${OPTARG};;` and it's usage(shown below) are probably not needed anymore. Can you double check and remove them?
```
sed -i 's/{{client-library-version}}/'"$version"'/' "$starter_artifactid"/pom.xml
```

spring-cloud-generator/generate-one.sh
```diff
@@ -44,6 +48,20 @@ if [[ $download_repos -eq 1 ]]; then
 fi
 
 cd googleapis
+git reset --hard $googleapis_commitish
+
+# In googleapis/WORKSPACE, find http_archive() rule with name = "gapic_generator_java",
```
zhumin8, line 53: Do you think it makes sense for readability to extract these replacement logic into a separate script?

zhumin8 (changes requested):

spring-cloud-generator/generate-one.sh
diegomarquezp: Sure, will use a separate file

spring-cloud-generator/generate-one.sh
diegomarquezp: Fixed in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1430

zhumin8 (approved):
LGTM

PR #1427, opened 2023-01-04
fix: use `rule_prefix` instead of `client_lib_name` in `generate-one.sh`

diegomarquezp:
Using client_lib_name will lead to wrong rule names. For example: `googleapis/google/X/Y/v1.BUILD.bazel` could have client_lib_name XY, but the build rules would have a prefix of the form `Y_java_gapic_spring` instead of `XY_java_gapic_spring` This computes the right prefix by checking the BUILD file and taking an already existing rule as a template for the correct rule name.

BEGIN_COMMIT_OVERRIDE
chore(codegen): use `rule_prefix` instead of `client_lib_name` in `generate-one.sh`
END_COMMIT_OVERRIDE


zhumin8 (approved):
LGTM

emmileaf (approved):

PR #1428, opened 2023-01-04
feat: specify bazel version in googleapis

diegomarquezp:
Uses a `.bazelversion` file as explained in [bazelisk docs](https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run)

### DIFF
--- a/spring-cloud-generator/generate-one.sh
+++ b/spring-cloud-generator/generate-one.sh
@@ -44,6 +44,8 @@ if [[ $download_repos -eq 1 ]]; then
 fi
 
 cd googleapis
+# tell bazelisk to use bazel version 4.2.2
+echo '4.2.2' > .bazelversion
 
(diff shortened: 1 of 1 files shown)


zhumin8:
the USE_BAZEL_VERSION variable is already set in `.bazeliskrc` file within googleapis. And takes priority over your setup here based on  [bazelisk docs](https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run).
Is this change needed? Or is your intent to have a fall back version here?

ref: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1409#pullrequestreview-1226714632

diegomarquezp:
> the USE_BAZEL_VERSION variable is already set in .bazeliskrc file within googleapis

Indeed, thanks!

PR #1430, opened 2023-01-04
fix: remove unnecessary `$version` variable in `generate-*.sh`

diegomarquezp:


zhumin8 (approved):
Leaving a note for future ref: `{{client-library-version}}` is removed in  https://github.com/googleapis/gapic-generator-java/pull/1141

PR #1431, opened 2023-01-04
chore(codegen): Fix sonar to exclude in jacoco

zhumin8:
Closed #1421 in favor of this change:
- ~~Reverts #1416 ignore path change;~~ Done in #1447
- Instead add file pattern exclusions to sonar properties. Refer to "File exclusion and inclusion" section in sonarcloud guide [here](https://docs.sonarcloud.io/advanced-setup/analysis-scope/). This setup is tested in #1446.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -458,6 +458,13 @@
 						<groupId>org.jacoco</groupId>
 						<artifactId>jacoco-maven-plugin</artifactId>
 						<version>0.8.8</version>
+						<configuration>
+							<excludes>
+								<exclude>
+									<exclude>spring-cloud-previews/**</exclude>
+								</exclude>
+							</excludes>
+						</configuration>
 						<executions>
 							<execution>
 								<goals>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -78,7 +78,8 @@
 		<!-- Other defaults -->
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<integration-test.pattern>**/*IntegrationTest*</integration-test.pattern>
-
+		<!-- exclude preview module code from sonar cloud analysis -->
+		<sonar.exclusions>**/spring/*SpringAutoConfiguration.*, **/spring/*SpringProperties.*</sonar.exclusions>
```
emmileaf, line 82: Naive question: is there a reason for matching on the two classes rather than something more broad like `spring-cloud-previews/**`?

emmileaf (approved):

pom.xml
meltsufin: +1 I think if we can get `spring-cloud-previews` in the pattern, it would be a more specific match.

zhumin8: Good point. updated to use folder for exclusion.
This was a miss on my end: jacoco exclusions does not accept folders well, and forgot to change back when moving to sonar.exclusions.

PR #1437, opened 2023-01-07
feat: add error handling for `generate-all.sh`

diegomarquezp:
With this PR, if one of the libraries generated by `generate-one.sh` fails, the now captured std* output will be stored in a file under `failed-library-generations` named the same as the client library (i.e. one file will be generated with std* of each failed lib). `geneate-all.sh` continues normally after the failed output is saved.


spring-cloud-generator/generate-all.sh
```diff
@@ -22,8 +30,20 @@ while IFS=, read -r library_name googleapis_location coordinates_version googlea
...
+  bash $WORKING_DIR/generate-one.sh \
+    -c $library_name \
+    -i $artifact_id \
+    -g $group_id \
+    -p $PROJECT_VERSION \
+    -f $googleapis_location \
+    -x $googleapis_commitish 2>&1 | tee tmp-generate-one-output || save_error_info $library_name
+  set +o pipefail
 done <<< $libraries
+rm tmp-generate-one-output
```
emmileaf, line 46: I think this rm is redundant (and fails) because you already use `mv tmp-generate-one-output` instead of `cp tmp-generate-one-output` above
```suggestion
```

spring-cloud-generator/generate-all.sh
```diff
@@ -13,6 +13,14 @@ cd ../
 PROJECT_VERSION=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout)
 cd spring-cloud-generator
 
+# When generate-one.sh fails, stores the captured stdout and stderr to a file
+# with the name of the client library
+# args 1 - library name;
+save_error_info () {
+  mkdir -p failed-library-generations
+  mv tmp-generate-one-output failed-library-generations/$1
```
emmileaf, line 21: Nit: perhaps we could rename this folder to something like `preview-module-errors`?

emmileaf (approved):
Thank you - this change is super helpful for troubleshooting!

spring-cloud-generator/generate-all.sh
zhumin8: I think either name is fine, or something along the lines of `generation-error-logs`, as long as we don't intend to merge in the changes to main. It would be weird to have a failed/errors in our repo's main.
I imagine the process goes something like this:
- trigger generation from GithubAction
- Check on generated PR, look for anything in this folder. 
- In case there are errors, we probably don't want to merge in the pr. Will try to troubleshot or exclude and regenerate.
- If no errors occurred, this folder will not be populated to the PR.

emmileaf:
@diegomarquezp bringing in the latest changes from main should unblock the releaseCheck issue here - thanks!

spring-cloud-generator/generate-all.sh
diegomarquezp: The action checks if any files are in `preview-module-errors`

diegomarquezp: I will modify the folder name to a more understandable one once #1341 is merged

PR #1424, opened 2023-01-04
feat: adapt to new library_list.txt

diegomarquezp:
* passes googleapis commitish to `generate-one.sh`
* has optional function for local development

PR #1438, opened 2023-01-07
chore: remove unnecessary optarg handling in `generate-library-list.sh`

diegomarquezp:
Initially brought by #1424, intended to handle a dev-env workflow that was later [deleted](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1424/commits/0da05fa43f104c7d74a751dc4c4d08cddadce494) (do `mvn install` in root). This is now unnecessary.

### DIFF
--- a/spring-cloud-generator/generate-all.sh
+++ b/spring-cloud-generator/generate-all.sh
@@ -1,13 +1,6 @@
 #!/bin/bash
 WORKING_DIR=`pwd`
 
-while getopts d: flag
-do
-    case "${flag}" in
-        d) dev_env=1;;
-    esac
-done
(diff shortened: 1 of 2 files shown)


zhumin8 (commented):
hmm. I think this is the commit hash or tag from monorepo to start search on and still needed.
Not related to the deleted code you are pointing to.

diegomarquezp:
@zhumin8 , I incorrectly removed the optargs from `generate-library-list.sh`. I just pushed a deletion on the right file (`generate-all.sh`)

zhumin8 (approved):

PR #1449, opened 2023-01-11
[DONOT MERGE] showing generated modules

zhumin8:
Before #1341 is ready, this is a pr generated from local containing all generated modules for review.

PR #1455, opened 2023-01-12
chore(codegen): add checkstyle suppressions for preview modules

emmileaf:
This PR defines more specific checkstyle suppressions for generated modules in `spring-cloud-previews` adapted from #1449's commit [9036f7d](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1449/commits/9036f7d80e678d3dce2367db52166e3fa4933c3c). It looks like `spring-cloud-gcp` enforces stricter checkstyle rules ([fail on warning](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/a5a346600e4b123643d9a507144d9167280c8870/pom.xml#L273-L274)) than what is used for the client libraries, and a number of abbreviation to lowerCamelCase edge cases are getting caught here.

- `AbbreviationAsWordInName` suppressions - for abbreviations in service names that appear in generated class names (e.g. `IDSSpringProperties`) , as well as abbreviations in rpc method names (affecting retry-related generated code) such as the [getOSPolicyAssignment method](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/a4a78760ea619a804102e2de017c961c069100c8/spring-cloud-previews/google-cloud-os-config-spring-starter/src/main/java/com/google/cloud/osconfig/v1/spring/OsConfigZonalServiceSpringAutoConfiguration.java#L163)

- `MethodName` and `ParameterName` suppressions - for abbreviations in service names, from lines such as this [iAMClient bean method](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/a4a78760ea619a804102e2de017c961c069100c8/spring-cloud-previews/google-iam-admin-spring-starter/src/main/java/com/google/cloud/iam/admin/v1/spring/IAMSpringAutoConfiguration.java#L761)

List of affected starter modules:
- google-cloud-os-config-spring-starter
- google-cloud-ids-spring-starter
- google-cloud-container-spring-starter
- google-cloud-iamcredentials-spring-starter
- google-iam-admin-spring-starter
- google-cloud-gsuite-addons-spring-starter

Related PR: https://github.com/googleapis/gapic-generator-java/pull/1245

### DIFF
--- a/src/checkstyle/checkstyle-suppressions.xml
+++ b/src/checkstyle/checkstyle-suppressions.xml
@@ -34,4 +34,16 @@
     <suppress checks="AbbreviationAsWordInName" files="/spring-cloud-gcp-samples/"/>
     <suppress checks="OneTopLevelClass" files="/spring-cloud-gcp-samples/"/>
 
+    <!--  Temporary suppress for spring-cloud-previews modules,
+      for lowerCamelCase abbreviation handling limitations in generated code -->
+    <suppress checks="AbbreviationAsWordInName" files="/spring-cloud-previews/"/>
+    <suppress checks="[MethodName|ParameterName]"
+              files="/spring-cloud-previews/google-cloud-ids-spring-starter/.*IDSSpringAutoConfiguration.java"/>
+    <suppress checks="[MethodName|ParameterName]"
+              files="/spring-cloud-previews/google-cloud-iamcredentials-spring-starter/.*IAMCredentialsSpringAutoConfiguration.java"/>
+    <suppress checks="[MethodName|ParameterName]"
+              files="/spring-cloud-previews/google-iam-admin-spring-starter/.*IAMSpringAutoConfiguration.java"/>
+    <suppress checks="[MethodName|ParameterName]"
+              files="/spring-cloud-previews/google-cloud-gsuite-addons-spring-starter/.*GSuiteAddOnsSpringAutoConfiguration.java"/>
+
 </suppressions>
--- a/src/checkstyle/checkstyle-suppressions.xml
+++ b/src/checkstyle/checkstyle-suppressions.xml
@@ -36,7 +36,22 @@
 
     <!--  Temporary suppress for spring-cloud-previews modules,
       for lowerCamelCase abbreviation handling limitations in generated code -->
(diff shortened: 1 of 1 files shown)


zhumin8 (approved):
Does it make sense to do `AbbreviationAsWordInName` suppression per module, or do we have many modules affected that blanket suppression works better?

emmileaf:
@zhumin8 That also makes sense, I was also debating between the two. Added in a new commit and updated the PR description to reflect this change.

src/checkstyle/checkstyle-suppressions.xml
```diff
@@ -34,4 +34,31 @@
     <suppress checks="AbbreviationAsWordInName" files="/spring-cloud-gcp-samples/"/>
     <suppress checks="OneTopLevelClass" files="/spring-cloud-gcp-samples/"/>
 
+    <!--  Temporary suppress for spring-cloud-previews modules,
+      for lowerCamelCase abbreviation handling limitations in generated code -->
+    <suppress checks="AbbreviationAsWordInName"
```
meltsufin, line 39: Since we have such a specific check, why not apply it to the entire `/spring-cloud-previews` directory to make it less verbose?

emmileaf: @meltsufin Ah, @zhumin8 and I were just having this discussion above. I think either way works for unblocking the current checkstyle failures, the per-file approach ([e3d2180](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1455/commits/e3d2180330189b27bfb66446d22e3310b8e5c035)) gives better visibility of which files are affected but the blanket approach ([ca8506e](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1455/commits/ca8506e8e056ab5e5c804cf325e8f81a84ee5499)) makes the file cleaner. Do you have a strong preference here on which one to apply?

meltsufin: No strong preference. This is fine.

meltsufin (approved):

diegomarquezp (approved):

PR #1480, opened 2023-01-19
chore: Update java version in github workflow

JoeWang1127:
- Set Java version to Java 17 in codeql-analysis.yaml
- Set Java version to Java 17 in updateDocs.yaml

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -56,7 +56,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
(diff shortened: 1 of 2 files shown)


meltsufin (approved):

emmileaf:
@JoeWang1127 would `chore:` or `ci` be a better prefix here? (I think `feat:` will get picked up by release-please as a unit to show up in release notes)

JoeWang1127:
> @JoeWang1127 would `chore:` or `ci` be a better prefix here? (I think `feat:` will get picked up by release-please as a unit to show up in release notes)

Good suggestion, I'll change the title.

PR #1489, opened 2023-01-20
chore(codegen): additional replacement logic in googleapis workspace file

emmileaf:
This PR unblocks generation script [errors](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/3953287474) when pointing to more recent googleapis committish. This is due to a change in `googleapis/WORKSPACE` ([d7e8de7](https://github.com/googleapis/googleapis/commit/d7e8de775ab827293821077c6c198b32a51260ff)) that modified this file to maven install the gapic-generator-java jar. The spring codegen workflow still relies on editing this to use a locally installed branch of `gapic_generator_java`. 
- This patch adds logic to remove the  `maven_install()` block and add back `gapic_generator_java_repositories()`. 

Note that the longer-term to address this to migrating the spring generator code out of the `gapic-generator-java` repo, and switching to using the `gapic-generator-java` jar.

### DIFF
--- a/spring-cloud-generator/googleapis-gapic-string.txt
+++ b/spring-cloud-generator/googleapis-gapic-string.txt
@@ -1,0 +1,3 @@
+load("\@gapic_generator_java//:repositories.bzl", "gapic_generator_java_repositories")
+
+gapic_generator_java_repositories()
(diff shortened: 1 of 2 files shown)


diegomarquezp (approved):
Thanks!

meltsufin (approved):
Do we have a tracking bug for switching to use the jar?

emmileaf:
Currently tracking the migration and switch as part of the project planning document.

PR #1498, opened 2023-01-20
chore(codegen): sanity checks for generation scripts

diegomarquezp:
A sanity check script that checks for:
* `library_list.txt` not being empty
* `library_list.txt` not having empty cells
* Confirm `library_list.txt` and generated folders have the same length
* for each starter, check there is
  * At least one `*AutoConfiguration.java` file
  * At least one `*Properties.java` file
  * A `pom.xml` file
  * Corresponding files in `src/main/resources/`

This is complimentary to already existing checks:
* [Generation errors reported by bazel](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/b444fcbbdd57984ac7304263490546219ec41c86/.github/workflows/generateAutoConfigs.yaml#L69-L80)
* [Compilation errors in the generated libraries](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/b444fcbbdd57984ac7304263490546219ec41c86/.github/workflows/generateAutoConfigs.yaml#L81-L91)
* [Check that there were changes in the starters before creating the PR](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/b444fcbbdd57984ac7304263490546219ec41c86/.github/workflows/generateAutoConfigs.yaml#L93-L107)


spring-cloud-generator/generate-all.sh
```diff
@@ -18,6 +20,7 @@ save_error_info () {
 # repos are downloaded once before all generation jobs and then removed
 
 bash download-repos.sh
+touch run-sanity-check/generate-all-started
```
meltsufin, line 23: Instead of creating these marker files by using `touch`, can we just rely on actual files with content that are generated as part of the process?

spring-cloud-generator/generate-all.sh
```diff
@@ -44,3 +47,4 @@ cd ../spring-cloud-previews
 cd ../spring-cloud-generator
 rm -rf googleapis
 rm -rf gapic-generator-java
+touch run-sanity-check/generate-all-finished
```
meltsufin, line 50: I don't think this is necessary.

spring-cloud-generator/generate-library-list.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+mkdir -p run-sanity-check
+touch run-sanity-check/generate-library-list-started
```
meltsufin, line 3: I think the library list being created is a sufficient side effect that we can check.

spring-cloud-generator/generate-library-list.sh
```diff
@@ -91,3 +93,4 @@ echo "Total in-scope client libraries: $count"
 
 # clean up
 rm -rf google-cloud-java/
+touch run-sanity-check/generate-library-list-finished
```
meltsufin, line 96: Same here. I think just check that `library_list.txt` was created and has some valid content is sufficient.

spring-cloud-generator/generate-all.sh
```diff
@@ -18,6 +20,7 @@ save_error_info () {
 # repos are downloaded once before all generation jobs and then removed
 
 bash download-repos.sh
+touch run-sanity-check/generate-all-started
 libraries=$(cat $WORKING_DIR/library_list.txt | tail -n+2)
 while IFS=, read -r library_name googleapis_location coordinates_version googleapis_commitish; do
   echo "processing library $library_name"
```
meltsufin, line 26: Maybe also write this to a file so that you can later use it in the sanity check?
Can we also tell for each library if there were any changes and log that?

meltsufin (changes requested):
The main thing I would suggest is to not use `touch` and instead use the produced files in sanity checks.

spring-cloud-generator/generate-all.sh
diegomarquezp: Removed - added check contents of library list and generated autoconfs

spring-cloud-generator/generate-all.sh
diegomarquezp: Removed

spring-cloud-generator/generate-library-list.sh
diegomarquezp: Added library list checks in `sanity-checks.sh`

spring-cloud-generator/library_list.txt
```diff
@@ -1,90 +1,88 @@
...
-transcoder, google/cloud/video/transcoder/v1, com.google.cloud:google-cloud-video-transcoder, 758f0d1217d9c7fe398aa5efb1057ce4b6409e55
-vmmigration, google/cloud/vmmigration/v1, com.google.cloud:google-cloud-vmmigration, 758f0d1217d9c7fe398aa5efb1057ce4b6409e55
-vpcaccess, google/cloud/vpcaccess/v1, com.google.cloud:google-cloud-vpcaccess, 758f0d1217d9c7fe398aa5efb1057ce4b6409e55
-webrisk, google/cloud/webrisk/v1, com.google.cloud:google-cloud-webrisk, 758f0d1217d9c7fe398aa5efb1057ce4b6409e55
-websecurityscanner, google/cloud/websecurityscanner/v1, com.google.cloud:google-cloud-websecurityscanner, 758f0d1217d9c7fe398aa5efb1057ce4b6409e55
-workflowexecutions, google/cloud/workflows/executions/v1, com.google.cloud:google-cloud-workflow-executions, 807125e7953e32535710924fe4bf0362ac93754d
-workflows, google/cloud/workflows/v1, com.google.cloud:google-cloud-workflows, 758f0d1217d9c7fe398aa5efb1057ce4b6409e55
+# api_shortname, googleapis-folder, distribution_name:version, googleapis_committish, monorepo_folder
+accessapproval, google/cloud/accessapproval/v1, com.google.cloud:google-cloud-accessapproval, 758f0d1217d9c7fe398aa5efb1057ce4b6409e55, java-accessapproval
```
meltsufin, line 2: Did you re-run the generation script for this? Is that why it got updated?

spring-cloud-generator/sanity-checks.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+WORKING_DIR=`pwd`
```
meltsufin, line 2: Please explain in the comments at the top of the script how to use it. Which scripts is it testing?

meltsufin (changes requested):
The checks look good to me. Very comprehensive. However, where are we calling this script from? Is it supposed to be simply validating the current state of the repository, or the execution of the generator in a clean directory?

spring-cloud-generator/scripts/resources/googleapis_modification_string.txt
```diff
@@ -1,4 +1,4 @@
-_spring_cloud_generator_version = "4.2.1-SNAPSHOT" # {x-version-update:spring-cloud-gcp:current}
+_spring_cloud_generator_version = "4.3.1-SNAPSHOT" # {x-version-update:spring-cloud-gcp:current}
```
emmileaf, line 1: Ah, this version didn't get updated with the annotation like the pom file did @emmileaf Note to self to follow-upon this (https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files)

spring-cloud-generator/sanity-checks.sh
diegomarquezp: Added overview of the script

spring-cloud-generator/library_list.txt
diegomarquezp: Unintended and removed :)

emmileaf (approved):
Thanks!

spring-cloud-generator/scripts/sanity-checks.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+# This script performs the following sanity checks for the library generation
```
diegomarquezp, line 2: @meltsufin I'm leaving a couple notes to indicate the reach of these checks.

We have the empty library list [check](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/07a3a1c6f8cbf66c9f488d447c75cec435a941ab/spring-cloud-generator/scripts/sanity-checks.sh#L22) to prevent the case where an empty list still gives a successful run of the generator. 

Then we have the [check](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/07a3a1c6f8cbf66c9f488d447c75cec435a941ab/spring-cloud-generator/scripts/sanity-checks.sh#L53) to see if the generated files are present in our previews folder (checks pom, classes, etc are present)

Since we [start with an empty previews folder](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/07a3a1c6f8cbf66c9f488d447c75cec435a941ab/.github/workflows/generateAutoConfigs.yaml#L69-L74), confirming the library list is not empty and checking that the generated source files are present **we can confirm the generator was run**. In other words, these two checks help confirm at least one library was successfully generated and written in the previews folder.

meltsufin: Thanks! Perfect!

meltsufin (approved):

PR #1499, opened 2023-01-20
chore(codegen): compute monorepo commitish (tag) from libraries bom

diegomarquezp:


spring-cloud-generator/generate-library-list.sh
```diff
@@ -1,13 +1,7 @@
...
-#cmd line:: ./generate-library-list.sh -c v1.1.0
-
-while getopts c: flag
-do
-    case "${flag}" in
-        c) commitish=${OPTARG};;
-    esac
-done
+#cmd line:: ./generate-library-list.sh
+commitish=$(bash compute-monorepo-tag.sh)
```
emmileaf, line 4: Would we need something like `v1.2.0` instead of just the parsed`1.2.0` here?
```suggestion
commitish="v$(bash compute-monorepo-tag.sh)"
```

meltsufin (commented):
LGTM, +1 @emmileaf's comment.

meltsufin (approved):

emmileaf (approved):

PR #1505, opened 2023-01-23
chore: update dependabot in the `3.x` branch

JoeWang1127:


.github/dependabot.yml
```diff
@@ -1,10 +1,28 @@
...
+
+# rules for the `3.x` branch
+- package-ecosystem: maven
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+  target-branch: "3.x"
   labels:
     - "dependencies"
```
emmileaf, line 27: Maybe it'd be helpful to differentiate the label, similar to when we used to have `2.x dependencies`?
```suggestion
    - "3.x dependencies"
```

emmileaf (approved):
Small suggestion, otherwise LGTM!

.github/dependabot.yml
JoeWang1127: Thanks for the suggestion, accepted.

meltsufin (approved):

PR #1469, opened 2023-01-19
chore(codegen): update documentation with link to additional starters

emmileaf:
This PR reverts #1369 - prepare to merge this when ready to release.
Merge order: Initial PR with preview modules (#1517 > this PR > #1415.

PR #1525, opened 2023-01-25
chore(codegen): cleanup step for GH action

diegomarquezp:
Adds a script that wipes out all the contents of `spring-cloud-previews` and replaces them with the `spring-cloud-previews-template`. The GH action should be consider only as the step in line 24, as the rest of it will be brought up by https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1341


emmileaf:
Thanks @diegomarquezp! Two quick questions/asks:
- Is the intention to remove the `.github/workflows/generateAutoConfigs.yaml` before merging, since that is WIP in #1341? 
- Apologies for the extra work here, but could you try copying the unit test file again from main, since there’s a small change from #1518 that probably got missed?

Note to self after this PR is merged: update the template README in (or after) #1469 to keep it in sync.

meltsufin (approved):
Let's try to get all of the various GH action-related PRs in and then do the necessary fixes. It's getting a bit hard to keep track when there are multiple PRs open that touch the same code.

diegomarquezp:
@emmileaf I will remove the `generateAutoConfigs`. I thought we may want to include it since it also includes the cleanup step, but we may want to remove it.

diegomarquezp:
I added the new unit test changes @emmileaf

emmileaf:
> @emmileaf I will remove the `generateAutoConfigs`. I thought we may want to include it since it also includes the cleanup step, but we may want to remove it.

Apologies for the confusion, I was just unsure if this PR is for adding the github action as well and superseding #1341, or just the scripts/setup needed to enable the github action. Please help clarify if I'm misunderstanding, otherwise the changes LGTM!

emmileaf (approved):

diegomarquezp:
@emmileaf It's not superseeding #1341. The commit that updates the GH action is also in #1341, so the merge order does not matter. We can still drop out `generateAutoConfigs` from this PR and expect it to be added in #1341

PR #1529, opened 2023-01-26
chore(codegen): fix script issues with readme and residual files

emmileaf:
This PR fixes [two items](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1528#issue-1558203562) that needed to be manually patched in the latest generated PR (#1528):

- Residual file from earlier sanity check WIP (#1498)
- Two places where updating README breaks given the current pattern matching logic:
  * Removing entries by matching on `$starter_artifactid` alone also removes the line that contains it in the sample dependency snippet above
  * Sorting entries pattern-matched on lines containing `:`, which caused issue when lines containing `:` were added in the descriptive section of the readme. Replaced this punctuation as a quick fix, but will need to look into a replacement for the regex going forward.
  
Added context for each change in in-line comments below.

### DIFF
--- a/spring-cloud-generator/generate-one.sh
+++ b/spring-cloud-generator/generate-one.sh
@@ -112,7 +112,7 @@ add_module_to_pom () {
     if [[ ${3:-1} -eq 1 ]]; then
       # we need to remove any previously existing entries so the entry points to the
       # right $monorepo_commitish's readme
-      sed -i "/$starter_artifactid/d" README.md
+      sed -i "/com.google.cloud:$starter_artifactid/d" README.md
       # also write to spring-cloud-previews/README.md
       # format |name|distribution name|
       echo -e "|[$monorepo_folder](https://github.com/googleapis/google-cloud-java/blob/$monorepo_commitish/$monorepo_folder/README.md)|com.google.cloud:$starter_artifactid|" >> README.md
--- a/spring-cloud-generator/generate-one.sh
+++ b/spring-cloud-generator/generate-one.sh
@@ -110,9 +110,6 @@ add_module_to_pom () {
     echo "adding module $starter_artifactid to pom"
     sed -i "/$2/a\ \ \ \ <module>"$starter_artifactid"</module>" $1
     if [[ ${3:-1} -eq 1 ]]; then
-      # we need to remove any previously existing entries so the entry points to the
-      # right $monorepo_commitish's readme
-      sed -i "/com.google.cloud:$starter_artifactid/d" README.md
       # also write to spring-cloud-previews/README.md
       # format |name|distribution name|
       echo -e "|[$monorepo_folder](https://github.com/googleapis/google-cloud-java/blob/$monorepo_commitish/$monorepo_folder/README.md)|com.google.cloud:$starter_artifactid|" >> README.md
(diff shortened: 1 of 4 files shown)


spring-cloud-generator/generate-one.sh
```diff
@@ -112,7 +112,7 @@ add_module_to_pom () {
     if [[ ${3:-1} -eq 1 ]]; then
       # we need to remove any previously existing entries so the entry points to the
       # right $monorepo_commitish's readme
-      sed -i "/$starter_artifactid/d" README.md
+      sed -i "/com.google.cloud:$starter_artifactid/d" README.md
```
emmileaf, line 115: Removing entries by matching on `$starter_artifactid` alone also removes an unrelated line that contains an example in the dependency xml snippet above.

spring-cloud-previews/README.md
```diff
@@ -4,7 +4,7 @@ Please see below for a list of additional preview starters provided, with depend
 
 These starters are not included in the Spring Framework on Google Cloud Bill of Materials (BOM), 
 and should be explicitly added as dependencies. For example, to use the starter for Cloud Natural Language, 
-add the following dependency to your `pom.xml`:
+add the following dependency to your `pom.xml`.
```
emmileaf, line 7: Sorting entries in the table below pattern-matched on lines containing `:`, which caused issue and included this line of the readme into the sort. Replaced this punctuation as a quick fix, but will need to look into a replacement for the regex going forward.

spring-cloud-generator/generate-one.sh
diegomarquezp: @emmileaf, I realized that since we are now [using the empty template](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/b256dc5e0bbdc9694ee3a2a19c2b5d91b30361ef/spring-cloud-generator/reset-previews-folder.sh#L7-L9) with an empty readme, we can maybe just remove this line

spring-cloud-previews/README.md
diegomarquezp: If we end up removing that conflicting `sed` line, we may just come back to use a colon here

spring-cloud-generator/generate-one.sh
emmileaf: @diegomarquezp Ah you're right, thanks for the suggestion! I'll quickly test and remove this line.

spring-cloud-generator/generate-one.sh
```diff
@@ -110,9 +110,6 @@ add_module_to_pom () {
     echo "adding module $starter_artifactid to pom"
     sed -i "/$2/a\ \ \ \ <module>"$starter_artifactid"</module>" $1
     if [[ ${3:-1} -eq 1 ]]; then
-      # we need to remove any previously existing entries so the entry points to the
-      # right $monorepo_commitish's readme
-      sed -i "/$starter_artifactid/d" README.md
```
emmileaf, line 115: Removed since we are now using the [template](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/b256dc5e0bbdc9694ee3a2a19c2b5d91b30361ef/spring-cloud-generator/reset-previews-folder.sh#L7-L9) with an empty readme, per https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1529#discussion_r1088223388

meltsufin (approved):

diegomarquezp (approved):

PR #1541, opened 2023-01-27
chore: Using trampoline strategy for releases

ddixit14:


.kokoro/trampoline.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+# Copyright 2018 Google Inc.
```
JoeWang1127, line 2: Do we need to update the year in Copyright?

.kokoro/stage.sh
```diff
@@ -28,17 +28,8 @@ MAVEN_SETTINGS_FILE=$(realpath .)/settings.xml
 setup_environment_secrets
 create_settings_xml_file $MAVEN_SETTINGS_FILE
 
-# run unit tests
-./mvnw verify --show-version --batch-mode
```
meltsufin, line 32: Why remove the testing?

meltsufin (changes requested):
I just realized we already use trampoline in google-cloud-java. It looks like this is matching that config.

meltsufin (approved):

.kokoro/stage.sh
ddixit14: Added it.

PR #1547, opened 2023-01-30
chore(codegen): update auto-generated commit and pr title by gh action

emmileaf:
A small change to update the commit message and PR title generated by the codegen github action, since future PRs will be updating the modules existing in main. 
- Defaulting prefix to chore, reviewer of the generated PR can override if needed.

Note: this PR is not a blocker for the 4.0.0 release.

### DIFF
--- a/.github/workflows/generateAutoConfigs.yaml
+++ b/.github/workflows/generateAutoConfigs.yaml
@@ -110,7 +110,7 @@ jobs:
         run: |
           set -x
           # stashes the source branch with generated libs 
-          MESSAGE=`echo "chore(codegen): update starter modules in spring-cloud-preview"`
+          MESSAGE=`echo "chore: update starter modules in spring-cloud-preview"`
           git stash push -- spring-cloud-previews/
           git reset --hard
           git remote update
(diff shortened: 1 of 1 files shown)


.github/workflows/generateAutoConfigs.yaml
```diff
@@ -110,7 +110,7 @@ jobs:
         run: |
           set -x
           # stashes the source branch with generated libs 
-          MESSAGE=`echo "feat(generated): Generate new autoconfig libraries"`
+          MESSAGE=`echo "chore(codegen): update starter modules in spring-cloud-preview"`
```
ddixit14, line 113: Can you just keep it as "chore" instead of "chore(codegen):" ? You can probably add the codegen keyword after the colon.

emmileaf: For sure - updated.

diegomarquezp (approved):

meltsufin (approved):
This does make me wonder if we can make the generation script smarter in reporting what changed. For instance, what if a new library is added? Our release notes won't have a mention of it, unless we update the commit manually. Maybe the commit description can list libraries that were updated and created.

cc/ @diegomarquezp

emmileaf:
> Maybe the commit description can list libraries that were updated and created.

Hmm that's a good idea to look into! Noted as automation enhancement.

PR #1553, opened 2023-01-31
chore: update publish_javadoc.sh

ddixit14:
Current `publish_javadoc.sh` was complaining about pyenv 3.7.2 not installed. 

image

Default pyenv version on the image is 2.3.9

image

The changes here include installing pyenv 3.7.2 and using it. 

I've also skipped the remove snapshot step that we don't need anymore - 
`./mvnw versions:set --batch-mode -DremoveSnapshot -DprocessAllModules`

### DIFF
--- a/.kokoro/publish_javadoc.cfg
+++ b/.kokoro/publish_javadoc.cfg
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
+  value: "gcr.io/cloud-devrel-kokoro-resources/java11"
 }
 
 env_vars: {
--- a/.kokoro/publish_javadoc.sh
+++ b/.kokoro/publish_javadoc.sh
@@ -5,7 +5,8 @@ if [[ -z "${CREDENTIALS}" ]]; then
   CREDENTIALS=${KOKORO_KEYSTORE_DIR}/73713_docuploader_service_account
 fi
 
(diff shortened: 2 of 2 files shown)


.kokoro/publish_javadoc.sh
```diff
@@ -5,15 +5,13 @@ if [[ -z "${CREDENTIALS}" ]]; then
   CREDENTIALS=${KOKORO_KEYSTORE_DIR}/73713_docuploader_service_account
 fi
 
+pyenv install 3.7.2
```
meltsufin, line 8: Should we just have it in the image we're using? Do we do it in other repos the same?

ddixit14: Having it default in the image is the best way. Other repos aren't setting the pyenv versions. Should I just try running it without setting the version?

.kokoro/publish_javadoc.cfg
```diff
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
+  value: "gcr.io/cloud-devrel-kokoro-resources/java11"
```
suztomo, line 6: Java 17 is needed, right?

ddixit14: not for javadocs I believe.

suztomo: OK. No harm trying.

suztomo (approved):

Issue #1551, opened 2023-01-31
SecretManagerTemplate bean no longer auto-configured by default in 4.0.0


breun:
After upgrading Spring Cloud GCP from 3.4.3 to 4.0.0 a test started failing that was checking for the presence of a `SecretManagerTemplate` bean when `spring.cloud.gcp.secretmanager.enabled` is set to `true`.

In Spring Cloud GCP 3.4.3 this bean was created by the (deprecated) `com.google.cloud.spring.autoconfigure.secretmanager.GcpSecretManagerBootstrapConfiguration` class, but this class was removed in 4.0.0.

According to the [Spring Cloud GCP Secret Manager documentation](https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#secret-manager) a `SecretManagerTemplate` bean should still be created automatically. 

It looks like `com.google.cloud.spring.autoconfigure.secretmanager.SecretManagerConfigDataLocationResolver#registerSecretManagerBeans` is now the only method in Spring Cloud GCP which creates and registers a `SecretManagerTemplate` bean, but I believe that only comes into play when you have configuration property values prefixed with `sm://`, right?

breun:
I've created a small project with a test that reproduces this issue: https://github.com/breun/secretmanagertemplate-test

rotilho:
Even with `sm://` the secret manager doesn't lookup the secret.

7a8e709c:
@rotilho adding `spring.config.import=sm://` to property file helps me.

rotilho:
@7a8e709c thanks, that definitely works!

JoeWang1127:
Hi @breun, adding `spring.config.import=sm://` to the `application.properties` file should fix the issue.

meltsufin:
The `SecretManagerTemplate` should always be provided by the starter. I'm not sure if the same bean can be re-used for config properties, but we should try. Adding `spring.config.import=sm://` should not be required.

breun:
[...]

PR #1561, opened 2023-02-01
feat: enable `SecretManagerTemplate` bean by default

JoeWang1127:
Fix https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1551.

Enable `SecretManagerTemplate` bean by default, i.e., without specifying `spring.config.import` in `application.properties`.

### DIFF
--- a/docs/src/main/asciidoc/secretmanager.adoc
+++ b/docs/src/main/asciidoc/secretmanager.adoc
@@ -101,6 +101,7 @@ The `SecretManagerTemplate` class simplifies operations of creating, updating, a
 
 To begin using this class, you may inject an instance of the class using `@Autowired` after adding the starter dependency to your project.
 
+Note that you should add `spring.config.import=sm://` in your `application.properties` file to enable `SecretManagerTemplate` bean creation.
 [source, java]
 ----
 @Autowired
--- a/docs/src/main/asciidoc/secretmanager.adoc
+++ b/docs/src/main/asciidoc/secretmanager.adoc
@@ -49,8 +49,6 @@ If set to `true`, `null` will be returned when accessing a non-existent secret; 
 
 The Spring Framework on Google Cloud integration for Google Cloud Secret Manager enables you to use Secret Manager as a bootstrap property source.
 
-This allows you to specify and load secrets from Google Cloud Secret Manager as properties into the application context during the https://cloud.spring.io/spring-cloud-commons/reference/html/#the-bootstrap-application-context[Bootstrap Phase], which refers to the initial phase when a Spring application is being loaded.
-
 The Secret Manager property source uses the following syntax to specify secrets:
 
 [source]
@@ -82,7 +80,6 @@ You can use this syntax in the following places:
 [source]
 ----
 # Example of the project-secret long-form syntax.
-spring.config.import=sm://
 spring.datasource.password=${sm://projects/my-gcp-project/secrets/my-secret}
 ----
 The former is used to enable https://spring.io/blog/2020/08/14/config-file-processing-in-spring-boot-2-4[Spring Boot's Config Data API].
@@ -101,7 +98,6 @@ The `SecretManagerTemplate` class simplifies operations of creating, updating, a
 
 To begin using this class, you may inject an instance of the class using `@Autowired` after adding the starter dependency to your project.
 
-Note that you should add `spring.config.import=sm://` in your `application.properties` file to enable `SecretManagerTemplate` bean creation.
 [source, java]
 ----
 @Autowired
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -263,7 +263,6 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>spring-cloud-gcp-secretmanager</artifactId>
-            <optional>true</optional>
         </dependency>
 
         <!-- KMS -->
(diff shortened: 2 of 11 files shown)


ddixit14:
> Fix an issue in the Secret Manager docs per #1551.
> 
> Specify `spring.config.import=sm://` should be in the properties file for `SecretManagerTemplate` to be auto-configured.

@JoeWang1127 Adding do not merge label as I'm testing javadoc job for 4.0.0 release, do not want any additional PR's into main for sometime. Will remove it when I'm done. Sorry for the inconvenience.

JoeWang1127:
@ddixit14 sure, no problem.

docs/src/main/asciidoc/secretmanager.adoc
```diff
@@ -101,6 +101,7 @@ The `SecretManagerTemplate` class simplifies operations of creating, updating, a
 
 To begin using this class, you may inject an instance of the class using `@Autowired` after adding the starter dependency to your project.
 
+Note that you should add `spring.config.import=sm://` in your `application.properties` file to enable `SecretManagerTemplate` bean creation.
```
meltsufin, line 104: Should we also add it to the migration guide?
Alternatively, can this be done automatically in the auto-configuration?

JoeWang1127:
> Let's try to fix up the secretsmanager autocnfiguration to provide the `SecretManagerTemplate` bean.

Sure, will do.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -263,7 +263,6 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>spring-cloud-gcp-secretmanager</artifactId>
-            <optional>true</optional>
```
meltsufin, line 266: Can't do that because it will turn on the autoconfiguration even when the secrets manager starter is not imported.

JoeWang1127: By adding `spring.config.import=sm://` implicitly to property files, Spring Boot will try to register secret-manager related beans and get the following error:
>java.lang.NoClassDefFoundError: com/google/cloud/secretmanager/v1/SecretManagerServiceClient

	at com.google.cloud.spring.autoconfigure.secretmanager.SecretManagerConfigDataLocationResolver.registerSecretManagerBeans(SecretManagerConfigDataLocationResolver.java:71)
	at com.google.cloud.spring.autoconfigure.secretmanager.SecretManagerConfigDataLocationResolver.resolve(SecretManagerConfigDataLocationResolver.java:60)

This is because `spring-cloud-gcp-secretmanager` dependency doesn't in the classpath.

JoeWang1127: Maybe `isResolvable` method in `SecretManagerConfigDataLocationResolver` should return `false` if `spring-cloud-gcp-secretmanager` is not in the classpath.

meltsufin: I don't know if we should try to add `spring.config.import=sm://` implicitly. Can we just register a `SecretManagerTemplate` bean in the application context? Can we use the regular auto-configuration, or is there some issue because we're dealing bootstrap context?

JoeWang1127: When we use Spring Config API (thus deprecate bootstrap phase), all the Secret Manager related beans creation steps move to `SecretManagerConfigDataLocationResolver.resolve()` which only gets call if `spring.config.import=sm://` presents.

If we register a `SecretManagerTemplate` bean using auto-configuration, it's essentially brought back `GcpSecretManagerBootstrapConfiguration`.

meltsufin: Kind of, but it doesn't need to be bootstrap. I was thinking something like this: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1567.

The beans created with `SecretManagerConfigDataLocationResolver` would still take precedence over `GcpSecretManagerAutoConfiguration`, but if the resolver is not used, the `SecretManagerTemplate` bean would be created by the autoconfig. WDYT?

JoeWang1127: Sure, this might be possible.

Would you like me to close this PR and start from yours?

meltsufin: > By adding spring.config.import=sm:// implicitly to property files, Spring Boot will try to resolve Config Data Locations in ConfigDataEnvironmentPostProcessor, then SecretManagerConfigDataLocationResolver trys to register beans which eventually gets the following error:

Can you explain a bit more why there is a difference between adding the property implicitly vs. explicitly? Wouldn't `SecretManagerConfigDataLocationResolver` be activated either way?

JoeWang1127: In current implementation, `SecretManagerConfigDataLocationResolver.resolve()` will not be called if no `spring.config.import` exists.

By adding `spring.config.import` explicitly, we expect customers to add secret manager starter dependency since they want to use secret manager. In this case, `SecretManagerConfigDataLocationResolver.resolve()` is called.

However, if we add `spring.config.import` implicitly, i.e., adding it in `GcpSecretManagerEnvironmentPostProcessor`, `SecretManagerConfigDataLocationResolver.resolve()` will always be called so secret manager starter dependency is required in classpath.

meltsufin: Got it, thanks!

JoeWang1127:
Go with #1567 instead.

PR #1567, opened 2023-02-03
fix: create Secret Manager beans even without `spring.config.import=sm://` property

meltsufin:
Fix https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1551

After the fix, `spring.config.import=sm://` is still required if the customer wants to retrieve secrets from Secret Manager. However, `SecretManagerTemplate` bean exists even if this property doesn't present in property files.


meltsufin (commented):
Let's go with this. WDYT @JoeWang1127?

JoeWang1127:
Sure, as #1561 requires a lot of engineering effort to make it right.

JoeWang1127 (approved):

Issue #1570, opened 2023-02-03
Remove `SecretManagerPropertySourceLocator` as it is not needed


JoeWang1127:
Remove `SecretManagerPropertySourceLocator` as it is not needed

PR #1571, opened 2023-02-03
feat: remove `SecretManagerPropertySourceLocator`

JoeWang1127:
Fix https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1570


meltsufin (approved):
Since it's a public class, and there might be usage outside of the

JoeWang1127:
> Since it's a public class, and there might be usage outside of the

Should we deprecate the class and remove it in the next major version?

meltsufin:
> > Since it's a public class, and there might be usage outside of the
> 
> Should we deprecate the class and remove it in the next major version?

Oops, I didn't mean to send that. I started writing that and then changed my mind after looking more into the use of the class exclusively for bootstrap. We are still keeping the actual PropertySource. So, I think it's fine to remove.

JoeWang1127:
Cool, then I'll proceed with the merge.

PR #1606, opened 2023-02-23
chore(codegen): allow base branch of github action to be different from main

emmileaf:
This PR updates the existing github action to update generated modules, so that when the workflow is triggered from a branch other than `main`, the generated PR uses that branch as the base branch.

Use case: this allows for the workflow to be triggered from a branch created by dependabot (e.g. https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1582), and generated PR will contain both the dependabot commits and any updates to `spring-cloud-previews`.


diegomarquezp (approved):
LGTM

PR #1608, opened 2023-02-23
chore(codegen): add workflow to update generated modules directly in dependabot branch

emmileaf:
EDIT: This PR updates the existing `generateAutoConfigs.yml` workflow and pushes a commit to the (dependabot) triggered-from branch directly instead of creating a new branch and PR.
 
This is an incremental change that still relies on manual trigger (workflow_dispatch) on a specific branch pattern (tested in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1605). 

A next step would be to auto-trigger this workflow upon incoming dependabot PR event for libraries-bom version bump in main.


meltsufin (commented):
Why did you decide to keep `generateAutoConfigs.yml`? This is creating a bit of code duplication.

emmileaf:
> Why did you decide to keep `generateAutoConfigs.yml`? This is creating a bit of code duplication.

@meltsufin The consideration was just to keep the existing workflow as an option, in case this change runs into issues interacting with an actual dependabot PR. 

I can update this PR to modify `generateAutoConfigs.yml` directly, and we can revert this change if needed?

diegomarquezp (approved):

meltsufin (approved):

Issue #1643, opened 2023-03-06
DefaultSubscriberFactory is lacking minDurationPerAckExtension and maxDurationPerAckExtension setter


hkinder:
**Describe the bug**
As per com.google.cloud.spring.pubsub.support.DefaultSubscriberFactory details, the getter exists for "minDurationPerAckExtension" and "maxDurationPerAckExtension" and support some inner field that should be settable.
But there is no setter for both attributes.

PR #1644, opened 2023-03-06
fix: provide setters for min/max duration per ack ext

meltsufin:
Fixes: #1643.


JoeWang1127:
Do you think we should add some unit tests?

JoeWang1127 (approved):

PR #1653, opened 2023-03-17
chore: update generateAutoConfigs action to run on pr event

emmileaf:
This PR allows the action for updating generated modules to auto-trigger on dependabot PRs to update libraries-bom (pattern matching on branch name`dependabot/maven/com.google.cloud-libraries-bom*`).

I couldn't figure out how to test this without merging in the changes to main, but the workflow's existing manual trigger from branch name is still available, and tested in a mock branch (4b6c7437ef4e2e0d559d12bce6b245f6bfe3b75a).

### DIFF
--- a/.github/workflows/generateAutoConfigs.yaml
+++ b/.github/workflows/generateAutoConfigs.yaml
@@ -1,5 +1,11 @@
 name: Generate Spring Auto-Configurations
-on: workflow_dispatch
+on:
+  pull_request_target:
+    types:
+      - opened
+    branches:
+      - 'dependabot/maven/com.google.cloud-libraries-bom*'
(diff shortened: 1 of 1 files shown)


meltsufin (approved):
Looks great! Regarding @diegomarquezp's suggestion on do-not-merge, I think if we make this check required, it will be taken care of because it will be pending while the generation is in progress and won't allow the PR merge.

emmileaf:
Ohh blocking merge is a good idea, thanks for the suggestions! I'll try making the check required in settings after merging in this PR, and see if it can work as expected. If not, will look into @diegomarquezp's suggestion for adding a label :)

PR #1659, opened 2023-03-17
chore: fixes for generateAutoConfigs action on pull requests

emmileaf:
This PR contains two follow-up fixes after https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1656:

- Also moves `type: “opened”` into the if condition, so that the check can be triggered and skipped on synchronize/re-open for existing PRs 
- Switches to use `pull_request` event instead of `pull_request_target`, since we only expect running this workflow on branches within the repo, and not external forks. (Looking into this a bit more, [`pull_request_target`](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows) gives more privileged read/write/secret permissions to the workflow, and runs the action from the trusted target branch instead of the pull request branch. This doesn't work with the existing action unless we additionally checkout and make modifications from pull request head, which is probably not a good idea in combination with the additional permissions allowed by this event.)

### DIFF
--- a/.github/workflows/generateAutoConfigs.yaml
+++ b/.github/workflows/generateAutoConfigs.yaml
@@ -1,13 +1,11 @@
 name: Generate Spring Auto-Configurations
 on:
-  pull_request_target:
-    types:
-      - opened
+  pull_request:
   workflow_dispatch:
 jobs:
   generateLibraries:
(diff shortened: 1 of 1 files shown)


diegomarquezp (approved):
LGTM

PR #1415, opened 2022-12-29
chore(codegen): add preview modules to release profile

zhumin8:
Prepare to merge this when ready to release. 
Alter title to reflect what's expected to publish in release notes.
ref: [release please commits](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

-----
Update (@emmileaf) : release-please override for release note message will be added as part of PR description for #1517.
Merge order: #1517 > #1469 > This PR.

PR #1686, opened 2023-03-30
chore: initial migration of spring generator code 

emmileaf:
This PR contains an initial migration of generator code from `gapic-generator-java` (see https://github.com/googleapis/gapic-generator-java/pull/1580 for source) to the `spring-cloud-gcp` repo (alongside scripts and generated code).  

End-to-end workflow can be verified from this PR by creating a branch with libraries-bom version change and running:
```
gh workflow run .github/workflows/generateAutoConfigs.yaml --ref generator-migration-test
-f branch_name=<test-branch-name> -f forked_repo=none
```

**For reviewers:** 
Please see https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686#issuecomment-1510004667 for an outline of file roles (migrated content vs. original changes introduced by the migration), to help with navigating the PR diff.

- Review follow-ups are tracked in [this comment](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686#issuecomment-1512084004).
----

**High-level migration approach:**
1. Set up `spring-cloud-generator` as a Bazel module (name: `spring_cloud_generator`, with [BUILD.bazel](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/generator-migration-test/spring-cloud-generator/BUILD.bazel) and [WORKSPACE](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/generator-migration-test/spring-cloud-generator/WORKSPACE))
2. Set up `spring-cloud-generator` as a Maven submodule in `spring-cloud-gcp` (artifact name: `com.google.api.spring-cloud-generator`, with [pom.xml](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/generator-migration-test/spring-cloud-generator/pom.xml))
    - Version is aligned with `spring-cloud-gcp`, and has dependency on published `gapic-generator-java` jar 
    - Not intended for release; updated script to [install a local SNAPSHOT](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/62c62c997d21054ef145411a1abda20ae98a6918/spring-cloud-generator/scripts/generate-all.sh#L42) version of this module for use in generation
3. Adjustments to (and restructuring of) [scripts](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/tree/generator-migration-test/spring-cloud-generator/scripts) and [github workflow](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/generator-migration-test/.github/workflows/generateAutoConfigs.yaml) to make local googleapis workspace modifications, trigger Bazel targets using the migrated generator, and process generated code into `spring-cloud-previews` 

**Other smaller changes:**
- Updated perl script replacement logic in scripts with buildozer for better readability
- Removed outdated `library_list.txt`, and added step in gh action to log contents after generating this file
- Updated license headers formatting (`//` => `/*`) to align with `spring-cloud-gcp` checks

**Planned enhancements (not included in this PR):** 
- Remove duplication of test-related code and helpers, after publishing test jar from gapic-generator-java. Currently this is copied to the following locations (see [0c26849](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/commit/0c26849fc27d4e8b2b3aca663d045db490a7d131) for scope of code duplication):
    * In [src/test/java/com/google/api/generator/spring/utils](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/tree/generator-migration-test/spring-cloud-generator/src/test/java/com/google/api/generator/spring/utils)
        - Golden testing helpers from gapic-generator-java [src/test/java/com/google/api/generator/test/framework](https://github.com/googleapis/gapic-generator-java/tree/main/gapic-generator-java/src/test/java/com/google/api/generator/test/framework)
        - Two test proto loaders from gapic-generator-java [src/test/java/com/google/api/generator/gapic/composer](https://github.com/googleapis/gapic-generator-java/tree/main/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer)
    * In [src/test/proto](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/tree/generator-migration-test/spring-cloud-generator/src/test/proto)
        - Two test protos from gapic-generator-java [src/test/proto](https://github.com/googleapis/gapic-generator-java/tree/main/gapic-generator-java/src/test/proto)
    * Testing-related dependencies in [pom.xml](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/generator-migration-test/spring-cloud-generator/pom.xml) (marked with [gapic-test]) 
- Additional simplification of scripts and github actions workflow


.github/workflows/testGenerateAutoConfigs.yaml
```diff
@@ -0,0 +1,119 @@
+name: Test Action for Generating Spring Auto-Configurations
```
emmileaf, line 1: This file will be removed before merging.

spring-cloud-generator/scripts/setup-googleapis-rules.sh
```diff
@@ -0,0 +1,27 @@
...
+buildozer 'set path "../../spring-cloud-generator"' WORKSPACE:spring_cloud_generator
+
+# delete existing maven_install rules
+buildozer 'delete' WORKSPACE:%maven_install
+
+# In googleapis/WORKSPACE, find maven_install() rule for gapic-generator-java jar, and remove
+# perl -0777 -pi -e "s{maven_install\(\n    artifacts = \[(.*?)_gapic_generator_java_version(.*?)\](.*?)\)}{}s" WORKSPACE
+
+# add custom maven_install rules
+perl -pi -e "s{(^_gapic_generator_java_version[^\n]*)}{\$1\n$(cat ../scripts/resources/googleapis_modification_string.txt)}" WORKSPACE
```
emmileaf, line 17: Modifies googleapis/WORKSPACE file to use similar approach as gapic libraries generation: build spring-cloud-generator local jar, then pass this into bazel plugin to generate the spring modules.

spring-cloud-generator/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>spring-cloud-gcp</artifactId>
+        <groupId>com.google.cloud</groupId>
+        <version>4.1.5-SNAPSHOT</version><!-- {x-version-update:spring-cloud-gcp:current} -->
+    </parent>
+
+    <groupId>com.google.api</groupId>
```
emmileaf, line 12: I kept this group ID same as the one used by gapic-generator-java jar, but not sure if it should be consistent with others from this repo (`com.google.cloud`) instead?

diegomarquezp: I noticed that we have `java-core` in the `gapic-generator-java` monorepo [using](https://github.com/googleapis/gapic-generator-java/blob/main/java-core/pom.xml) the `com.google.cloud` group ID, unlike the `com.google.api` group as in other libraries in there, so an argument of consistency may not be as strong (although `java-core` was recently added to the monorepo).

IMO if we have no technical difficulties to keep the group ID consistent with `spring-cloud-gcp` then it may make more sense to assimilate it into spring-cloud-gcp rather than making it consistent with what is now a dependency of the spring generator.

meltsufin:
Since this is 5K+ line PR, would you be able to provide some guidance on what needs to be reviewed vs. code that was migrated verbatim?

emmileaf:
> Since this is 5K+ line PR, would you be able to provide some guidance on what needs to be reviewed vs. code that was migrated verbatim?

For sure! For this PR I am mostly looking for a review on the maven/bazel setup and module structure. Here’s a quick outline of the different files’ roles/sources of migration, which hopefully helps with navigating the PR diff (glad to explain any change in more detail if needed):

### Setup-related or edited files, looking for review:

**Setup (maven):**
- pom.xml
- spring-cloud-generator/pom.xml

**Setup (bazel):**
- spring-cloud-generator/BUILD.bazel
- spring-cloud-generator/WORKSPACE
- spring-cloud-generator/java_gapic_spring.bzl (migrated verbatim from [#1580](https://github.com/googleapis/gapic-generator-java/pull/1580))

**Scripts:**
- spring-cloud-generator/scripts
   * I relocated these scripts along with making edits to them, and github was able to pick up the diff pretty well with the exception of `setup-build-rule.sh` and `setup-googleapis-rules.sh`: these two show up as new files in git likely because of the larger scope of edits; aside from changes to get the bazel setup working, the previous perl logic was also replaced with buildozer where possible to help with maintainability.

**Github action:**
- ./github/workflows
  * Changes to `generateAutoConfigs.yaml` intend to be merged; `testGenerateAutoConfigs.yaml` is added for ease of testing end-to-end workflow on this branch, and will be removed before merging

### Migrated files, not looking for review of contents:

**Java files:**
- No original code changes introduced, however I updated the license headers formatting to comply with spring-cloud-gcp enforced checks
- spring-cloud-generator/src/main: migrated verbatim from [#1580](https://github.com/googleapis/gapic-generator-java/pull/1580)
- spring-cloud-generator/src/test: migrated verbatim from [#1580](https://github.com/googleapis/gapic-generator-java/pull/1580) with the exception of /utils, which contain:
  * Golden testing helpers from (copied from gapic-generator-java [src/test/java/com/google/api/generator/test/framework](https://github.com/googleapis/gapic-generator-java/tree/main/gapic-generator-java/src/test/java/com/google/api/generator/test/framework), with intention of eventual removal)
  * Two test proto loaders from (copied from gapic-generator-java [src/test/java/com/google/api/generator/gapic/composer](https://github.com/googleapis/gapic-generator-java/tree/main/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer) , with intention of eventual removal)

**Proto files:**
- spring-cloud-generator/src/test/proto: copied from [gapic-generator-java](https://github.com/googleapis/gapic-generator-java/tree/main/gapic-generator-java/src/test/proto)

spring-cloud-generator/pom.xml
emmileaf: That's a good point! Yeah switching to use `com.google.cloud` here shouldn't be a technically difficult one, just wanted to check if we have any conventions/preferences here (in case we decide to also publish this generator jar in the future).

spring-cloud-generator/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    <properties>
+        <!-- [gapic-test]: for compiling test protos -->
+        <protobuf.version>3.21.12</protobuf.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.google.api</groupId>
+            <artifactId>gapic-generator-java</artifactId>
+            <version>2.17.0</version>
```
meltsufin, line 26: @blakeli0 Can we add `gapic-generator-java` to the [gapic-generator-java-bom](https://github.com/googleapis/gapic-generator-java/tree/main/gapic-generator-java-bom) so that we don't duplicate version specs?

spring-cloud-generator/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+            <extension>
+                <!-- [gapic-test]: for use in protobuf-maven-plugin below -->
+                <groupId>kr.motd.maven</groupId>
+                <artifactId>os-maven-plugin</artifactId>
+                <version>1.7.1</version>
+            </extension>
+        </extensions>
+        <plugins>
+            <plugin>
+                <!-- [gapic-test]: for compiling test protos -->
```
meltsufin, line 53: Is this an attempt to replace a Bazel target with a Maven plugin? If so, what's the benefit over keeping it in Bazel?

spring-cloud-generator/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>spring-cloud-gcp</artifactId>
+        <groupId>com.google.cloud</groupId>
+        <version>4.1.5-SNAPSHOT</version><!-- {x-version-update:spring-cloud-gcp:current} -->
+    </parent>
+
+    <groupId>com.google.api</groupId>
+    <artifactId>spring-cloud-generator</artifactId>
```
meltsufin, line 13: Is there a Maven release exclusion for this module? I don't think we want to be released together with other Spring Cloud GCP maven modules just yet.

meltsufin (changes requested):
I think it's important to make sure that we don't accidentally publish the new `spring-cloud-generator` module to Maven Central.
Aside from that, LGTM!

spring-cloud-generator/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -0,0 +1,1146 @@
...
+    String credentialsProviderName = serviceNameLowerCamel + "Credentials";
+    String transportChannelProviderName = "default" + serviceName + "TransportChannelProvider";
+    String clientName = serviceNameLowerCamel + "Client";
+    GapicClass.Kind kind = Kind.MAIN;
+
+    GapicServiceConfig gapicServiceConfig = context.serviceConfig();
+
+    Expr thisExpr = ValueExpr.withValue(ThisObjectValue.withType(dynamicTypes.get(className)));
+    Transport transport = context.transport();
+    // TODO(emmwang): this condition is adapted from latest gapic-generator-java changes as part of
```
blakeli0, line 115: Can we fix this TODO as part of this PR?

spring-cloud-generator/pom.xml
blakeli0: I think this is more to make the generation process similar to gapic-generator-java, meaning that we build a `spring-cloud-generator` jar with Maven first and use it in the generation process. I don't think we can run tests through Maven without this change, similar to what [gapic-generator-java](https://github.com/googleapis/gapic-generator-java/blob/eee46fe2cbc466549115b943a79d7863bdc74c40/gapic-generator-java/pom.xml#L248) is doing.

spring-cloud-generator/pom.xml
blakeli0: I think we could, not sure why it was excluded initially though, maybe @suztomo knows? We might be able to extend the protobuf version from [here](https://github.com/googleapis/gapic-generator-java/blob/eee46fe2cbc466549115b943a79d7863bdc74c40/gapic-generator-java-pom-parent/pom.xml#L33) as well.

spring-cloud-generator/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    <name>Spring Code Generator for Google Cloud</name>
+    <description>Spring Code Generator for Google Cloud</description>
+
+    <properties>
+        <!-- [gapic-test]: for compiling test protos -->
+        <protobuf.version>3.21.12</protobuf.version>
```
blakeli0, line 19: See comment below for the potential of extending from parent poms.

spring-cloud-generator/pom.xml
blakeli0: I feel `com.google.cloud` makes more sense as I consider this project is closer to `Spring-cloud-gcp` than `gapic-generator-java`, as this project is very specific to `Spring-cloud-gcp`, but we could `gapic-generator-java` as a generic tool.

pom.xml
```diff
@@ -371,6 +371,7 @@
 				<module>spring-cloud-gcp-kms</module>
 
 				<module>spring-cloud-previews</module>
+				<module>spring-cloud-generator</module>
```
blakeli0, line 374: Is it going to be release automatically if we add it here?

spring-cloud-generator/WORKSPACE
```diff
@@ -0,0 +1,72 @@
...
+_spring_cloud_generator_version = "4.1.4-SNAPSHOT" # {x-version-update:spring-cloud-gcp:current}
+
+maven_install(
+    artifacts = [
+        "com.google.api:spring-cloud-generator:" + _spring_cloud_generator_version,
+    ],
+    fail_on_missing_checksum = False,
+    repositories = [
+        "m2Local",
+        "https://repo.maven.apache.org/maven2/",
```
blakeli0, line 51: We can remove this line if we are always going to install it from local

spring-cloud-generator/WORKSPACE
```diff
@@ -0,0 +1,72 @@
+workspace(name = "spring_cloud_generator")
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+
+#googleapis
+
+_googleapis_commit = "7438480b2a1bc6371d748e974f7a3647f90c4e8d"
```
blakeli0, line 8: What is this googleapis commit use for? I thought for the actual generation process we are going to dynamically compute the committish for each library?

spring-cloud-generator/scripts/generate-all.sh
```diff
@@ -24,32 +24,36 @@ save_error_info () {
...
 
 done <<< $libraries
 
+# install local snapshot jar for spring generator
+cd ${WORKING_DIR} && mvn install
+
 # fetches all `*java_gapic_spring` build rules and build them at once
 cd googleapis
-bazelisk query "attr(name, '.*java_gapic_spring', //...)" | xargs bazelisk build 2>&1 \
+bazelisk query "attr(name, '.*java_gapic_spring', //...)" \
```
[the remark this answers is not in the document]
blakeli0: If I'm understanding correctly, this is only an issue for spring-cloud-gcp 4.x.x, right? Do we support SpringCodeGen for spring-cloud-gcp 3.x.x version? If yes, we may want to make it a variable and still support Java 8.

spring-cloud-generator/scripts/resources/googleapis_modification_string.txt
```diff
@@ -0,0 +1,14 @@
+_spring_cloud_generator_version = "4.1.5-SNAPSHOT" # {x-version-update:spring-cloud-gcp:current}
+
+maven_install(
+    artifacts = [
+        "com.google.api:gapic-generator-java:" + _gapic_generator_java_version,
```
blakeli0, line 5: We probably don't need to install `gapic-generator-java` anymore, as `spring-cloud-generator` should already have it.

spring-cloud-generator/pom.xml
emmileaf: Ah yeah this is part of copying over testing setup from gapic-generator-java - specifically it's for compiling protos used for unit tests ([source](https://github.com/googleapis/gapic-generator-java/blob/eee46fe2cbc466549115b943a79d7863bdc74c40/gapic-generator-java/pom.xml#L248)) (allowing `TestProtoLoader` to parse the proto).   

It can later be removed given a published test jar from gapic-generator-java as discussed; currently this setup allows `mvn test -DupdateUnitGoldens` to work in this submodule.

spring-cloud-generator/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
emmileaf: Thanks for catching this! Will update as part of this PR.

spring-cloud-generator/scripts/generate-all.sh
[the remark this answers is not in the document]
emmileaf: That's correct, and we only support the codegen modules on spring-cloud-gcp 4.x.x (main).

spring-cloud-generator/pom.xml
emmileaf: Addressing this comment along with https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686#discussion_r1169187271:

This PR adds `spring-cloud-generator` as a module under spring-cloud-gcp's default profile:

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/7d14fa45496bbe71ac252f8fd98ddb9489aec463/pom.xml#L342-L349

But not the release profile:
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/7d14fa45496bbe71ac252f8fd98ddb9489aec463/pom.xml#L565-L570

This is similar to what we did for the `spring-cloud-previews` module before their actual release as part of https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1415. 

Please let me know if there is anything I've overlooked though - I can also double check with running a stage-only job from this PR before merging.

pom.xml
emmileaf: See https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686#discussion_r1169251614.

emmileaf:
Thanks so much @diegomarquezp @meltsufin @blakeli0  for the first pass review! I’m replying to the comments ongoingly, but will track a code change to-do list here from the thread discussions and explicitly re-request a re-review on the PR after all of them are made (and workflow run re-verified).

Really appreciate all your help and patience with reviewing this!

---

**Code change todos:**
- [ ] Switch group id to com.google.cloud ([comment](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686#discussion_r1165625893))
- [ ] Fix inline todo ([comment](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686#discussion_r1169167504))
- [ ] Check/remove unused googleapis modifications ([comment1](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686#discussion_r1169190414), [comment2](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686#discussion_r1169207230))

spring-cloud-generator/pom.xml
meltsufin: The `spring-cloud-previews` also has the following section:
```
  <build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <version>${maven-deploy-plugin.version}</version>
        <configuration>
          <skip>true</skip>
        </configuration>
      </plugin>
    </plugins>
  </build>
```

I think we put it in as a second layer prevention in addition to the profiles in parent pom.

spring-cloud-generator/WORKSPACE
emmileaf: You're right, I accidentally left this in from copying gapic-generator-java setup but it should be unused in the spring code generation process. Will remove.

emmileaf: Removed in [d9bb09f](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686/commits/d9bb09f1bc7eb9edd5e35ce0ed61028b866ac54e).

spring-cloud-generator/WORKSPACE
emmileaf: Removed in [d9bb09f](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686/commits/d9bb09f1bc7eb9edd5e35ce0ed61028b866ac54e).

spring-cloud-generator/scripts/resources/googleapis_modification_string.txt
emmileaf: Removed in [d9bb09f](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686/commits/d9bb09f1bc7eb9edd5e35ce0ed61028b866ac54e).

spring-cloud-generator/pom.xml
emmileaf: Thanks for the pointer! Added this skip configuration to the pom in [079b08f](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686/commits/079b08fd47bca2b36db0757a160a6c44f2d6c0c5).

spring-cloud-generator/pom.xml
suztomo: No concern in adding it to the BOM. Because it's not a user-facing artifact, we didn't add it to the BOM so far.

emmileaf: Thanks! Added to BOM in https://github.com/googleapis/gapic-generator-java/pull/1645; will update this PR when this becomes available in the next release.

spring-cloud-generator/src/test/java/com/google/cloud/generator/spring/utils/TestUtils.java
```diff
@@ -0,0 +1,80 @@
...
+
+package com.google.cloud.generator.spring.utils;
+
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+// Renamed from Utils
+public class TestUtils {
```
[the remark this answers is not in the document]
emmileaf: That’s a good point, I think we could benefit from cleaning this up at the gapic-generator-java source as well, since it gets confusing with classes that are meant to be tests for utils vs. utils for tests 🤯 

I’ve started a [WIP PR](https://github.com/googleapis/gapic-generator-java/pull/1648) in gapic-generator-java to publish a testlib jar that can hopefully get rid of the need for this file duplication altogether. In the meanwhile, will rename this helper class to something less ambiguous here.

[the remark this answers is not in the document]
emmileaf: Renamed to `GoldenFileWriter`.

spring-cloud-generator/pom.xml
emmileaf: Added `dependencyManagement` block with the newest release of gapic-generator-java-bom  2.18.0.

spring-cloud-generator/pom.xml
emmileaf: @blakeli0 I wasn't able to find a straightforward way to inherit this version from `gapic-generator-java-bom`, since here `protobuf.version` is used as a property for configuring `protobuf-maven-plugin`, rather than as a dependency. 

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/7d14fa45496bbe71ac252f8fd98ddb9489aec463/spring-cloud-generator/pom.xml#L53-L61

This plugin is only used for compiling test proto files for unit testing, so I’ve updated the inline comment to explain this for now. To follow-up on this PR, I’ll look into whether it can be removed altogether with a testlib jar published from `gapic-generator-java`.

spring-cloud-generator/WORKSPACE
```diff
@@ -0,0 +1,58 @@
...
+
+load("@rules_jvm_external//:setup.bzl", "rules_jvm_external_setup")
+
+rules_jvm_external_setup()
+
+load("@rules_jvm_external//:defs.bzl", "maven_install")
+
+
+# Spring generator
+_spring_cloud_generator_version = "4.1.4-SNAPSHOT" # {x-version-update:spring-cloud-gcp:current}
```
blakeli0, line 29: I think this version needs to be in sync with the latest spring-cloud-gcp version? I'm actually not sure how it worked with the wrong version.

spring-cloud-generator/java_gapic_spring.bzl
```diff
@@ -0,0 +1,146 @@
+# Copyright 2022 Google LLC
```
blakeli0, line 1: New files in this repo need to be have `2023` in the copyright header

blakeli0: Also the copyright format is different from [other files](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-storage/src/main/java/com/google/cloud/spring/storage/GoogleStorageProtocolResolver.java#L1C1-L15) in this repo

spring-cloud-generator/WORKSPACE
emmileaf: You're right - thanks for catching this! I'm also confused to how this slipped past and kept working without the version being in sync - will take a look.

emmileaf: Ah, this may have worked with the incorrect version because the setup here is redundant/unused in the generation process. I've trimmed away most of this file in [87df937](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686/commits/87df9371a34c499de8e5df0f799c8263440462ba), with a quick explanation below:

When initially following the `gapic-generator-java` approach, I was under the impression that the generator WORKSPACE needed a `maven_install()` for the end-to-end generation, and followed most of this setup. 

On a second look at `gapic-generator-java/WORKSPACE`, I realize that it's [fetching a snapshot jar](https://github.com/googleapis/gapic-generator-java/blob/4054bc668daf797a8fce798843e0ec4cda669faa/WORKSPACE#L63), so probably just to allow for bazel invocations from within the repo (for integration tests?), while `googleapis/WORKSPACE` [fetches a published version](https://github.com/googleapis/googleapis/blob/bab324b44955494aebc6d572cee061202c4f811f/WORKSPACE#L268). In the spring codegen process, because we are not publishing the generator, the local copy of `googleapis/WORKSPACE` is already being modified with a [maven_install()](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/generator-migration-test/spring-cloud-generator/scripts/resources/googleapis_modification_string.txt) block to fetch the [locally built](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/87df9371a34c499de8e5df0f799c8263440462ba/spring-cloud-generator/scripts/generate-all.sh#L47-L48) generator as part of scripts. The version was correct there, which is why the generation workflow had worked.

This setup may be needed again when adding integration tests for the spring generator, but I will remove it for now given the additional version management.

spring-cloud-generator/java_gapic_spring.bzl
emmileaf: Per discussion - updated license header year to 2023 ([eefaeb6](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1686/commits/eefaeb6aa3b2c0b5c84d420e82bce1c8df204195)) for migrated files that previously lived on branch, but not published to gapic-generator-java main.

spring-cloud-generator/pom.xml
emmileaf: Makes sense, thanks for the explanation! 

I think renovate bot won't pick up on updating this version, since the only usage of the property is in a configuration literal and not a maven `<dependency>` block. Will keep an eye out and add an exclusion if one does come through.

blakeli0 (approved):

PR #1782, opened 2023-04-28
fix: Implementing count with aggregation query

### LINKED ISSUE #1781: Datastore: Using new datastore aggregation queries for count
In the current implementation, the library performs client-side aggregations by fetching all the keys on the client side and then figuring out the count. In theory, the new aggregation queries and COUNT aggregation should be faster than the current one, as the current implementation makes use of the lazy iterator from the underlying client library that initiates multiple backend calls in cases where the number of keys is greater than the page size specified. And on top of that getting all the keys on the client side will have **more egress cost** compared to getting a count value.

Reference: [COUNT aggregation and Aggregation queries in datastore](https://cloud.google.com/datastore/docs/aggregation-queries)

its-snorlax:
Modifying the implementation of `DatastoreTemplate#count` to use recently introduced [COUNT aggregation and Aggregation queries in datastore](https://cloud.google.com/datastore/docs/aggregation-queries).

fix #1781


meltsufin (approved):
LGTM. @jainsahab Please take a look as well.

jainsahab (approved):
LGTM

its-snorlax:
@meltsufin @jainsahab  are we good to merge now?

meltsufin:
> @meltsufin @jainsahab are we good to merge now?

It was [merged](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/commit/1ee2244c9b82980948be0a5d3bac38f76857dfe7), but due to some quirk in GitHub the PR didn't close.

PR #1800, opened 2023-05-05
fix: regression of outbound pubsub message handling (success/failure callbacks both called regardless of success/failure)

### LINKED ISSUE #1799: Spring Boot 3 changes to PubSubPublishCallback have resulted in failure callback being called when operation was successful
**Describe the bug**
Recent `PubSubPublishCallback` changes results in pub sub outbound message handling calling both the success and failure callbacks, regardless of success or failure. This is confusing and a regression from previous behaviour.

nathanpiper:
Fix #1799.


meltsufin:
@nathanpiper Thanks for the fix. Can you please sign the CLA?

JoeWang1127 (approved):

nathanpiper:
Thanks for reviewing. I just realised my commits have the wrong email address. What’s the best way to fix this? Should I just close this PR, create a new one and reference this one?

JoeWang1127:
I think you can change the email in the commits.

nathanpiper:
Thanks, I have updated it now.

PR #1816, opened 2023-05-08
test: add language starter module back to sonar analysis

emmileaf:
This PR update sonar exclusion on `spring-cloud-previews` so that analysis runs on `google-cloud-language-spring-starter`, which contain the handwritten unit tests. This allows for some indication of missing coverage should there be broader configuration code changes introduced by the generator.
* Also updates the tests so that existing code coverage is sufficient (previously missing for some method-level retry property setters)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<integration-test.pattern>**/*IntegrationTest*</integration-test.pattern>
 		<!-- exclude preview module code from sonar cloud analysis -->
-		<sonar.exclusions>spring-cloud-previews/**/!(LanguageServiceSpring*).java</sonar.exclusions>
+		<sonar.exclusions>spring-cloud-previews/!(google-cloud-language-spring-starter)/**</sonar.exclusions>
 	</properties>
 
 	<dependencyManagement>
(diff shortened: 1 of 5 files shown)


emmileaf:
Note: the final sonarcloud comment will show "no coverage info" on this PR because it does not make any changes to the autoconfiguration code. With the mock changes in commit https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1816/commits/b7acf97456a63812f7e07a792c4c4627c2a95d87, this analysis looks like below, catching changes to the `google-cloud-language-spring-starter` module and indicating missing coverage on new lines:

image

meltsufin (approved):
LGTM.
PS: I didn't realize we were mixing auto-generated code with handwritten, although it's just tests. I'm curious @blakeli0's opinion on this. Is this a testing strategy we want to employ?

meltsufin:
> LGTM. About test strategy, we created manual written unit test for one sample generated module and it is "mixed-in" as copied from template folder to generated code to run test with any updates to the generated module. It's the simplest setup to have coverage on the generated code. Alternative could be to copy generated "language-starter" module after generated to a separate un-published module where these tests locates and do testing there. What do you all think? @emmileaf @meltsufin @blakeli0 Is it worth the change so we can keep generated code modules clean from manual code?

I think that the purpose of the tests is to test the Spring starter generator. So, I wonder if you considered generating a starter for Showcase (or another API) and writing manual tests against it, all inside the `spring-cloud-generator` module?

PR #1849, opened 2023-05-15
feat: add support for rest-only services in generator 

emmileaf:
#### Generator changes:
* Removed exclusion for rest-only libraries in `generate-library-list.sh`
* Removed exclusion for services with rest-only transport in `SpringComposer`  
* Added golden tests
* In `java-compute`, `newBuilder()`, [`defaultTransportChannelProvider()`](https://github.com/googleapis/google-cloud-java/blob/720796b7e1218b9517a3f2ebcacd3119448e2e9c/java-compute/google-cloud-compute/src/main/java/com/google/cloud/compute/v1/stub/AddressesStubSettings.java#L347), and [`defaultApiClientHeaderProviderBuilder()`](https://github.com/googleapis/google-cloud-java/blob/720796b7e1218b9517a3f2ebcacd3119448e2e9c/java-compute/google-cloud-compute/src/main/java/com/google/cloud/compute/v1/stub/AddressesStubSettings.java#L352) all default to REST implementation, so spring composer logic is kept the same as handling GRPC-only libraries: do not generate configurable `useRest` property, and use the default builders above in generated auto-configuration.

#### For reviewers: 
 * Looking for a review on generator changes to include a starter module for compute (REST-only, GAPIC library): `spring-cloud-previews/google-cloud-compute-spring-starter`
 * Corresponding changes to generated code are now included this PR at https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849/commits/34495a0c768097d5aa430b9a8fe8edc16f2eef21. (Previously in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849/commits/213add001a6310bb1900a9a5f81df077b4fd6efe, [#1850](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1850))
   * Added module `google-cloud-compute-spring-starter`
   * Across all generated modules: updated javadoc comment for `TransportChannelProvider` bean to reflect the current default behavior
   * Updated copyright year in license headers for auto-generated files (from upstream template change in `gapic-generator-java`

----

Memo on outstanding blocking changes:
- [x] Update to next release of gapic-generator-java containing [license header year fix](https://github.com/googleapis/sdk-platform-java/pull/1720), and re-run changes to generated code ([thread](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849#discussion_r1207189165))

Memo on non-blocking discussions / open follow-ups:
- Alternative strategies for disabling/enabling autoconfiguration classes, and how these affect application start-up time ([discussion](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1850#issuecomment-1549882399))
- Consider renaming and perhaps expanding configuration scope of the `<service.prefix>.useRest` property ([discussion](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849#discussion_r1201403836))
- Workflow to push generated code changes introduced by generator changes as part of CI ([discussion](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849#issuecomment-1559746967))

----

BEGIN_COMMIT_OVERRIDE

feat: add google-cloud-compute-spring-starter to starter modules available for preview

END_COMMIT_OVERRIDE


spring-cloud-generator/src/test/proto/compliance.proto
```diff
@@ -0,0 +1,232 @@
+// Copyright 2021 Google LLC
```
blakeli0, line 1:
```suggestion
// Copyright 2023 Google LLC
```

spring-cloud-generator/src/test/proto/compliance.proto
```diff
@@ -0,0 +1,232 @@
...
+
+package google.showcase.v1beta1;
+
+option go_package = "github.com/googleapis/gapic-showcase/server/genproto";
+option java_package = "com.google.showcase.v1beta1";
+option java_multiple_files = true;
+
+// This service is used to test that GAPICs can transcode proto3 requests to
+// REST format correctly for various types of HTTP annotations.
+service Compliance {
```
blakeli0, line 29: Why do we add a new proto? Is there any test cases can not be tested by `echo.proto`?

spring-cloud-generator/src/test/java/com/google/cloud/generator/spring/utils/RestTestProtoLoader.java
```diff
@@ -0,0 +1,124 @@
+// Copyright 2021 Google LLC
```
blakeli0, line 1:
```suggestion
// Copyright 2023 Google LLC
```

blakeli0 (commented):
> In java-compute, newBuilder(), [defaultTransportChannelProvider()](https://github.com/googleapis/google-cloud-java/blob/720796b7e1218b9517a3f2ebcacd3119448e2e9c/java-compute/google-cloud-compute/src/main/java/com/google/cloud/compute/v1/stub/AddressesStubSettings.java#L347), and [defaultApiClientHeaderProviderBuilder()](https://github.com/googleapis/google-cloud-java/blob/720796b7e1218b9517a3f2ebcacd3119448e2e9c/java-compute/google-cloud-compute/src/main/java/com/google/cloud/compute/v1/stub/AddressesStubSettings.java#L352) all default to REST implementation

This is good to know but also surprising, because it means if we change a `REST` only client to `GRPC+REST`, it would be a breaking change.

blakeli0:
> Please see https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1850 for showing of generated code corresponding to these changes

Not part of this PR, I wonder if it makes sense to regenerate the auto configs as part of the CI if we have changes in SpringCodeGen, so that we can see the changes directly in the same PR.

spring-cloud-generator/src/test/proto/compliance.proto
emmileaf: No specific reason; I added the compliance proto because it was used in the gapic generator golden tests for REST-specific composers, but the spring composers don't currently implement anything that's specific to be tested by this proto. 

If the concern is to reduce the number of ptotos copied and maintained, these golden tests can switch to parse and use `echo.proto` (as `Transport.REST`).

spring-cloud-generator/java_gapic_spring.bzl
```diff
@@ -78,6 +78,7 @@ def java_gapic_spring_library(
         gapic_yaml = None,
         service_yaml = None,
         transport = None,
+        rest_numeric_enums = False,
```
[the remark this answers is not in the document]
emmileaf: Ah, you're right; will revert this change.

spring-cloud-generator/src/test/java/com/google/cloud/generator/spring/utils/RestTestProtoLoader.java
emmileaf: This helper was also copied (source: [RestTestProtoLoader](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/test/java/com/google/api/generator/test/protoloader/RestTestProtoLoader.java)), and soon to be removed when the next gapic-generator-java release publishes the testlib. Would this warrant a header year update, or should the same copyright year be kept as source?

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -113,7 +113,7 @@ public GapicClass generate(GapicContext context, Service service) {
     Expr thisExpr = ValueExpr.withValue(ThisObjectValue.withType(dynamicTypes.get(className)));
     Transport transport = context.transport();
     boolean hasRestOption =
-        (transport.equals(Transport.GRPC_REST) || transport.equals(Transport.REST))
+        transport.equals(Transport.GRPC_REST)
```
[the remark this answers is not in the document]
blakeli0: Thanks, it makes sense. Can we rename this property to make it clearer? Maybe something like `shouldSupportRestForGrpcClient`?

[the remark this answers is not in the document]
blakeli0: In addition, since this logic is shared between two composers, we may want to extract it to a common place(Util or maybe a AbstractSpringComposer).

spring-cloud-generator/src/test/java/com/google/cloud/generator/spring/utils/RestTestProtoLoader.java
blakeli0: I see, I think it makes sense to keep it as it is then.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringAutoConfigClassComposer.java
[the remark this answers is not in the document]
zhumin8: A quick follow-up on this:
I haven't dig into the generator code on rest-only service, but is it true that as far as for spring code, "rest-only" services are expected to generate exactly the same as "grpc-only" service? Can we add a comment somewhere in the code to document this behavior? I think it's helpful for future maintenance to have a quick link on the code base for background.

spring-cloud-generator/src/test/proto/compliance.proto
emmileaf: Proto removed per https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849#discussion_r1198382941.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringAutoConfigClassComposer.java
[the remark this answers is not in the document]
emmileaf: @blakeli0 Yup, renaming and extracting the logic makes sense. Github did not refresh properly for me and I had missed these latest comments when requesting your re-review; please feel free to ignore!

> "rest-only" services are expected to generate exactly the same as "grpc-only" service?

@zhumin8 Yes, that is the current implementation, which I did not expect at first as well. I can definitely add in a comment to explain this along with the `hasRestOption` refactoring above.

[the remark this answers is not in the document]
emmileaf: Update in [8d7954f](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849/commits/8d7954fc43498ebeccb4a58be5f1972d97814f0a) with both suggestions above. 

> Can we rename this property to make it clearer? Maybe something like shouldSupportRestForGrpcClient?

@blakeli0 One quick clarification - for this comment, did you mean renaming `hasRestOption` and related variables/utils within the generator code, or renaming the exposed Spring configuration property `useRest`? I made changes for the former since the latter would be a breaking change for users, but just wanted to make sure I'm not misunderstanding. Thanks!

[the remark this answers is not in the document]
blakeli0: Yes, the former.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -387,7 +385,8 @@ private static MethodDefinition createTransportChannelProviderBeanMethod(
             .setReturnType(STATIC_TYPES.get("TransportChannelProvider"))
             .build();
 
-    if (hasRestOption) {
+    // For GRPC+REST libraries, choose provider according to configuration property
+    if (ComposerUtils.shouldSupportRestOptionWithGrpcDefault(transport, service)) {
       //      if (this.clientProperties.isUseRest()) {
```
meltsufin, line 390: Can this commented out code be removed?
(Also other places in this file.)

spring-cloud-generator/src/test/java/com/google/cloud/generator/spring/composer/goldens/EchoSpringAutoConfigurationRest.golden
```diff
@@ -0,0 +1,230 @@
...
+      }
+      this.credentialsProvider =
+          ((CredentialsProvider) new DefaultCredentialsProvider(this.clientProperties));
+    } else {
+      this.credentialsProvider = credentialsProvider;
+    }
+  }
+
+  /**
+   * Provides a default transport channel provider bean. The default is gRPC and will default to it
```
meltsufin, line 68: I guess this javadoc is no longer correct for REST-only or gRPC-only services.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringPropertiesClassComposer.java
```diff
@@ -158,7 +150,9 @@ private static List<Statement> createMemberVariables(
             "executorThreadCount", TypeNode.INT_OBJECT, false, null, null);
     statements.add(SpringPropertiesCommentComposer.createExecutorThreadCountPropertyComment());
     statements.add(executorThreadCountVarStatement);
-    if (hasRestOption) {
+    // For GRPC+REST libraries, provide configuration property to override GRPC default and use REST transport instead
+    // For GRPC-only and REST-only libraries, this property is not provided
+    if (ComposerUtils.shouldSupportRestOptionWithGrpcDefault(transport, service)) {
       ExprStatement useRestVarStatement =
           ComposerUtils.createMemberVarStatement(
               "useRest",
```
meltsufin, line 158: Have you considered a `transport` property with an enum value of "REST" or "GRPC" instead? I feel that would be more natural than `useRest`.

spring-cloud-generator/src/test/java/com/google/cloud/generator/spring/composer/goldens/EchoSpringAutoConfigurationRest.golden
emmileaf: Thanks for catching this! I missed updating the Javadoc generation logic here, will put in a fix.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringPropertiesClassComposer.java
emmileaf: Hmm, I can't recall if this was discussed when the `useRest` property was initially implemented, but two small complications I can see are:

1. This would be a breaking change for any existing users of this property (though now would be the time if it's a change we'd like to eventually make)
2. Possibly needing to document or account for case sensitivity in user-provided property strings? (e.g. REST vs. Rest)

If I'm understanding correctly, the user configuration to override would change from something like
```
<service.prefix>.useRest=true
```
to
```
<service.prefix>.transport=REST
```
This property is only exposed for libraries that support grpc+rest (and default to grpc), and having a property named `useRest` as an explicit override makes sense in my head as well, so I don't feel too strongly about either naming choice. Curious to hear any other concerns behind this suggestion though!

blakeli0:
> > Not part of this PR, I wonder if it makes sense to regenerate the auto configs as part of the CI if we have changes in SpringCodeGen, so that we can see the changes directly in the same PR.
> 
> @blakeli0 This suggestion actually makes me wonder - what would be the preferred timing to push any generated code changes introduced by changes to the generator? Currently I see the possibility for:
> 
> 1. As a commit added to the PR with generator changes (similar to your suggestion)
> 2. As a separate PR along with (or after) merging the generator changes
> 3. Not merged when introudcing generator changes, but will be included in the pre-release workflow in the PR to re-generate libraries as part of updating libraries-bom.
> 
> The status quo is 3, though this PR has been the first large generator feature change since initial release. Do we have any preference on what approach to take with this PR (which I will do manually), and which (if not the same) approach to follow in general? I think the possible CI/automation enhancements will look slightly different for each option.

We would always need 3 because the libraries bom could include new proto changes. In order for us to easily the generator changes, I think either 1 or 2 makes sense, otherwise all the generator changes and proto changes are in one huge PR before a new release. I would prefer 1 because it's easier to see all the generated changes in one PR, but 2 also has the advantage of seeing the generated changes separately so it's easier to review. This is not urgent, we can discuss it in next project meeting.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringAutoConfigClassComposer.java
meltsufin: I would add additional comments instead of the code.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringPropertiesClassComposer.java
meltsufin: We're able to make this kind of change because we're still in preview. I believe that Spring Boot's relaxed binding should be case-insensitive for enums.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringAutoConfigClassComposer.java
blakeli0: I would prefer to remove them or add more comment to indicate this commented out code is a sample of generated code, otherwise developers that are not familiar with the codebase can easily get confused. Another thing you can try is only adding the sample of generated code on top of the method, so it's less likely to get confused.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringPropertiesClassComposer.java
blakeli0: I do think explicitly specify`<service.prefix>.transport=REST` might be better in the long term, but there are more things we need to consider for this change. 
For example, what would be the default value if this property is not specified? `REST` for compute and `GRPC` for the others? What's the criteria for us to decide the default value? What if a customer configured `GRPC` for compute, are we going to error out on start up? 
In addition, are we going to support this property for existing spring-cloud-gcp products as well? Some of the service does not support `REST` yet or only support `REST`(bigquery), but I think some service like `spanner` already support both.

emmileaf: > Spring Boot's relaxed binding should be case-insensitive for enums

Oh that’s good to know! I can give this change a try in a separate PR.

> For example, what would be the default value if this property is not specified? REST for compute and GRPC for the others? What's the criteria for us to decide the default value? What if a customer configured GRPC for compute, are we going to error out on start up?

@blakeli0  My understanding of this suggested change is just in name/type (`useRest` boolean => `transport` enum) rather than the underlying behaviour: the current strategy is to only provide and support this property for libraries that support both GRPC+REST transport options. For GRPC and REST-only libraries where this property is not supported, any user configuration will not do anything (unless they override the transport provider and setting beans themselves), and the client library default builders will be used.

Would there be concern in this strategy itself, or is it more that having a renamed transport property only supported and applicable to some libraries could be confusing to users?

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringAutoConfigClassComposer.java
emmileaf: Makes sense, I see the source for confusion now. Cleaned up the comments in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849/commits/2628a462e8f5a00761231e2d3552be661f356ee3.

spring-cloud-generator/src/test/java/com/google/cloud/generator/spring/composer/goldens/EchoSpringAutoConfigurationRest.golden
emmileaf: Updated javadoc content in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849/commits/06b437cd44febaaa6592052bed9b643c4dfdb4e9.

spring-cloud-generator/src/main/java/com/google/cloud/generator/spring/composer/SpringPropertiesClassComposer.java
meltsufin: I think @blakeli0 brings up a good point about more things to consider. For instance, we can imagine a global `spring.cloud.gcp.transport` property that could switch all integrations to that transport. I almost feel like the property should be a list of desired transports. For instance, `GRPC,REST` would be the default and mean that choose REST only if GRPC is unavailable. Users, can still specify something like 'REST', which should fail with an unsupported-transport exception for libraries that only support GRPC.

blakeli0: I feel this is a relatively large topic that worth it's own discussion, can we leave it as it is for now? And put all the discussion around the naming together in a separate doc? We can always change it in preview but we should finalize it before GA.

emmileaf: For sure - I can put together a doc on this topic to further explore/discuss (similar to the other conversation on strategy to enable/disable configuration classes).

spring-cloud-generator/src/test/java/com/google/cloud/generator/spring/composer/SpringAutoConfigClassComposerTest.java
```diff
@@ -16,55 +16,78 @@
...
+  private Service echoRestProtoService;
   private Service wickedProtoService;
 
   @Before
   public void setUp() {
-    this.context = TestProtoLoader.instance().parseShowcaseEcho();
-    this.echoProtoService = this.context.services().get(0);
+    this.echoContext = TestProtoLoader.instance().parseShowcaseEcho();
+    this.echoProtoService = this.echoContext.services().get(0);
+    this.echoGrpcRestContext = GrpcRestTestProtoLoader.instance().parseShowcaseEcho();
```
emmileaf, line 47: Todo: this introduction of `GrpcRestTestProtoLoader` is the source of failing checks since switching to gapic-generator-java test protoloaders in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849/commits/f418b3e68bdff1ed4dc3b5b3e8b838bb08f7e918.

Options to fix:
1.  Switch `GRPC_REST` tests back to using `TestProtoLoader` and changing transport in parsed context
2. Copy over [`echo_grpcrest.proto`](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/test/proto/echo_grpcrest.proto), which seem to be what is used by `GrpcRestTestProtoLoader`.

blakeli0: Does the second option mean we have to keep an identical copy of `echo_grpcrest.proto` in spring-cloud-gcp? If that's the case, I would prefer the first option.

emmileaf: Agreed - I've updated the setup with the first option in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849/commits/0ff01b2a11f741c35f72a8b749c5d42e3137a119.

Trying to wrap my head around the variations of echo proto handling - I'm guessing there are different test proto loaders set up by transport in `gapic-generator-java` for use in its golden tests for transport-specific configs/implementations? So far spring composer tests using the echo proto will just need the `Transport` types to differ accordingly in the parsed `GapicContext`, so this setup is sufficient. In the future, if more test proto parsing logic start to diverge from gapic-generator-java's setup, we can consider placing them into a new `SpringTestProtoLoader` (maintained here).

emmileaf:
@blakeli0 @meltsufin  Thanks for your patience in reviewing this! I’ve run the code generation workflow on this feature branch to now include the generated code changes as well, but please let me know if you have preferences on whether to merge these together or separately.

I think this PR is otherwise ready for a re-review at your convenience. Quick summary of three items from discussions above that I will follow-up on, separately from this PR:
1. Alternative strategies for disabling/enabling autoconfiguration classes, and how these affect application start-up time ([discussion](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1850#issuecomment-1549882399))
2. Consider renaming and perhaps expanding configuration scope of the `<service.prefix>.useRest` property ([discussion](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849#discussion_r1201403836))
3. Workflow to push generated code changes introduced by generator changes as part of CI ([discussion](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849#issuecomment-1559746967))

meltsufin (approved):

spring-cloud-previews/google-cloud-compute-spring-starter/src/main/java/com/google/cloud/compute/v1/spring/AcceleratorTypesSpringAutoConfiguration.java
```diff
@@ -0,0 +1,223 @@
+/*
+ * Copyright 2022 Google LLC
```
blakeli0, line 2: Noticed the copyright year is still 2022, this is probably a bug in the generator though, similar to [what I did](https://github.com/googleapis/sdk-platform-java/pull/951) last year. Can you take a look?

emmileaf: Hmm yeah this part is being handled by the `gapic-generator-java` composers; I've opened https://github.com/googleapis/sdk-platform-java/pull/1720 for the corresponding fix. Thanks for catching!

emmileaf:
Reverted changes to generated code, and retriggering to include generator updates above in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/5245239436.

spring-cloud-previews/google-cloud-compute-spring-starter/src/main/java/com/google/cloud/compute/v1/spring/AcceleratorTypesSpringAutoConfiguration.java
emmileaf: License headers (note: for all generated modules) updated in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849/commits/34495a0c768097d5aa430b9a8fe8edc16f2eef21.

blakeli0 (approved):

emmileaf:
Merging this in, with corresponding release note message (through commit override) as `feat: add google-cloud-compute-spring-starter to starter modules available for preview`

PR #1850, opened 2023-05-15
feat: add starter module for compute to spring-cloud-previews

emmileaf:
Opening this PR to show expected code changes corresponding to generator changes in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849 
* From a local run of generator workflow, using monorepo tag v1.10.0
* Additional `library_list.txt` entry for compute: `compute, google/cloud/compute/v1, com.google.cloud:google-cloud-compute, f64a3c8418e2e4e493b6a432220d774e7942141a, java-compute`


meltsufin:
Wow! That's a lot of autoconfiguration classes that will be automatically enabled. It makes me wonder if we want to disable the autoconfiguration here by default.

emmileaf:
@meltsufin Hmm true... In comparison, the generated module with most autoconfiguration classes so far is [dialogflow](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/tree/main/spring-cloud-previews/google-cloud-dialogflow-spring-starter/src/main/java/com/google/cloud/dialogflow/v2/spring) at 17, compared to the 89 here for compute.

Perhaps we can make an exception for compute to have `com.google.cloud.compute.v1.<service-name>.enabled = false` by default, and add a section to explain this in the documentation.

meltsufin:
@emmileaf @zhumin8  Have you checked how the app startup time is impacted when we include the compute starter? I'm starting to think that maybe we should just disable autoconfiguration across all autogen modules and require the users to specify the enabled property correspending to what they want to use.

emmileaf:
> Have you checked how the app startup time is impacted when we include the compute starter?

Not yet, I'll look into this. 

> I'm starting to think that maybe we should just disable autoconfiguration across all autogen modules

Is the broader concern here for when the autogenerated starters eventually get added to the BOM, there could be a large overhead of autoconfiguration classes enabled that aren't used by the application?

meltsufin:
> > I'm starting to think that maybe we should just disable autoconfiguration across all autogen modules
> 
> Is the broader concern here for when the autogenerated starters eventually get added to the BOM, there could be a large overhead of autoconfiguration classes enabled that aren't used by the application?

Inclusion in the BOM won't have any impact because you still have to add the dependency explicitly. My concern is for modules like Compute where there are many unrelated clients in the same module and autoconfigure will load all of those beans (service clients) into the application context whether they are used or not. There could be a significant overhead incurred, and explicitly disabling all of the unneeded autocnfiguration classes in a module like Compute would be extremely tedious.

emmileaf:
@meltsufin Ah, that makes sense, thanks for explaning!

Checking with a very simple application against this branch, the startup time with/without pulling in the compute starter is 1.453s vs. 2.935s. 

Without compute starter (https://github.com/emmileaf/tidbits/commit/4758234bd4537008d5c0ec2e92724912c5577c2a):
```
INFO 40393 --- [           main] com.example.demo.DemoApplication         : Started DemoApplication in 1.453 seconds (process running for 1.744)
```
With compute starter, all autoconfigurations enabled by default (https://github.com/emmileaf/tidbits/commit/92f4c2c35d24f70e8c6cac20c171cf1ab1e312f7): 
```
INFO 40866 --- [           main] com.example.demo.DemoApplication         : Started DemoApplication in 2.935 seconds (process running for 3.367)
```

meltsufin:
@emmileaf I guess that doesn't seem too bad. I wonder how much memory is being used etc.

emmileaf:
Closing this PR in favour of having triggered workflow to push (updated) generated code changes directly to https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1849.

```
gh workflow run 'Generate Spring Auto-Configurations' \
--ref <feature-branch> \
-f branch_name=<feature-branch> \
-f forked_repo=none
```

Will follow-up on strategy to enable/disable autoconfigurations separately, as discussed offline.

PR #1853, opened 2023-05-15
[DRAFT] test: update sonar analysis exclusions

emmileaf:
Opening a PR to troubleshoot failing sonar checks on main: [details](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/runs/13494665939), after merging exclusion update from
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1816

Something about the glob pattern is not working as intended, and a recently added starter module in `spring-cloud-previews` is still picked up as in scope for analysis.

PR #1854, opened 2023-05-15
test: revert back to blanket sonar exclusion on spring-cloud-previews

emmileaf:
This PR reverts the sonar exclusion change from https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1816 to unblock CI.
 
I must have missed something in the exclusion glob pattern, and sonar analysis failed on main after merge ([details](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/runs/13494665939)). Seems like the analysis is picking up on code from generated modules (new in last 30 days: [google-cloud-dataproc-metastore-spring-starter](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/tree/main/spring-cloud-previews/google-cloud-dataproc-metastore-spring-starter)) when they intend to be excluded.

Going to verify this configuration out a bit more on a branch and retry the change (or possibly relocate the tests as suggested).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<integration-test.pattern>**/*IntegrationTest*</integration-test.pattern>
 		<!-- exclude preview module code from sonar cloud analysis -->
-		<sonar.exclusions>spring-cloud-previews/!(google-cloud-language-spring-starter)/**</sonar.exclusions>
+		<sonar.exclusions>spring-cloud-previews/**/*</sonar.exclusions>
 	</properties>
 
 	<dependencyManagement>


pom.xml
```diff
@@ -75,7 +75,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<integration-test.pattern>**/*IntegrationTest*</integration-test.pattern>
 		<!-- exclude preview module code from sonar cloud analysis -->
-		<sonar.exclusions>spring-cloud-previews/!(google-cloud-language-spring-starter)/**</sonar.exclusions>
```
meltsufin, line 78: Maybe you needed to escape the `-` with `\-`?

meltsufin (approved):

pom.xml
emmileaf: Oh possibly...my other suspicion is that maybe `/**` only matches directories, and `/**/*` is still needed to match files?

Retrying this change in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1853, but will merge this revert for now. Thanks!

Issue #1599, opened 2023-02-20
BigQueryTemplate.writeJsonStream API throwing java.lang.OutOfMemoryError: unable to create new native thread


pradeepnr:
**Describe the bug**

Problem : We are processing huge number of requests in the range of 100s of millions. Then pushing the records to Bigquery using the [BigQueryTempate::writeJsonStream API](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java#L253-L267). After running for more than 2 hours and calling the API 1000 times, it is throwing the below exception

```
java.lang.OutOfMemoryError: unable to create new native thread
at java.lang.Thread.start0(Native Method)
at java.lang.Thread.start(Thread.java:719)
at com.google.cloud.spring.bigquery.core.BigQueryTemplate.writeJsonStream(BigQueryTemplate.java:310)
...
...
```
**Library**
spring-cloud-gcp-starter-bigquery
**Bom**
```
<dependency>
    <groupId>com.google.cloud</groupId>
    <artifactId>spring-cloud-gcp-dependencies</artifactId>
    <version>3.4.0</version>
    <type>pom</type>
    <scope>import</scope>
</dependency>
```
**Details about the run**
We are receiving the messages from another system and processing it and writing it to BigQuery in batch of 5000.
After 700-800 API calls we are receiving the out of memory error.

mpeddada1:
Hi @pradeepnr Thank you filing this issue! A few questions to help us understand this better:

1) Does this issue also occur with the latest versions of spring-cloud-gcp? Including `3.4.4` and `4.1.0`.
2) Is this a recent occurrence? More specifically, is there a version of spring-cloud-gcp where the `OutOfMemoryError` doesn't occur?
3) Would it be possible to provide a reproducer that can help us investigate this further?

pradeepnr:
Hi @mpeddada1,
Please find my response below

1. I tired with only 3.4.0 version.
2. The issue was found during development stage.
3. I don't think, I will be able to do that. You can try running a for loop for 1000 times and calling the API to insert a dummy record to BQ.
[...]

PR #1855, opened 2023-05-16
fix: Wiring `bigQueryThreadPoolTaskScheduler ` with `writeJsonStream`

prash-mi:
This PR Wires `ThreadPoolTaskScheduler` with `writeJsonStream`. 

User can use `bigQueryThreadPoolTaskScheduler` in order to avoid `java.lang.OutOfMemoryError` which was arising due to too many concurrent write threads working. 

Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1599

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
@@ -29,6 +29,8 @@ import com.google.cloud.spring.core.UserAgentHeaderProvider;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
@@ -119,6 +121,12 @@ public class GcpBigQueryAutoConfiguration {
   }
 
   @Bean
+  @ConditionalOnMissingBean(name = "jsonWriterExecutorService")
+  public ExecutorService jsonWriterExecutorService() {
+    return Executors.newFixedThreadPool(threadPoolSize);
+  }
+
+  @Bean
   @ConditionalOnMissingBean(name = "bigQueryThreadPoolTaskScheduler")
   public ThreadPoolTaskScheduler bigQueryThreadPoolTaskScheduler() {
     ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
@@ -134,11 +142,16 @@ public class GcpBigQueryAutoConfiguration {
       BigQuery bigQuery,
       BigQueryWriteClient bigQueryWriteClient,
       @Qualifier("bigQueryThreadPoolTaskScheduler")
-          ThreadPoolTaskScheduler bigQueryThreadPoolTaskScheduler) {
+      ThreadPoolTaskScheduler bigQueryThreadPoolTaskScheduler,
+      @Qualifier("jsonWriterExecutorService") ExecutorService jsonWriterExecutorService) {
     Map<String, Object> bqInitSettings = new HashMap<>();
     bqInitSettings.put("DATASET_NAME", this.datasetName);
     bqInitSettings.put("JSON_WRITER_BATCH_SIZE", this.jsonWriterBatchSize);
     return new BigQueryTemplate(
-        bigQuery, bigQueryWriteClient, bqInitSettings, bigQueryThreadPoolTaskScheduler);
+        bigQuery,
+        bigQueryWriteClient,
+        bqInitSettings,
+        bigQueryThreadPoolTaskScheduler,
+        jsonWriterExecutorService);
   }
 }
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
@@ -147,6 +147,7 @@ public class GcpBigQueryAutoConfiguration {
     Map<String, Object> bqInitSettings = new HashMap<>();
     bqInitSettings.put("DATASET_NAME", this.datasetName);
     bqInitSettings.put("JSON_WRITER_BATCH_SIZE", this.jsonWriterBatchSize);
+    bqInitSettings.put("JSON_WRITER_THREAD_POOL_SIZE", this.threadPoolSize);
     return new BigQueryTemplate(
         bigQuery,
         bigQueryWriteClient,
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
@@ -138,7 +138,6 @@ public class GcpBigQueryAutoConfiguration {
     Map<String, Object> bqInitSettings = new HashMap<>();
     bqInitSettings.put("DATASET_NAME", this.datasetName);
     bqInitSettings.put("JSON_WRITER_BATCH_SIZE", this.jsonWriterBatchSize);
-    bqInitSettings.put("JSON_WRITER_THREAD_POOL_SIZE", this.threadPoolSize);
     return new BigQueryTemplate(
         bigQuery, bigQueryWriteClient, bqInitSettings, bigQueryThreadPoolTaskScheduler);
   }
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
@@ -29,8 +29,6 @@ import com.google.cloud.spring.core.UserAgentHeaderProvider;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
@@ -121,12 +119,6 @@ public class GcpBigQueryAutoConfiguration {
   }
 
   @Bean
-  @ConditionalOnMissingBean(name = "jsonWriterExecutorService")
-  public ExecutorService jsonWriterExecutorService() {
-    return Executors.newFixedThreadPool(threadPoolSize);
-  }
-
-  @Bean
   @ConditionalOnMissingBean(name = "bigQueryThreadPoolTaskScheduler")
   public ThreadPoolTaskScheduler bigQueryThreadPoolTaskScheduler() {
     ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
@@ -142,17 +134,12 @@ public class GcpBigQueryAutoConfiguration {
       BigQuery bigQuery,
       BigQueryWriteClient bigQueryWriteClient,
       @Qualifier("bigQueryThreadPoolTaskScheduler")
-      ThreadPoolTaskScheduler bigQueryThreadPoolTaskScheduler,
-      @Qualifier("jsonWriterExecutorService") ExecutorService jsonWriterExecutorService) {
+      ThreadPoolTaskScheduler bigQueryThreadPoolTaskScheduler) {
     Map<String, Object> bqInitSettings = new HashMap<>();
     bqInitSettings.put("DATASET_NAME", this.datasetName);
     bqInitSettings.put("JSON_WRITER_BATCH_SIZE", this.jsonWriterBatchSize);
     bqInitSettings.put("JSON_WRITER_THREAD_POOL_SIZE", this.threadPoolSize);
     return new BigQueryTemplate(
-        bigQuery,
-        bigQueryWriteClient,
-        bqInitSettings,
-        bigQueryThreadPoolTaskScheduler,
-        jsonWriterExecutorService);
+        bigQuery, bigQueryWriteClient, bqInitSettings, bigQueryThreadPoolTaskScheduler);
   }
 }
--- a/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
+++ b/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
@@ -319,7 +319,11 @@ public class BigQueryTemplate implements BigQueryOperations {
     writeApiFutureResponse.whenComplete(
         (writeApiResponse, exception) -> {
           if (exception != null || !writeApiResponse.isSuccessful()) {
-            logger.error("asyncTask interrupted");
+            if (exception != null) {
+              logger.error("asyncTask interrupted", exception);
+            } else {
+              logger.warn("Write operation failed");
+            }
             return;
           }
           logger.info("Data successfully written");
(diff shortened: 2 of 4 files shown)


spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -247,36 +255,33 @@ public Table createTable(
...
-                logger.warn(String.format("Error: %s %n", e.getMessage()), e);
-              } catch (Exception e) {
-                writeApiFutureResponse.completeExceptionally(e);
-                // Restore interrupted state in case of an InterruptedException
-                Thread.currentThread().interrupt();
-              }
-            });
-    asyncTask
-        .start(); // run the thread async so that we can return the writeApiFutureResponse. This
+    executorService.schedule(asyncTask, 0, TimeUnit.MICROSECONDS); // schedule the task with 0 delay
```
burkedavison, line 276: You might consider the slightly simpler syntax:
```java
CompletableFuture<WriteApiResponse> futureResponse = 
    CompletableFuture.supplyAsync(() -> { ... }, executorService);
```

spring-cloud-gcp-bigquery/src/test/java/com/google/cloud/spring/bigquery/BigQueryTemplateTest.java
```diff
@@ -223,6 +228,44 @@ void writeJsonStreamTest()
     assertEquals(0, apiRes.getErrors().size());
   }
 
+  @Test
+  void writeJsonStreamNegativeTest()
+      throws DescriptorValidationException, IOException, InterruptedException, ExecutionException {
+    InputStream jsonInputStream = new ByteArrayInputStream(newLineSeperatedJson.getBytes());
```
burkedavison, line 234: Should this (and the input stream in writeJsonStreamThrowTest) be in a try-with-resources block?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -86,6 +90,7 @@ public class BigQueryTemplate implements BigQueryOperations {
   private final Logger logger = LoggerFactory.getLogger(BigQueryTemplate.class);
 
   private final int jsonWriterBatchSize;
+  private final ScheduledExecutorService executorService;
```
meltsufin, line 93: Why not just `ExecutorService`? You're not using the scheduling features.
Also, the variable should probably be called `jsonWriterExecutorService` for clarity.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -115,6 +120,9 @@ public BigQueryTemplate(
     this.datasetName = bqDatasetName;
     this.taskScheduler = taskScheduler;
     this.bigQueryWriteClient = bigQueryWriteClient;
+    ThreadPoolTaskScheduler threadPoolTaskScheduler = (ThreadPoolTaskScheduler) taskScheduler;
```
meltsufin, line 123: `taskScheduler` is "used to poll for the status of long-running BigQuery operations". I don't think it should be re-purposed for json writing as well.
Instead, I would add a setter for `jsonWriterExecutorService` and wire in a default one from `BigQueryAutoConfiguration`.

meltsufin (changes requested):

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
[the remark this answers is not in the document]
prash-mi: Changed it to `ExecutorService` instead

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -247,36 +255,33 @@ public Table createTable(
...
 
-    CompletableFuture<WriteApiResponse> writeApiFutureResponse =
-        new CompletableFuture<>();
+    CompletableFuture<WriteApiResponse> writeApiFutureResponse = new CompletableFuture<>();
+    Runnable asyncTask =
+        () -> {
+          try {
+            WriteApiResponse apiResponse = getWriteApiResponse(tableName, jsonInputStream);
+            writeApiFutureResponse.complete(apiResponse);
+          } catch (DescriptorValidationException | IOException e) {
```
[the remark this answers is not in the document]
prash-mi: sure, done

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -247,36 +255,33 @@ public Table createTable(
...
+            writeApiFutureResponse.completeExceptionally(e);
+            Thread.currentThread().interrupt();
+            logger.warn(String.format("Error: %s %n", e.getMessage()), e);
```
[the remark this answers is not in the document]
prash-mi: sure, updated it

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -247,36 +255,33 @@ public Table createTable(
...
+    // register success and failure callback which simply logs the event
     writeApiFutureResponse.whenComplete(
         (writeApiResponse, exception) -> {
-          if (exception != null) {
-            asyncTask.interrupt(); // interrupt the thread as the developer might have cancelled the
-            // Future.
-            // This can be replaced with interrupting the ExecutorService when it has been wired-in
-            logger.info("asyncTask interrupted");
+          if (exception != null || !writeApiResponse.isSuccessful()) {
+            logger.error("asyncTask interrupted", exception);
```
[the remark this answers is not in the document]
prash-mi: thanks for pointing this, I tested and it does not work with `null` exception! I have removed the exception from `logger.error` and the developer can anyways get it using `Future.get`

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: Wired `ExecutorService` and replaced this with `ExecutorService.submit(...)`

spring-cloud-gcp-bigquery/src/test/java/com/google/cloud/spring/bigquery/BigQueryTemplateTest.java
prash-mi: It shouldn't cause problems as it's a stream out of a local resource, still, I have updated it with an outer try-with-resources block

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: sure, that's a better approach, I have updated the implementation accordingly, thanks

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -247,36 +253,28 @@ public Table createTable(
...
+    // register success and failure callback which simply logs the event
     writeApiFutureResponse.whenComplete(
         (writeApiResponse, exception) -> {
-          if (exception != null) {
-            asyncTask.interrupt(); // interrupt the thread as the developer might have cancelled the
-            // Future.
-            // This can be replaced with interrupting the ExecutorService when it has been wired-in
-            logger.info("asyncTask interrupted");
+          if (exception != null || !writeApiResponse.isSuccessful()) {
+            logger.error("asyncTask interrupted");
```
meltsufin, line 277: Does "asyncTask interrupted" still make sense? Why not just log the exception and/or the write response error message?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -96,12 +98,15 @@ public class BigQueryTemplate implements BigQueryOperations {
...
+   * @param jsonWriterExecutorService the {@link ExecutorService} used to run the thread required
+   *     for the BigQuery JSON Writer
    */
   public BigQueryTemplate(
       BigQuery bigQuery,
       BigQueryWriteClient bigQueryWriteClient,
       Map<String, Object> bqInitSettings,
-      TaskScheduler taskScheduler) {
+      TaskScheduler taskScheduler,
+      ExecutorService jsonWriterExecutorService) {
```
meltsufin, line 109: This is a breaking change. Either add a setter or an overloaded constructor.

meltsufin (changes requested):

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: My bad, I thought that we have updated the `bigQueryTemplate` bean @ `GcpBigQueryAutoConfiguration` , so we are safe from this breaking change! But you are right and a user might directly initialise `BigQueryTemplate`.

So, I have added  a overloaded constructor where the user need not pass `jsonWriterExecutorService `, added another setter so that they can override `jsonWriterThreadPoolSize`

meltsufin (changes requested):
@prash-mi Can we try re-using the `bigQueryThreadPoolTaskScheduler` instead of introducing a new `ExecutorService` specific to jsonWriter? I think it will be simpler and more clear.

prash-mi:
> 



> @prash-mi Can we try re-using the `bigQueryThreadPoolTaskScheduler` instead of introducing a new `ExecutorService` specific to jsonWriter? I think it will be simpler and more clear.

@meltsufin sure, I have reused `bigQueryThreadPoolTaskScheduler ` PTAL. 
Currently I have left `jsonWriterExecutorService` as is in the code as later we might be able to use it with any new capabilities we implement (with supporting documentation). Please let me know if it looks good, or if we should remove it from all the layers (`GcpBigQueryAutoConfiguration` and Unit Tests)

meltsufin (changes requested):
I was looking to have the `jsonWriterExecutorService` and everything associated with it to be removed entirely. Do you see a problem with that?

prash-mi:
> I was looking to have the `jsonWriterExecutorService` and everything associated with it to be removed entirely. Do you see a problem with that?

Done @meltsufin, I have removed `jsonWriterExecutorService` from all the layers

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
```diff
@@ -134,10 +134,11 @@ public BigQueryTemplate bigQueryTemplate(
       BigQuery bigQuery,
       BigQueryWriteClient bigQueryWriteClient,
       @Qualifier("bigQueryThreadPoolTaskScheduler")
-          ThreadPoolTaskScheduler bigQueryThreadPoolTaskScheduler) {
+      ThreadPoolTaskScheduler bigQueryThreadPoolTaskScheduler) {
     Map<String, Object> bqInitSettings = new HashMap<>();
     bqInitSettings.put("DATASET_NAME", this.datasetName);
     bqInitSettings.put("JSON_WRITER_BATCH_SIZE", this.jsonWriterBatchSize);
+    bqInitSettings.put("JSON_WRITER_THREAD_POOL_SIZE", this.threadPoolSize);
```
meltsufin, line 141: Is this still needed?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -81,6 +83,9 @@ public class BigQueryTemplate implements BigQueryOperations {
   private static final int DEFAULT_JSON_STREAM_WRITER_BATCH_SIZE =
       1000; // write records in batches of 1000
 
+  private static final int DEFAULT_JSON_WRITER_THREAD_POOL_SIZE =
```
meltsufin, line 86: Seems unused.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -247,36 +252,33 @@ public Table createTable(
...
+            writeApiFutureResponse.complete(apiResponse);
+          } catch (Exception e) {
+            writeApiFutureResponse.completeExceptionally(e);
+            // Restore interrupted state in case of an InterruptedException
+            Thread.currentThread().interrupt();
+            logger.warn("Unable to get write API response.", e);
+          }
+        };
+    ScheduledFuture<?> asyncTaskScheduledFuture =
+        taskScheduler.schedule(asyncTask, Instant.now()); // Run this task
```
meltsufin, line 269: Can you just call `taskScheduler.execute(asyncTask)`?

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -247,36 +252,33 @@ public Table createTable(
...
+              logger.error("asyncTask interrupted", exception);
+              if (exception instanceof CancellationException) { // user have cancelled it
+                asyncTaskScheduledFuture.cancel(true);
+              }
+            } else {
+              logger.warn("Write operation failed");
+            }
             return;
           }
           logger.info("Data successfully written");
```
meltsufin, line 284: Can you restructure to:
`if (exception != null) ... else if (!writeApiResponse.isSuccessful()) ... else logger.info("Data successfully written");`
?

meltsufin (changes requested):
Just final polish comments.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
prash-mi: No, it was used with the previous ExecutorService. Removed it

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: Removed

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: `TaskScheduler` doesn't expose `execute(...)`. Ref: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/TaskScheduler.html
Down-casting to a concrete implementation might be tricky, as the user might wire a different concrete implementation.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: Done

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
meltsufin: Ah true. I missed that.

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -247,39 +249,34 @@ public Table createTable(
...
-            // Future.
-            // This can be replaced with interrupting the ExecutorService when it has been wired-in
-            logger.info("asyncTask interrupted");
-            return;
+            logger.error("asyncTask interrupted", exception);
+            if (exception instanceof CancellationException) { // user have cancelled it
+              asyncTaskScheduledFuture.cancel(true);
+            }
+          } else if (writeApiResponse != null && !writeApiResponse.isSuccessful()) {
+            logger.warn("Write operation failed");
```
meltsufin, line 276: Would it make sense to log a message from the `writeApiResponse` here?

meltsufin (approved):

spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
prash-mi: It might be tricky to log it here as `WriteApiResponse` returns a `List<StorageError>` using `writeApiResponse.getErrors()` (as it's a batch operation),  users have access to `writeApiResponse`, so they might need to process it.

PR #1931, opened 2023-06-07
chore: adding profiles for native compile testings

zhumin8:
Based off https://github.com/mpeddada1/spring-cloud-gcp-current/pull/1, adding `spring-native` for parent pom, and `native-sample` for samples pom.
These added profile should not have any affect on existing workflows.
To be followed by #1933 .

Install and use GraalVM compiler and install native image extension with `gu install native-image`.
To package and run sample as native, instead of `mvn spring-boot:run`:
` mvn clean package -Pnative-sample -DskipTests` then execute the target built.
To run integration tests of sample as native, add relevant system property to `systemPropertyVariables` :
` mvn -Pnative-sample clean test`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -354,7 +354,7 @@
 			<dependency>
 				<groupId>org.junit.platform</groupId>
 				<artifactId>junit-platform-launcher</artifactId>
-				<version>1.9.2</version>
+				<version>1.9.3</version>
 				<scope>test</scope>
 			</dependency>
 		</dependencies>
@@ -363,13 +363,13 @@
 				<plugin>
 					<groupId>org.apache.maven.plugins</groupId>
 					<artifactId>maven-surefire-plugin</artifactId>
-					<version>3.0.0</version>
+					<version>${maven-surefire-plugin.version}</version>
 					<configuration>
 						<!-- https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#integration-tests.no-starter-parent -->
 						<excludes combine.self="override" />
 						<includes>
 							<!--including all integration tests for testing purposes. -->
-							<include>**/*IntegrationTests.java</include>
+							<include>${integration-test.pattern}</include>
 						</includes>
 						<classesDirectory>${project.build.outputDirectory}</classesDirectory>
 						<systemPropertyVariables>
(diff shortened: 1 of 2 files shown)


meltsufin (approved):

spring-cloud-gcp-samples/pom.xml
```diff
@@ -78,6 +78,82 @@
...
+		<profile>
+			<id>native-sample</id>
+			<dependencies>
+				<dependency>
+					<groupId>org.junit.platform</groupId>
+					<artifactId>junit-platform-launcher</artifactId>
+					<version>1.9.3</version>
+					<scope>test</scope>
+				</dependency>
+			</dependencies>
```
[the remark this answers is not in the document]
zhumin8: Actually, `org.springframework.boot:spring-boot-maven-plugin` is already included in `native`/`nativeTest` profiles [here](https://github.com/spring-projects/spring-boot/blob/a9ff0f0ca3aa3f5f39c57097a3ccc6ab6c44d458/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle#L248-L249)

pom.xml
```diff
@@ -336,10 +336,88 @@
 					</execution>
 				</executions>
 			</plugin>
+
+			<plugin>
+				<groupId>org.graalvm.buildtools</groupId>
+				<artifactId>native-maven-plugin</artifactId>
+				<version>0.9.20</version>
```
mpeddada1, line 343: How about we store this version in a property? Looks like we use it in a couple of places in this file.

mpeddada1 (approved):
Looks great!

pom.xml
zhumin8: Done! 
Also removed explicit declare of `spring-boot-maven-plugin` version as it's inherited from spring-boot-dependencies and keeps in sync with spring boot versions. (ref [here](https://github.com/spring-projects/spring-boot/blob/a9ff0f0ca3aa3f5f39c57097a3ccc6ab6c44d458/spring-boot-project/spring-boot-dependencies/build.gradle#L1308))
As for `junit-platform-launcher` version, kept as is because according to [doc](https://graalvm.github.io/native-build-tools/latest/maven-plugin.html#testing-support-version-compatibility), we should be able to remove it after we upgrade Maven Surefire plugin to 3.1.2 (#1936)

PR #1933, opened 2023-06-07
fix: add runtime hints for logging

zhumin8:
~~Need rebase after setups in #1931 are merged in.~~

This pr adds:
- runtime hints for springframework aot to allow Logging module work with Native build.
- a relevant unit test
- runtime hints needed for sample integration test. (this is test setup specific and is not needed to run sample application).

### CI for native is setup to run on sample integration tests that:
- has `it.xx` property added to [systemPropertyVariables](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/f11fde736c783c4334ae50cb31e8ccf879738daa/spring-cloud-gcp-samples/pom.xml#L133) 
- module added to [native-sample-config](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/f11fde736c783c4334ae50cb31e8ccf879738daa/spring-cloud-gcp-samples/pom.xml#L92C2-L92C2) profile.

Modification to `spring-cloud-gcp-samples/pom.xml` so that:
- by default, all modules are included.
- To run native test only for selected modules, use `--define notAllModules=true` in mvn cmd.

This is needed because native test would fail on no test configuration found.

### To verify locally:
- Run the sample manually:
   - Follow guide in README, instead of `mvn spring-boot:run`, package with `mvn clean package -Pnative-sample-config -Pnative`
   - Run executable `./target/spring-cloud-gcp-logging-sample `
   - Follow guide to issue a request and observe log lines from cloud console.
   - Log entry written by sample:
```
{
insertId: "1c08xcxf1v07m9"
jsonPayload: {
message: "This line was also written to the log with the same Trace ID."
}
labels: {3}
logName: "projects/[project-id]/logs/spring.log"
receiveTimestamp: "2023-06-06T14:15:08.236653291Z"
resource: {2}
severity: "INFO"
timestamp: "2023-06-06T14:15:08.122Z"
trace: "projects/[project-id]/traces/trace-id-to-show-1234"
}
``` 
- Run sample test:
    - `mvn -PnativeTest clean test -Pnative-sample-config`
    - native test are run at the end of the build.
```
com.example.LoggingSampleApplicationIntegrationTests > testLogRecordedInStackDriver() SUCCESSFUL


Test run finished after 12975 ms
[         3 containers found      ]
[         0 containers skipped    ]
[         3 containers started    ]
[         0 containers aborted    ]
[         3 containers successful ]
[         0 containers failed     ]
[         1 tests found           ]
[         0 tests skipped         ]
[         1 tests started         ]
[         0 tests aborted         ]
[         1 tests successful      ]
[         0 tests failed          ]

```


zhumin8:
force-pushed to rebase on main.

.github/workflows/NativeTests.yaml
```diff
@@ -0,0 +1,121 @@
...
+    paths-ignore:
+    - 'spring-cloud-generator/**'
+    - 'spring-cloud-previews/**'
+  workflow_dispatch:
+
+
+jobs:
+  NativeTests:
+    if: |
+      github.actor != 'dependabot[bot]' && ((
```
burkedavison, line 18: Is there a reason for this limitation? I would think we want to run native tests when updating dependencies.

.github/workflows/NativeTests.yaml
```diff
@@ -0,0 +1,121 @@
...
+          credentials_json: ${{ secrets.SPRING_CLOUD_GCP_CI_NATIVE_SA_KEY }}
+      - name: Setup gcloud
+        uses: google-github-actions/setup-gcloud@v1
+        with:
+          project_id: spring-cloud-gcp-ci-native
+      - name: Install pubsub-emulator
+        if: ${{ matrix.it == 'pubsub-emulator' }}
+        run: |
+          gcloud components install pubsub-emulator beta && \
+            gcloud components update
```
burkedavison, line 45: I think `matrix.it` is always null. Can this step be removed?

spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/aot/LoggingRuntimeHints.java
```diff
@@ -0,0 +1,43 @@
...
+            Arrays.asList(
+                TypeReference.of(LoggingAppender.class),
+                TypeReference.of(TraceIdLoggingEnhancer.class),
+                TypeReference.of(StackdriverJsonLayout.class)),
+            hint ->
+                hint.withMembers(
+                    MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
+                    MemberCategory.INVOKE_PUBLIC_METHODS));
+  }
+}
```
mpeddada1, line 43: Out of curiosity, do we notice any advantages with using the RuntimeHintsRegistrar class over using static reflection configurations? 
Mostly in terms of choosing the static jsons vs a dynamic way of registering classes for reflection, I'm wondering if we will run into any breaking changes when we upgrade our Spring stack that introduces some changes to it's class definition.

.github/workflows/NativeTests.yaml
zhumin8: will do.

.github/workflows/NativeTests.yaml
zhumin8: AFAIK, we exclude dependency upgrades or any other pr from outside for security reasons.

spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/aot/LoggingRuntimeHints.java
zhumin8: Good question! I went this way because it's [documented](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#native-image.advanced.custom-hints) and to be consistent to spring conventions, I notice spring-boot and spring-data projects are using this `RuntimeHintsRegistar` (e.g. [logging autoconfig](https://github.com/spring-projects/spring-boot/blob/6176eaa0189ee71638693079eb35fcd9dbb0d01f/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackRuntimeHints.java))
Some other benefit include: this allows hints be tested; It's easy use spring concepts when registering hints (for example in [vision](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/5375e0c554b1c3c737e41889615c53478d58f6b4/spring-cloud-gcp-vision/src/test/java/com/google/cloud/spring/vision/nativeimage/TestRuntimeHints.java#L28-L29), this is for the test resources, but you can easily refer to spring resources when registering.)

burkedavison (approved):

spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/aot/LoggingRuntimeHints.java
mpeddada1: Thanks for the context! Considering this resolved to unblock this PR. Discussion has been moved offline.

mpeddada1 (approved):

PR #1956, opened 2023-06-15
chore: spring-cloud-generator script refactor

emmileaf:
This PR refactors spring-cloud-generator scripts, with the following goals:
* Small fix for pattern-matching when sorting README table entries (follow-up on [1529-comment](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1529#discussion_r1088179838))
* Post-migration follow-up: simplify amount of script processing and files
  - Different piecewise scripts (e.g. `generate-one.sh`, `setup-build-rule.sh`, ...) consolidated into functions living in `generate-steps.sh`
  - Main script of `generate.sh` contains end-to-end workflow that invokes the various functions defined above
  - This also sets up pieces of the generation generation steps for potential reuse / reducing duplication in the showcase testing generation script ([#1957](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1957)), though related changes are not made in this PR to keep the refactoring incremental.

- [x] Adjustments to github actions workflow using refactored scripts tested in ([workflow run](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/5283179730)) - only expected change pushed is tweak to README from template change (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1956/commits/70e22abebae923894de1e3560bbcc23f5a234442)


spring-cloud-generator/scripts/generate-steps.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash
+
+# Cleans the spring-cloud-previews folder and replaces its contents with an
+# empty setup (i.e. empty readme, no libraries, pom without modules)
+function reset_previews_folder() {
+  cd ${SPRING_GENERATOR_DIR}
```
diegomarquezp, line 6: nit: we could avoid assuming `SPRING_GENERATOR_DIR` is set by relying on a function in this same file. Maybe it can fail if it's not set instead of defaulting to `pwd`, or maybe add a main comment section at the beginning of the file explaining that it needs to be set

spring-cloud-generator/scripts/generate-steps.sh
```diff
@@ -0,0 +1,195 @@
...
+    | xargs bazelisk build --tool_java_language_version=17 --tool_java_runtime_version=remotejdk_17 2>&1 \
+    | tee tmp-output || save_error_info "bazel_build"
+}
+
+# Add module to aggregator POM in the appropriate location, with check for existence.
+#
+# $1 - path-to-pom-file (e.g. path/to/pom.xml)
+# $2 - starter-artifact-id (e.g. google-cloud-accessapproval-spring-starter)
+function add_module_to_pom () {
+  xmllint --debug --nsclean --xpath  "//*[local-name()='module']/text()" $1 \
```
diegomarquezp, line 120: nit: for consistency among functions, we can declare a variable containing `$1` like others do (`path_to_pom=$1`)

diegomarquezp (approved):
LGTM overall, just a couple nits. Thanks for tidying this up!

spring-cloud-generator/scripts/generate-library-list.sh
```diff
@@ -10,20 +13,17 @@ do
 done
 echo "Monorepo tag: $commitish";
 
-# install jq for json parsing if not already installed
-sudo apt-get -y install jq
```
diegomarquezp, line 14: does `jq` come by default in GH actions?

meltsufin (approved):

PR #1999, opened 2023-07-05
chore: fix spring-cloud-generator scripts after refactor

emmileaf:
Fixing a miss from script refactor in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1956:

Invoking `scripts/generate-library-list.sh` missed a flag for passing the `google-cloud-java` version to checkout

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/8fafffde74da07e47cd1248c9d22058c92d63256/spring-cloud-generator/scripts/generate-library-list.sh#L20-L21

`git checkout` allows for empty argument, so this did not fail but would result in incorrectly picked up changes from monorepo head, e.g. 30117afda623c1fd203fb0f2e7919b37e086aa73.

Corrections in this PR:
* Fixes invocation with flag
* Adds check for monorepo commitish to not be empty

### DIFF
--- a/spring-cloud-generator/scripts/generate-library-list.sh
+++ b/spring-cloud-generator/scripts/generate-library-list.sh
@@ -13,6 +13,11 @@ do
 done
 echo "Monorepo tag: $commitish";
 
+if [[ -z "$commitish" ]]; then
+  echo "Missing google-cloud-java commitish to checkout"
+  exit 1
+fi
+
 # download the monorepo, need to loop through metadata there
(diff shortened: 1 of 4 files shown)


meltsufin (approved):

emmileaf:
Verifying with workflow run in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/5467405496/jobs/9953665888 - will merge when confirmed.

Issue #1996, opened 2023-07-05
Precision of timestampNanos in logs is miliseconds instead of nanoseconds


rafal-dudek:
**Describe the bug**
StackdriverJsonLayout sets value of 'timestampNanos' log field as 
`TimeUnit.MILLISECONDS.toNanos(event.getTimeStamp() % 1_000)`  which has precision of miliseconds.
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/StackdriverJsonLayout.java#L276
Proper value should be `event.getNanoseconds()`

This is causing out-of-order logs in GCP Logs Explorer in some cases.

**Sample**
```
    Logger log = LoggerFactory.getLogger(GenericSpringBootApplication.class);
    log.info("1");
    log.info("2");
    log.info("3");
    log.info("4");
    log.info("5");
```
Result in kubectl logs:

> {"timestampSeconds":1688547662,"timestampNanos":662000000,"severity":"INFO","thread":"scheduling-1","message":"1"}
{"timestampSeconds":1688547662,"timestampNanos":663000000,"severity":"INFO","thread":"scheduling-1","message":"2"}
{"timestampSeconds":1688547662,"timestampNanos":663000000,"severity":"INFO","thread":"scheduling-1","message":"3"}
{"timestampSeconds":1688547662,"timestampNanos":663000000,"severity":"INFO","thread":"scheduling-1","message":"4"}
{"timestampSeconds":1688547662,"timestampNanos":663000000,"severity":"INFO","thread":"scheduling-1","message":"5"}

Result in GCP Logs Explorer:
![image](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/assets/32459286/86346ebf-20fd-4a66-8c5f-33560c5267b1)

meltsufin:
Would you mind submitted a fix?

PR #2003, opened 2023-07-06
fix: timestampNanos nanos precision.

rafal-dudek:
Fix https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1996
I hope that testTimestampNanos() is enough. I am testing 2 logs to lower false negative possibility (when all micros and nanos are 0) to 1:1e+12

### DIFF
--- a/spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/StackdriverJsonLayout.java
+++ b/spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/StackdriverJsonLayout.java
@@ -271,9 +271,10 @@ public class StackdriverJsonLayout extends JsonLayout {
       map.put(
           StackdriverTraceConstants.TIMESTAMP_SECONDS_ATTRIBUTE,
           TimeUnit.MILLISECONDS.toSeconds(event.getTimeStamp()));
+      int nanoseconds = event.getNanoseconds();
       map.put(
           StackdriverTraceConstants.TIMESTAMP_NANOS_ATTRIBUTE,
-          TimeUnit.MILLISECONDS.toNanos(event.getTimeStamp() % 1_000));
+          nanoseconds == -1 ? TimeUnit.MILLISECONDS.toNanos(event.getTimeStamp() % 1_000) : nanoseconds);
     }
 
     add(
--- a/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/StackdriverJsonLayoutLoggerTests.java
+++ b/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/StackdriverJsonLayoutLoggerTests.java
@@ -247,6 +247,22 @@ class StackdriverJsonLayoutLoggerTests {
   }
 
   @Test
+  void testTimestampNanos() {
+    LOGGER.warn("test1");
+    LOGGER.warn("test2");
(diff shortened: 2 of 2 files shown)


rafal-dudek:
Working on CLA

spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/StackdriverJsonLayout.java
```diff
@@ -273,7 +273,7 @@ protected Map<String, Object> toJsonMap(ILoggingEvent event) {
           TimeUnit.MILLISECONDS.toSeconds(event.getTimeStamp()));
       map.put(
           StackdriverTraceConstants.TIMESTAMP_NANOS_ATTRIBUTE,
-          TimeUnit.MILLISECONDS.toNanos(event.getTimeStamp() % 1_000));
+          event.getNanoseconds());
```
meltsufin, line 276: According to [documentation](https://logback.qos.ch/apidocs/ch/qos/logback/classic/spi/ILoggingEvent.html#getNanoseconds()), `getNanoseconds()` may return `-1`. We need to handle that case and default to the old method.

rafal-dudek: Good point. Standard LoggingEvent from logback has it but there may be other implementations without it.

spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/StackdriverJsonLayoutLoggerTests.java
```diff
@@ -246,6 +246,22 @@ void testJsonSeverityLevelMapping() {
...
+
+    List<String> jsonLogRecords = Arrays.asList(new String(logOutput.toByteArray()).split("\n"));
+
+    List<Double> logTimestampNanos =
+            jsonLogRecords.stream()
+                    .map(record -> GSON.fromJson(record, Map.class))
+                    .map(data -> (Double) data.get(StackdriverTraceConstants.TIMESTAMP_NANOS_ATTRIBUTE))
+                    .collect(Collectors.toList());
+
+    assertThat(logTimestampNanos).anySatisfy(nanos -> assertThat(nanos % 1000000).isGreaterThan(0));
```
meltsufin, line 262: Can't nanos be `0` though?

rafal-dudek: Yes, I wanted to create test to ensure that there is nano precision and not just mili.
I am testing 2 logs with any greater than 0 to lower false negative possibility (when all micros and nanos are 0) to 1:1e+12.

Not sure if you like it this way or maybe just simple 1 log with nanos greaterOrEqual(0).

meltsufin: I think the way you have it is fine. As you point out, the likelihood of flakiness due to this is tiny. You might just want to make sure you are getting 2 timestamps in the array. You can even make the number of logs a variable.

rafal-dudek:
Sorry, my company can't sign the CLA. Closing.

PR #2012, opened 2023-07-08
fix: timestampNanos field with nanos precision.

Raf0:
Fix https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1996
@meltsufin


meltsufin:
Will you be able to sign the CLA as an individual contributor?

PR #2068, opened 2023-07-27
chore: enable sub jobs for native image modules

mpeddada1:


mpeddada1:
Stacktrace:
```
[INFO] Running com.example.LoggingSampleApplicationIntegrationTests
Warning:  Tests run: 1, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 0.01 s - in com.example.LoggingSampleApplicationIntegrationTests
[INFO] 
[INFO] Results:
[INFO] 
Warning:  Tests run: 1, Failures: 0, Errors: 0, Skipped: 1
[INFO] 
[INFO] 
[INFO] --- native:0.9.23:test (native-test) @ spring-cloud-gcp-logging-sample ---
Error:  Test configuration file wasn't found. Make sure that test execution wasn't skipped.
```

mpeddada1:
Native image tests for modules being split into sub-jobs as expected. Next, apply same changes for sample tests.

spring-cloud-gcp-samples/pom.xml
```diff
@@ -95,6 +95,7 @@
 				<module>spring-cloud-gcp-integration-storage-sample</module>
 				<module>spring-cloud-gcp-trace-sample</module>
 				<module>spring-cloud-gcp-vision-api-sample</module>
+				<module>spring-cloud-gcp-vision-ocr-demo</module>
```
burkedavison, line 98: Is this "-demo" suffix a distinct concept from "-sample"s? Is it being handled the way you expect w/ native-image-tests.sh?

.github/workflows/scripts/native-image-tests.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# This script executes native image tests based on the MODULE_UNDER_TEST provided.
+# If MODULE_UNDER_TEST is 'vision', for example, then it runs tests under spring-cloud-gcp-vision.
+# If it is 'vision-sample' then it runs tests under spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample
+# and spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo.
+
+# Native image support has not been added for this module yet.
+EXCLUDED_MODULES=("spring-cloud-spanner-spring-data-r2dbc")
```
burkedavison, line 9: Seems unfortunate to need to define this here.

spring-cloud-gcp-samples/pom.xml
mpeddada1: Comparing the code and the READMEs, there doesn't seem to be a distinction between the two. This is the only sample that has the suffix `demo` so it may be a good idea to rename this `sample` to follow convention. cc/ @meltsufin

.github/workflows/scripts/native-image-tests.sh
mpeddada1: As an alternative option, perhaps this could be passed in as an argument to the script?

mpeddada1: Updated the script to accept this value as an argument to make it less hidden. Let me know what you think?

burkedavison (approved):

PR #2127, opened 2023-08-16
test: add debug logs and check for empty directory

alicejli:


meltsufin:
@alicejli Should this be a draft PR?

PR #2164, opened 2023-09-13
feat: add support for Firestore database id configuration

### LINKED ISSUE #2145: Add support for specifying Firestore/Datastore DBs
It is currently quite difficult to contain a Firestore/Datastore database to a single microservice. Currently the only solution is to use the (default) database and ensure microservices do not use the same collection names.

GCP has recently added [support](https://cloud.google.com/firestore/docs/manage-databases#create_a_database) for multiple databases, meaning it's now possible to assign one per microservice. This could possibly be implemented by creating an optional system property such as `spring.cloud.gcp.firestore.database = "xyz"`

This could also possibly be implemented by adding the field to a Document annotation such as `@Document(collectionName="users", databaseName="xyz")`

Thanks!

meltsufin:
Fixes #2145.

### DIFF
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -68,7 +68,7 @@ The following configuration options are available:
 | Name | Description | Required | Default value
 | `spring.cloud.gcp.datastore.enabled` | Enables the Cloud Datastore client | No | `true`
 | `spring.cloud.gcp.datastore.project-id` | Google Cloud project ID where the Google Cloud Datastore API is hosted, if different from the one in the <<spring-cloud-gcp-core,Spring Framework on Google Cloud Core Module>> | No |
-| `spring.cloud.gcp.datastore.database-id` | Google Cloud project can host multiple databases. You can specify which database will be used. | No |
+| `spring.cloud.gcp.datastore.database-id` | Google Cloud project can host multiple databases. You can specify which database will be used. If not specified, the database id will be "(default)". | No |
 | `spring.cloud.gcp.datastore.credentials.location` | OAuth2 credentials for authenticating with the Google Cloud Datastore API, if different from the ones in the <<spring-cloud-gcp-core,Spring Framework on Google Cloud Core Module>> | No |
 | `spring.cloud.gcp.datastore.credentials.encoded-key` | Base64-encoded OAuth2 credentials for authenticating with the Google Cloud Datastore API, if different from the ones in the <<spring-cloud-gcp-core,Spring Framework on Google Cloud Core Module>> | No |
 | `spring.cloud.gcp.datastore.credentials.scopes` | https://developers.google.com/identity/protocols/googlescopes[OAuth2 scope] for Spring Framework on Google CloudDatastore credentials | No | https://www.googleapis.com/auth/datastore
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
@@ -61,6 +61,8 @@ public class GcpFirestoreAutoConfiguration {
 
   private final String projectId;
 
+  private final String databaseId;
+
   private final CredentialsProvider credentialsProvider;
 
   private final String hostPort;
@@ -74,6 +76,7 @@ public class GcpFirestoreAutoConfiguration {
       throws IOException {
 
     this.projectId = gcpFirestoreProperties.getResolvedProjectId(projectIdProvider);
+    this.databaseId = gcpFirestoreProperties.getDatabaseId();
 
     if (gcpFirestoreProperties.getEmulator().isEnabled()) {
       // if the emulator is enabled, create CredentialsProvider for this particular case.
@@ -92,13 +95,19 @@ public class GcpFirestoreAutoConfiguration {
   @Bean
   @ConditionalOnMissingBean
   public FirestoreOptions firestoreOptions() {
-    return FirestoreOptions.getDefaultInstance().toBuilder()
+
+    FirestoreOptions.Builder builder = FirestoreOptions.getDefaultInstance().toBuilder()
         .setCredentialsProvider(this.credentialsProvider)
         .setProjectId(this.projectId)
         .setHeaderProvider(USER_AGENT_HEADER_PROVIDER)
         .setChannelProvider(
-            InstantiatingGrpcChannelProvider.newBuilder().setEndpoint(this.hostPort).build())
-        .build();
+            InstantiatingGrpcChannelProvider.newBuilder().setEndpoint(this.hostPort).build());
+
+    if (databaseId != null) {
+      builder.setDatabaseId(databaseId);
+    }
+
+    return builder.build();
   }
 
   @Bean
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreProperties.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreProperties.java
@@ -40,6 +40,8 @@ public class GcpFirestoreProperties implements CredentialsSupplier {
 
   private String projectId;
 
+  private String databaseId;
+
   /**
    * The host and port of the Firestore emulator service; can be overridden to specify an emulator.
    */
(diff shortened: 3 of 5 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreAutoConfigurationTests.java
```diff
@@ -60,6 +60,7 @@ void testDatastoreOptionsCorrectlySet() {
         context -> {
           FirestoreOptions datastoreOptions = context.getBean(Firestore.class).getOptions();
           assertThat(datastoreOptions.getProjectId()).isEqualTo("test-project");
+          assertThat(datastoreOptions.getDatabaseId()).isEqualTo("(default)");
```
JoeWang1127, line 63: why there's a pair of parentheses?

meltsufin: That's just the name of the default database.
See: https://firebase.google.com/docs/firestore/manage-databases#access_a_named_database_with_a_client_library

suztomo:
@diegomarquezp I found the error message your pull request should fix "Caused by: java.sql.SQLException: Access denied for user 'root'@'cloudsqlproxy~20.49.35.21' (using password: YES)"

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/6177169510/job/16767718916?pr=2164

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -92,13 +95,19 @@ public class GcpFirestoreAutoConfiguration {
...
+    FirestoreOptions.Builder builder = FirestoreOptions.getDefaultInstance().toBuilder()
         .setCredentialsProvider(this.credentialsProvider)
         .setProjectId(this.projectId)
         .setHeaderProvider(USER_AGENT_HEADER_PROVIDER)
         .setChannelProvider(
-            InstantiatingGrpcChannelProvider.newBuilder().setEndpoint(this.hostPort).build())
-        .build();
+            InstantiatingGrpcChannelProvider.newBuilder().setEndpoint(this.hostPort).build());
+
+    if (databaseId != null) {
```
blakeli0, line 106: Do we want to check for empty String as well?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreProperties.java
```diff
@@ -82,7 +92,8 @@ public void setEmulator(FirestoreEmulatorProperties emulator) {
   }
 
   public String getFirestoreRootPath(GcpProjectIdProvider projectIdProvider) {
-    return String.format(ROOT_PATH_FORMAT, this.getResolvedProjectId(projectIdProvider));
+    return String.format(ROOT_PATH_FORMAT, this.getResolvedProjectId(projectIdProvider),
+        this.getDatabaseId() == null ? "(default)" : this.getDatabaseId());
```
blakeli0, line 96: Nice to have: This can be extracted to a method `getResolvedDatabaseId()`, similar to `getResolvedProjectId()`.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreAutoConfigurationTests.java
```diff
@@ -95,6 +96,20 @@ void testTransactionManagerExcludedWithoutAutoConfiguration() {
             });
   }
 
+  @Test
+  void testDatabaseId() {
```
blakeli0, line 100: Maybe this test can be named a little more specific? The above test is testing the default behavior, so this one can be something like `testDatabaseIdOverride`. 
See also [unit test naming](https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#naming), I know we don't have this naming convention in Spring Cloud GCP yet, so we don't have to strictly follow it, the idea is good though to include the behavior and outcome in the test name.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
meltsufin: I don't think it's needed. We check for null across the board in this project in scenarios like this.

meltsufin:
```
Error:  Errors: 
Error:    FirestoreSampleApplicationIntegrationTests.saveUserTest » StatusRuntime INVALID_ARGUMENT: Missing routing header. Please fill in the request header with format x-goog-request-params:project_id=spring-cloud-gcp-ci&database_id=firestoredb. Please refer to https://firebase.google.com/docs/firestore/manage-databases#access_a_named_database_with_a_client_library for more details.
```

@blakeli0 Do I need a newer version of GAX or Firestore client lib?

lqiu96:
Iirc, Firestore temporarily removed code for gRPC's Dynamic Routing Headers a while back (due to an encoding issue). I think they only recently added that back in (https://github.com/googleapis/java-firestore/pull/1418) and I'm think this would work with Firestore v3.14.3+

blakeli0:
As Lawrence mentioned, they recently added the routing headers back and Spring Cloud GCP didn't get it yet. But if this is the root cause, it means that the server made a behavior breaking change that requires the customers to update their client code, which is very bad.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -150,12 +160,27 @@ public FirestoreTemplate firestoreTemplate(
...
     }
 
+    @Bean
+    @ConditionalOnMissingBean(name = "firestoreRoutingHeadersInterceptor")
+    public ClientInterceptor firestoreRoutingHeadersInterceptor() {
+      // add routing header for custom database id
+      Metadata routingHeader = new Metadata();
+      Metadata.Key<String> key =
+          Metadata.Key.of(Headers.DYNAMIC_ROUTING_HEADER_KEY, Metadata.ASCII_STRING_MARSHALLER);
+      routingHeader.put(key,
```
blakeli0, line 170: This may work for this specific case, but routing header config could change and I'm not sure it works for all the new [routing header configs](https://github.com/googleapis/java-firestore/pull/1418/files#diff-8c7c9e5aae7297f519f9531432656cb5c8eaa1548796e5c60a091252a3cbc24eR301-R463), or maybe those are not required. 
In long term, switching to GAPIC stubs should solve this issue, but I'm not sure it is compatible with the Reactive paradigm. Why did we choose to use the gRPC generated stubs instead of GAPIC ones? Only because of Reactive?

blakeli0: I see. My concern is that every time the backend makes the routing header required for a new RPC, or change how the routing header is constructed, we would have to come back and manually change it in spring-cloud-gcp. It may not in scope of this PR, can we investigate the possibility of switching to the GAPIC stubs?

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -68,7 +68,7 @@ The following configuration options are available:
 | Name | Description | Required | Default value
 | `spring.cloud.gcp.datastore.enabled` | Enables the Cloud Datastore client | No | `true`
 | `spring.cloud.gcp.datastore.project-id` | Google Cloud project ID where the Google Cloud Datastore API is hosted, if different from the one in the <<spring-cloud-gcp-core,Spring Framework on Google Cloud Core Module>> | No |
-| `spring.cloud.gcp.datastore.database-id` | Google Cloud project can host multiple databases. You can specify which database will be used. | No |
+| `spring.cloud.gcp.datastore.database-id` | Google Cloud project can host multiple databases. You can specify which database will be used. If not specified, the database id will be "(default)". | No |
```
suztomo, line 71: "Default value" column is missing.

suztomo (approved):

thoastbrot:
Hi @meltsufin, this breaks my (test) configuration where I only define service credentials via spring.cloud.gcp.firestore.credentials.encoded-key - this contains the project ID, and somehow this worked till 4.7.2, but crashes here, as the percent escaper receives projectId null. Do you have an idea how this _should_ work?

meltsufin:
@thoastbrot I guess it's because your project ID is `null` in the configuration. You can configure a dummy project ID for now as a workaround. Would you mind filing a separate issue?

thoastbrot:
I'll do that. dummy project-id doesn't work the way you assume.

Issue #2163, opened 2023-09-13
Invalid updateTime returned from the save operation when document does not change


arturomelo-kareo:
**Describe the bug**
If I use the FirestoreReactiveRepository.save method to updated an existing document that contains the same values as the document stored in the database, the field annotated with the @UpdateTime(version = true) annotation returns a new timestamp. However, the value reported by Firestore is the previous timestamp as nothing was really updated.

**Sample**
This behavior causes in my code a precondition error because we are passing an invalid version (timestamp in the future)

"message": "FAILED_PRECONDITION: the stored version (1694626589979859) does not match the required base version (1694626366831818)",

**Version**
spring-cloud-gcp-data-firestore:3.5.3
google-cloud-firestore:3.14.0

meltsufin:
I'm able to repro with:
```java
  @Test
  public void testUpdateTimeNoDocumentChange() throws InterruptedException {
    User user = new User();
    user.setName("Axle");
    user.setAge(25);
    userRepository.save(user).block();
    userRepository.save(user).block();
    user.setAge(26);
    userRepository.save(user).block();
  }
```

PR #2165, opened 2023-09-14
fix: Firestore updateTime extraction after commit

meltsufin:
This change ensures that the correct updateTime is extracted for each entity after commit. Previously, commit timestamp was used, but it's wrong when there is no change to the entity.

Fixes: #2163.


JoeWang1127:
Could you add a unit test as well?

arturomelo-kareo:
Hi @meltsufin 
Thank you for working on this bug so fast.
Due to JRE 1.8 constraint we can not go above version 

`<groupId>com.google.cloud</groupId>
                <artifactId>spring-cloud-gcp-dependencies</artifactId>
                
                <version>3.5.3</version>`
                
                How could we grab this fixed once approved? Do you have a way to branch/tag of of that version?

meltsufin:
@arturomelo-kareo We will make sure to backport the fix to the 3.x branch.

meltsufin:
> Could you add a unit test as well?

I updated the tests that already test this.

JoeWang1127 (approved):

PR #2168, opened 2023-09-15
fix(3.x): Firestore updateTime extraction after commit

meltsufin:
This change ensures that the correct updateTime is extracted for each entity after commit. Previously, commit timestamp was used, but it's wrong when there is no change to the entity.

Backport of #2165.
Fixes: #2163 for 3.x branch.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
+++ b/spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
@@ -42,6 +42,7 @@ import com.google.firestore.v1.RunQueryResponse;
 import com.google.firestore.v1.StructuredQuery;
 import com.google.firestore.v1.Value;
 import com.google.firestore.v1.Write;
+import com.google.firestore.v1.WriteResult;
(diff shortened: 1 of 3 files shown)


JoeWang1127 (approved):

spring-cloud-gcp-data-firestore/src/test/java/com/google/cloud/spring/data/firestore/FirestoreTemplateTests.java
```diff
@@ -280,9 +281,11 @@ private void mockCommitMethod() {
     doAnswer(
             invocation -> {
               StreamObserver<CommitResponse> streamObserver = invocation.getArgument(1);
+              com.google.protobuf.Timestamp ts = Timestamp.ofTimeMicroseconds(123456789).toProto();
               CommitResponse response =
                   CommitResponse.newBuilder()
-                      .setCommitTime(Timestamp.ofTimeMicroseconds(123456789).toProto())
+                      .addWriteResults(WriteResult.newBuilder().setUpdateTime(ts).build())
+                      .addWriteResults(WriteResult.newBuilder().setUpdateTime(ts).build())
```
arturomelo-kareo, line 288: Do you need both lines?

PR #2157, opened 2023-09-11
chore: add secure password to root user in mysql CI db

diegomarquezp:
This is intended to fix a vulnerability brought by [b/296915631](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/b/296915631). The fix is about [SQL_WEAK_ROOT_PASSWORD](http://vulnstore/#advisories&expand=true&q=publisher:SHA+reference:SQL_WEAK_ROOT_PASSWORD)

PR #2185, opened 2023-09-20
chore(3.x): add secure password to root user in mysql CI db

meltsufin:
Backport: #2157.


diegomarquezp (approved):
Thanks for solving this!

Issue #2205, opened 2023-09-28
project ID is not read from the service account json set via credentials.encoded-key anymore


thoastbrot:
**Describe the bug**
Hi, I just migrated from 4.7.2 to 4.8.0 and my test configuration stopped working. I use firestore and for an integration test using `-Dspring.cloud.gcp.firestore.credentials.encoded-key=...`

The encoded service account credentials contain the project ID. Since https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/2164, this is broken - project ID is resolved at a later stage (I guess) and null, thus the code sanitizing the project ID fails.

Setting a dummy project-id in the configuration doesn't help either - it will crash on `Cloud Firestore API has not been used in project dummy`, as "dummy" wins.

meltsufin:
Can you include a stack trace and also explain what you mean by "project ID is resolved at a later stage (I guess) and null, thus the code sanitizing the project ID fails"?

thoastbrot:
Yeah, sorry - was a bit thin layer of information.

Stacktrace to the original issue:

```
Error creating bean with name 'firestoreRoutingHeadersInterceptor' defined in class path resource [com/google/cloud/spring/autoconfigure/firestore/GcpFirestoreAutoConfiguration$FirestoreReactiveAutoConfiguration.class]: Failed to instantiate [io.grpc.ClientInterceptor]: Factory method 'firestoreRoutingHeadersInterceptor' threw exception with message: Cannot invoke "String.length()" because "s" is null
	at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:801)
	at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:545)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1332)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1162)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:560)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:520)
[...]

PR #2208, opened 2023-09-29
fix: guard against null projectId in Firestore routing header

meltsufin:
Fixes: #2205.


burkedavison (approved):

JoeWang1127:
Just curious do we need a test to verify the code?

PR #2233, opened 2023-10-10
fix: removed topic existence & creation from provisionConsumerDestination

### LINKED ISSUE #2231: Disable the topic existence check in PubSubChannelProvider
**Describe the bug**

Scenario:
Let say I have two GCP projects A and B. I have created a topic in GCP Project B `Topic01` and subscription in Project A referring to a topic `Topic01` in project B. 
From application side added consumer using spring cloud stream PubSub binder to attach to subscription in Project A. 

Problem:
As topic should not be required for consumer even If I added a configuration `autoCreateResources=false`. Google Pubsub binder still calls a method `ensureTopicExists` in a call `PubSubChannelProvisioner`. So if I dont add a topic application will throw an exception `Non-existing topic.`
Sample Config:
```spring:
  cloud:
    stream:
      gcp:
        pubsub:
          default:
            consumer:
              subscriptionName: mySubscription
              autoCreateResources: false
      bindings:
        consumer-in-0:
          group: consumer-group
```
And in case I add a topic and as per the GCO config project A does not have any permission on topic created in Project B. So I get permission denied error. (As ideally consumer should not have any access to topic, let me know in case I have a wrong assumption about this)

Sample Config:
```spring:
   cloud:
    stream:
      gcp:
        pubsub:
          default:
            consumer:
              subscriptionName: mySubscription
              autoCreateResources: false
      bindings:
        consumer-in-0:
          destination: Topic01
          group: consumer-group
```
Spring Cloud GCP PubSub Stream binder version: 4.8.0

So I think in case of consumer or in case of autoCreateResources=false gcp-pubsub-binder should not check for topic.

SheheryarAamir:
fix: removed topic existence and creation from provisionConsumerDestination

As Pub/Sub subscriber role does have permission to access or create topic so in case `autoCreateResources` is set to false `provisionConsumerDestination` should not perform any admin task e.g fetch or create  subscription or topic. 
So in case subscription name is provided in the configuration with `autoCreateResources` to false provisionConsumerDestination should return `PubSubConsumerDestination` with provided subscription name.

also removed test against creation validation of topic for provision consumer Fixes #2231

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
@@ -68,10 +68,10 @@ public class PubSubChannelProvisioner
       String group,
       ExtendedConsumerProperties<PubSubConsumerProperties> properties) {
 
-    String subscriptionName;
-    Subscription subscription;
-
+    String subscriptionName = null;
     String customName = properties.getExtension().getSubscriptionName();
+
+
     boolean autoCreate = properties.getExtension().isAutoCreateResources();
     PubSubConsumerProperties.DeadLetterPolicy deadLetterPolicy =
         properties.getExtension().getDeadLetterPolicy();
@@ -79,7 +79,6 @@ public class PubSubChannelProvisioner
     // topicName may be either the short or fully-qualified version.
     String topicShortName =
         TopicName.isParsableFrom(topicName) ? TopicName.parse(topicName).getTopic() : topicName;
-    Topic topic = ensureTopicExists(topicName, autoCreate);
 
     if (StringUtils.hasText(customName)) {
       if (StringUtils.hasText(group)) {
@@ -89,33 +88,20 @@ public class PubSubChannelProvisioner
                 + customName
                 + "'.");
       }
-      subscription = this.pubSubAdmin.getSubscription(customName);
       subscriptionName = customName;
     } else if (StringUtils.hasText(group)) {
       subscriptionName = topicShortName + "." + group;
-      subscription = this.pubSubAdmin.getSubscription(subscriptionName);
-    } else {
-      // Generate anonymous random group since one wasn't provided
-      subscriptionName = "anonymous." + topicShortName + "." + UUID.randomUUID().toString();
-      subscription =
-          this.createSubscription(subscriptionName, topicName, deadLetterPolicy, autoCreate);
-      this.anonymousGroupSubscriptionNames.add(subscriptionName);
     }
 
-    if (subscription == null) {
-      if (autoCreate) {
-        this.createSubscription(subscriptionName, topicName, deadLetterPolicy, autoCreate);
-      } else {
-        throw new ProvisioningException("Non-existing '" + subscriptionName + "' subscription.");
+    if (autoCreate) {
+      if (!StringUtils.hasText(subscriptionName)) {
+        // Generate anonymous random group since one wasn't provided
+        subscriptionName = "anonymous." + topicShortName + "." + UUID.randomUUID();
+        this.anonymousGroupSubscriptionNames.add(subscriptionName);
       }
-    } else if (!subscription.getTopic().equals(topic.getName())) {
-      throw new ProvisioningException(
-          "Existing '"
-              + subscriptionName
-              + "' subscription is for a different topic '"
-              + subscription.getTopic()
-              + "'.");
+      ensureSubscriptionExists(subscriptionName, topicName, deadLetterPolicy, autoCreate);
     }
+
     return new PubSubConsumerDestination(subscriptionName);
   }
 
@@ -149,6 +135,27 @@ public class PubSubChannelProvisioner
       }
     }
     throw new ProvisioningException("Non-existing '" + topicName + "' topic.");
+  }
+
+  Subscription ensureSubscriptionExists(
+      String subscriptionName,
+      String topicName,
+      PubSubConsumerProperties.DeadLetterPolicy deadLetterPolicy,
+      boolean autoCreate) {
+    Subscription subscription;
+    subscription = this.pubSubAdmin.getSubscription(subscriptionName);
+    if (subscription != null) {
+      if (!subscription.getTopic().equals(topicName)) {
+        throw new ProvisioningException(
+            "Existing '"
+                + subscriptionName
+                + "' subscription is for a different topic '"
+                + subscription.getTopic()
+                + "'.");
+      }
+      return subscription;
+    }
+    return createSubscription(subscriptionName, topicName, deadLetterPolicy, autoCreate);
   }
 
   private Subscription createSubscription(
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
@@ -71,7 +71,6 @@ public class PubSubChannelProvisioner
     String subscriptionName = null;
     String customName = properties.getExtension().getSubscriptionName();
 
-
     boolean autoCreate = properties.getExtension().isAutoCreateResources();
     PubSubConsumerProperties.DeadLetterPolicy deadLetterPolicy =
         properties.getExtension().getDeadLetterPolicy();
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
```diff
@@ -141,49 +141,10 @@ void testProvisionConsumerDestination_customSubscription() {
...
-  void testProvisionConsumerDestination_noSubscriptionException() {
-    when(this.pubSubConsumerProperties.isAutoCreateResources()).thenReturn(false);
-
-    assertThatExceptionOfType(ProvisioningException.class)
-        .isThrownBy(
-            () ->
-                this.pubSubChannelProvisioner.provisionConsumerDestination(
-                    "topic_A", "group_A", this.properties))
-        .withMessage("Non-existing 'topic_A.group_A' subscription.");
-  }
```
burkedavison, line 167: Previously:
> When auto-create-resources is `false`, and `topicName.groupName` does not result in an existing subscription, then throw a ProvisioningException.

Now:
> When auto-create-resources is `false`, and `topicName.groupName` does not result in an existing subscription, then return a new PubSubConsumerDestination instance.

Is this change intentional? Could you clarify?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
```diff
@@ -141,49 +141,10 @@ void testProvisionConsumerDestination_customSubscription() {
...
-    when(this.pubSubAdminMock.getSubscription("topic_A.group_A"))
-        .thenReturn(Subscription.newBuilder().setTopic("topic_B").build());
-
-    assertThatExceptionOfType(ProvisioningException.class)
-        .isThrownBy(
-            () ->
-                this.pubSubChannelProvisioner.provisionConsumerDestination(
-                    "topic_A", "group_A", this.properties))
-        .withMessage("Existing 'topic_A.group_A' subscription is for a different topic 'topic_B'.");
-  }
```
burkedavison, line 181: Previously:
> When auto-create-resources is false, and `topicA.groupName` corresponds to a subscription assigned to `topicB`, then throw a ProvisioningException.

Now:
> When auto-create-resources is false, and `topicName.groupName` is provided, then do not check for the existence of the subscription or check its topic assignment. Instead, return a new PubSubConsumerDestination instance.

Like the above, this seems like an unintentional change in behavior outside the scope of the primary ensureTopicExists change. If it's intended, please provide some clarification.

burkedavison (changes requested):

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
SheheryarAamir: Yes, as a Pub/Sub Subscriber role `consumer` does not have access to `getSubscription` and it is not required as well until resource needs to be created by specifying `isAutoCreateResources=true`. 
So this test is no longer required  as it will not check for subscription and in case `isAutoCreateResources` is true it will create a subscription if not available.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
SheheryarAamir: As behaviour is changed, subscription will be verified only if `isAutoCreateResources` is set to true. So I have added a new test instead with a different name `testProvisionConsumerDestination_subscriptionHasDifferentTopic`. I can rename it back to previous if required

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
meltsufin: As I understand, this no longer validates the existence of the subscription, thereby allowing the more limited-access "Pub/Sub Subscriber" role to work.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
```diff
@@ -141,49 +141,10 @@ void testProvisionConsumerDestination_customSubscription() {
...
-
   @Test
   void testProvisionConsumerDestination_anonymousGroup() {
     // should work with auto-create = false
```
meltsufin, line 146: I guess this comment is no longer accurate. Does this also change the failure message for the scenario?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
```diff
@@ -266,20 +227,6 @@ void testAfterUnbindConsumer_nonAnonymous() {
     verify(this.pubSubAdminMock, never()).deleteSubscription(result.getName());
   }
 
-  @Test
-  void testProvisionConsumerDestination_concurrentTopicCreation() {
```
meltsufin, line 270: Why is this no longer a valid scenario?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
```diff
@@ -289,4 +236,56 @@ void testProvisionConsumerDestination_recursiveExistCalls() {
...
+    assertThatExceptionOfType(IllegalArgumentException.class)
+        .isThrownBy(
+            () ->
+                this.pubSubChannelProvisioner.provisionConsumerDestination(
+                    "topic_A", null, this.properties))
+        .withMessage("Subscription Name cannot be null or empty");
+  }
+
+  @Test
+  void testProvisionConsumerDestination_createAnonymousSubscription() {
```
meltsufin, line 253: How's this different from `testProvisionConsumerDestination_anonymousGroup()` above?

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -89,33 +87,21 @@ public ConsumerDestination provisionConsumerDestination(
...
-      throw new ProvisioningException(
-          "Existing '"
-              + subscriptionName
-              + "' subscription is for a different topic '"
-              + subscription.getTopic()
-              + "'.");
+      ensureSubscriptionExists(subscriptionName, topicName, deadLetterPolicy, autoCreate);
     }
+
+    Assert.hasText(subscriptionName, "Subscription Name cannot be null or empty");
```
meltsufin, line 104: This is an improvement unrelated to the goal of the PR, correct?

meltsufin (commented):
@SheheryarAamir Thanks for you contribution! It's looking pretty good to me. I just have a few questions/comments.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
SheheryarAamir: Yes, I will remove it

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
SheheryarAamir: because `provisionConsumerDestination` now does not have topic creation functionality

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
SheheryarAamir: correct, `testProvisionConsumerDestination_createAnonymousSubscription` is not required

spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
SheheryarAamir: Yes

burkedavison (approved):

meltsufin (approved):

PR #2239, opened 2023-10-11
chore: expose additional topic name information for IT debugging (#2238)

burkedavison:


SheheryarAamir:
I think issue is with how we pass `topicName` in configuration. Previously code was retrieving topic name with `this.pubSubAdmin.getTopic(topicName)` and then doing the comparison. Now we can not call `this.pubSubAdmin.getTopic(topicName)` because of permission limitation. So may be we can do something like this `subscription.getTopic().equals(this.pubSubAdmin.getTopic(topicName).getName())`

burkedavison:
@SheheryarAamir : Thanks for the suggestion. After discussing with @meltsufin , we've decided to remove this topic name comparison as it's now an inconsistent (only when auto-create-resources=true) unnecessary API call to validate configurations at runtime which was not a requirement.

Note: This is likely to fail some unit tests as-is. I'll clean them up tomorrow.

PR #2266, opened 2023-10-18
ci: add parallel integration and native tests (#2265)

burkedavison:
* Adds a parallel-integrationTests and parallel-NativeTests job that copies the non-parallel workflow except for the blocking step prior to integration testing.

* Removes 'compile' from the mvn dependency:go-offline command and adds `-T 1.5C -ntp -B` to speed this step up (~10mins per test before change). Now ~1m

* Uses Temurin for integration testing

### DIFF
--- a/.github/workflows/NativeTests.yaml
+++ b/.github/workflows/NativeTests.yaml
@@ -11,9 +11,11 @@ on:
     - 'spring-cloud-previews/**'
   workflow_dispatch:
 
+env:
+  EXCLUDED_MODULES: spring-cloud-spanner-spring-data-r2dbc
 
 jobs:
-  NativeTests:
+  parallel-NativeTests:
     if: |
       github.actor != 'dependabot[bot]' && ((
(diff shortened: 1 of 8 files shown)


.github/workflows/NativeTests.yaml
```diff
@@ -63,9 +64,71 @@ jobs:
...
             --define maven.javadoc.skip=true \
             install
-      - name: Retry install on failure
-        id: install2
-        if: steps.install1.outcome == 'failure'
+      # Unlike for the other NativeTests job, we don't block here as these run in parallel.
+      - uses: graalvm/setup-graalvm@v1
+        with:
+          version: '22.3.2'
+          java-version: '17'
```
zhumin8, line 71: FYI - This github action `setup-graal` has some changes since we last introduced them following GraalVm version naming changes. [readme](https://github.com/graalvm/setup-graalvm#migrating-from-graalvm-223-or-earlier-to-the-new-graalvm-for-jdk-17-and-later). If we want to move to test against graalvm for jdk 17

zhumin8 (approved):

PR #2273, opened 2023-10-19
ci (3.x): backport parallel integration testing

burkedavison:


.github/workflows/integrationTests.yaml
```diff
@@ -97,9 +70,83 @@ jobs:
...
+          DB_PASSWORD: ${{ secrets.SPRING_CLOUD_GCP_CI_DB_ROOT_PASSWORD }}
+      - name: Aggregate Report
+        run: |
+          ./mvnw \
+            --batch-mode \
+            --define aggregate=true \
+            surefire-report:failsafe-report-only
+      - name: Archive logs
+        if: always()
+        continue-on-error: true
```
mpeddada1, line 97: Do we still require the `continue-on-error` here?

PR #2007, opened 2023-07-06
chore: add pubsub sample tests to native ci

zhumin8:
Adding tests these samples to native ci:
- `spring-cloud-gcp-pubsub-sample`
- `spring-cloud-gcp-pubsub-reactive-sample`
     - Needs  [-H:+RunReachabilityHandlersConcurrently](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/7bd644151bd69fe0c07d9222dfc80ef1d8a1d600/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-reactive-sample/pom.xml#L77-L84) arg to run successfully. May want to hold off for client library support upgrade.
- `spring-cloud-gcp-integration-pubsub-sample` 
     - `spring-cloud-gcp-integration-pubsub-sample-test` module is not included because the way test setup gives issue.
     - pubsub-integration functionality covered in the other 3 module tests included.

Note: reactive-sample test runs successfully with [-H:+RunReachabilityHandlersConcurrently](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/7bd644151bd69fe0c07d9222dfc80ef1d8a1d600/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-reactive-sample/pom.xml#L77-L84) argument (ci for [7bd6441](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/2007/commits/7bd644151bd69fe0c07d9222dfc80ef1d8a1d600) should pass). And fails without it (ci for [4e20cf5](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/2007/commits/4e20cf5cd2575f382f5994b8d6f10da84e3d5273) expected to fail).

Also notice the ci is getting slow with these tests added.

[Update from @mpeddada1,  Oct 5, 2023] 
- Added `spring-cloud-gcp-pubsub-bus-config-sample`, `spring-cloud-gcp-pubsub-stream-dead-letter-sample`, `spring-cloud-gcp-pubsub-stream-functional-sample` and `spring-cloud-gcp-integration-pubsub-json-sample` to CI. 

[...]

PR #2288, opened 2023-10-24
chore: split pubsub tests into sub-jobs

mpeddada1:
As discovered in #2007, the Pub/Sub native image job is taking a really long time to run (approx. 145 mins). Instead of all Pub/Sub tests running in a single job, this PR splits them into multiple sub-jobs:
- pubsub-sample which runs template sample tests
- pubsub-integration-sample which runs Spring Integration tests
- pubsub-stream-sample which runs Cloud Stream tests
- pubsub-bus-sample which runs tests under `spring-cloud-gcp-pubsub-bus-config-sample`

### DIFF
--- a/.github/workflows/scripts/native-image-tests.sh
+++ b/.github/workflows/scripts/native-image-tests.sh
@@ -1,25 +1,63 @@
 #!/bin/bash
-
 # This script executes native image tests based on the MODULE_UNDER_TEST provided.
 # If MODULE_UNDER_TEST is 'vision', for example, then it runs tests under spring-cloud-gcp-vision.
 # If it is 'vision-sample' then it runs tests under spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample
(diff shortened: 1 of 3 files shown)


.github/workflows/scripts/native-image-tests.sh
```diff
@@ -10,14 +9,30 @@ set -eo pipefail
 # Get git repo root
 scriptDir=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
 cd "${scriptDir}/../../.."
+IFS='-' read -ra MATRIX_SUBSTRINGS <<< "$MODULE_UNDER_TEST"
```
burkedavison, line 12: Leaving IFS set to the non-default value can cause issues in the rest of the script.

Consider:
```
OLD_IFS=IFS
IFS=...
IFS=OLD_IFS
```

burkedavison (approved):

.github/workflows/scripts/native-image-tests.sh
mpeddada1: Done, thank you!

PR #2303, opened 2023-10-27
ci: remove mvn go-offline

burkedavison:
Go-offline isn't required, takes 10mins if compile is performed at the same time, and appears to be break without compile. This PR provides an option to just remove the stage.


JoeWang1127:
@burkedavison Can you add reason (or create an issue) for this change?

burkedavison:
Done

meltsufin (approved):

lqiu96 (approved):

PR #2310, opened 2023-10-30
chore: add kms module and sample to native ci.

zhumin8:
Adding to native ci. No module code changes.
- Remove hard coded project id and encrypted text, put in `application.properties` which can be overriden with system properties in pom for different profiles.

- [x] identify if tests can support running in parallel 
  - Shared resources are 2 crypto keys in ci project. No creation/deletion of keys.


zhumin8:
` PERMISSION_DENIED: Cloud Key Management Service (KMS) API has not been used in project xxxx before or it is disabled.`
To fix ci setup.

burkedavison (approved):

PR #2356, opened 2023-11-15
sample: Transactional annotation sample to demonstrate Read-write transaction

jainsahab:
Adding a capability to increment the *read count* of a book, It starts a transaction, and then fetches a book by ID and update the count followed by persisting the changes back to the database before committing the transaction.

image

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-spanner-r2dbc-samples/src/main/java/com/example/Book.java
+++ b/spring-cloud-gcp-samples/spring-cloud-spanner-r2dbc-samples/src/main/java/com/example/Book.java
@@ -20,13 +20,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import org.springframework.data.annotation.Id;
+import org.springframework.data.annotation.Transient;
 import org.springframework.data.domain.Persistable;
 import org.springframework.data.relational.core.mapping.Column;
 import org.springframework.data.relational.core.mapping.Table;
 
 /** Book entity. */
 @Table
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-samples/spring-cloud-spanner-r2dbc-samples/src/main/java/com/example/Book.java
```diff
@@ -44,11 +45,19 @@ public class Book implements Persistable {
   @Column("CATEGORIES")
   private List<String> categories;
 
+  @Column("COUNT")
+  private int count;
+
+  @Transient
+  private boolean isNew;
+
```
meltsufin, line 53: Can you just re-use the existing fields. You can add a method like `updateTitle`.

jainsahab: wanted to keep the UI part simple, that would need one more text box. let me know if you want that.

meltsufin: That's fine.

meltsufin (approved):

spring-cloud-gcp-samples/spring-cloud-spanner-r2dbc-samples/src/main/java/com/example/Book.java
jainsahab: Thanks,

PR #2359, opened 2023-11-18
fix: update ghpages.sh

### LINKED ISSUE #2101: Reference Documentation issues
- [x] Currently 4.x refdoc generation is failing
- [x] 4.x ref documentation is not listed in the README
- [ ] We shouldn't publish refdocs for SNAPSHOT versions
- [ ] Latest should be the latest released version of the highest major version
- [x] Auto-update the links to latest refdocs for each major version in the README
- [ ] We should check how Spring currently generates their refdocs and see if we need to update

JoeWang1127:
In this PR:
- fix #2101 
- update GH action versions
- code refactor

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -51,6 +51,42 @@
 					<plugin>
 						<groupId>org.asciidoctor</groupId>
 						<artifactId>asciidoctor-maven-plugin</artifactId>
+						<executions>
+							<execution>
+								<id>generate-html-documentation</id>
+								<phase>process-resources</phase>
+								<goals>
+									<goal>process-asciidoc</goal>
+								</goals>
+								<configuration>
+									<backend>spring-html</backend>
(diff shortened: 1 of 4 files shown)


docs/pom.xml
```diff
@@ -24,12 +24,8 @@
 
 		<!--maven-resources-plugin:3.2.0:copy-resources fails: https://issues.apache.org/jira/browse/MSHARED-966-->
 		<maven-resources-plugin.version>3.3.1</maven-resources-plugin.version>
-		<!-- override version inherited from spring-cloud-build parent to one that is available on Maven Central -->
-		<spring-asciidoctor-backends.version>0.0.4</spring-asciidoctor-backends.version>
```
JoeWang1127, line 28: This is not necessary as the asciidoctor backend version (0.0.5) inherited from spring-cloud-build is available on Maven Central.

PR #2365, opened 2023-11-20
fix: downgrade spring-cloud-build

JoeWang1127:


meltsufin:
Can you please provide some more context for this PR?

JoeWang1127:
This is an experiment of solving #2101.

I'll create a doc to explain my thoughts and trials.

PR #2370, opened 2023-11-22
feat: only commit change if it's not a snapshot version

JoeWang1127:
In this PR:
- Fix: #2372 
- Code refactor

### DIFF
--- a/docs/src/main/asciidoc/ghpages.sh
+++ b/docs/src/main/asciidoc/ghpages.sh
@@ -234,6 +234,15 @@ function commit_changes_if_applicable() {
     fi
 }
 
+# Switch back to the previous branch and exit block
+function checkout_previous_branch() {
+    # If -version was provided we need to come back to root project
+    cd ${ROOT_FOLDER}
+    git checkout ${CURRENT_BRANCH} || echo "Failed to check the branch... continuing with the script"
+    if [ "$dirty" != "0" ]; then git stash pop; fi
+    exit 0
+}
+
 # Assert if properties have been properly passed
 function assert_properties() {
 echo "VERSION [${VERSION}], DESTINATION [${DESTINATION}], CLONE [${CLONE}]"
@@ -328,3 +337,4 @@ build_docs_if_applicable
 retrieve_doc_properties
 stash_changes
 add_docs_from_target
+checkout_previous_branch
--- a/docs/src/main/asciidoc/ghpages.sh
+++ b/docs/src/main/asciidoc/ghpages.sh
@@ -234,15 +234,6 @@ function commit_changes_if_applicable() {
     fi
 }
 
-# Switch back to the previous branch and exit block
-function checkout_previous_branch() {
-    # If -version was provided we need to come back to root project
-    cd ${ROOT_FOLDER}
-    git checkout ${CURRENT_BRANCH} || echo "Failed to check the branch... continuing with the script"
-    if [ "$dirty" != "0" ]; then git stash pop; fi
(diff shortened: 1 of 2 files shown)


JoeWang1127:
If the version is not set or contains `SNAPSHOT`, the workflow will not commit change to `gh-pages` branch, thus no reference docs update.
```
+ [[ -z '' ]]
+ echo -e 'SNAPSHOT version, do not commit.'
```

docs/src/main/asciidoc/ghpages.sh
```diff
@@ -132,7 +132,12 @@ function add_docs_from_target() {
     else
         copy_docs_for_provided_version
     fi
-    commit_changes_if_applicable
+
+    if [[ -z "${VERSION}" ]] || [[ "${VERSION}" == *"SNAPSHOT" ]] ; then
```
lqiu96, line 136: Do we ever generate docs off a branch/ non-released tagged version of spring-cloud-gcp? I wonder if it would make sense to try and guarantee that a version is always specified.

Do you think it may make sense to move the `if [[ -z "${VERSION}" ]]` check to the bottom and fail if it isn't specified? And remove the logic for `retrieve_current_branch`?

JoeWang1127: > Do we ever generate docs off a branch/ non-released tagged version of spring-cloud-gcp? I wonder if it would make sense to try and guarantee that a version is always specified.

When you merge a PR to the main branch, there's no version specified.
This is part of the log of my last commit to main branch:
```
+ [[ 1 > 0 ]]
+ key=--build
+ case ${key} in
+ BUILD=yes
+ shift
+ [[ 0 > 0 ]]

assert_properties
+ assert_properties
+ echo 'VERSION [], DESTINATION [], CLONE []'
+ [[ '' != '' ]]
+ [[ '' != '' ]]
+ [[ '' != '' ]]
```

[full log](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/6945801341/job/18895996406)

JoeWang1127: > And remove the logic for retrieve_current_branch?

Makes sense, I'll remove this function.

lqiu96: > When you merge a PR to the main branch, there's no version specified.

I'm thinking that it would make sense to not run the refdocs job on every merge to main (possibly removing this line: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/dfb0e2d2790b1a86ed0cd5d7b982e8b6cd3eb2d6/.github/workflows/updateDocs.yaml#L64-L65). Only when we create a release which is what I think this line is doing: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/dfb0e2d2790b1a86ed0cd5d7b982e8b6cd3eb2d6/.github/workflows/updateDocs.yaml#L60-L61

I was thinking that maybe the if check could go into the bottom of the script so that it'll stop the script before any logic runs (in the event that the job is still triggered on a snapshot version somehow).

lqiu96 (approved):

docs/src/main/asciidoc/ghpages.sh
lqiu96: After talking Joe, we decided this is fine and we will a way to improve the generation doc in the future.

JoeWang1127: Thanks for the quick review, @lqiu96.

PR #2407, opened 2023-12-05
chore: exclude commons logging from artifacts.

zhumin8:
Spring Boot starters ships with `spring-jcl` module ([doc](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/logging.adoc)), which comes with its own Commons Logging bridge implemented in the `spring-jcl` module ([Spring Framework doc ](https://docs.spring.io/spring-framework/reference/core/spring-jcl.html)). This has potential conflicts with commons-logging.jar from classpath.

In particular, this causes failure for native compilation when logging gets accidentally initialized at runtime instead of build time. 

This pr tries to exclude `commons-logging` from SCGCP artifacts. With 2 exceptions: preview modules and `spanner-data-spring-data-r2dbc` module.
Also removes this flag for native builds:
`<buildArg>--initialize-at-build-time=org.apache.commons.logging.LogFactory</buildArg>`

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-vision/pom.xml
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-vision/pom.xml
@@ -24,10 +24,5 @@
 			<artifactId>spring-cloud-gcp-vision</artifactId>
 		</dependency>
 
-		<dependency>
-			<groupId>com.google.cloud</groupId>
-			<artifactId>google-cloud-vision</artifactId>
-		</dependency>
-
 	</dependencies>
 </project>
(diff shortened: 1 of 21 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-vision/pom.xml
```diff
@@ -24,10 +24,5 @@
 			<artifactId>spring-cloud-gcp-vision</artifactId>
 		</dependency>
 
-		<dependency>
```
zhumin8, line 27: This dependency is redundant as it's part of `spring-cloud-gcp-vision` above.

burkedavison:
If we merge this PR, please create an issue to remove it once https://github.com/googleapis/google-http-java-client/pull/1905 has been released.

burkedavison (approved):

PR #2489, opened 2023-12-19
fix: config release-please to stop at original 5.0.0 merge commit

burkedavison:
See also https://github.com/googleapis/release-please/blob/611db3d5628d1ff4cd7c40259894daf0c13f8e17/docs/manifest-releaser.md?plain=1#L134-L147

This PR is intended to resolve the incorrect release-please 6.0.0 PR: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/2467

### DIFF
--- a/.github/release-please.yml
+++ b/.github/release-please.yml
@@ -1,6 +1,7 @@
 primaryBranch: main
 releaseType: java-yoshi
 handleGHRelease: true
+last-release-sha: 103846dbe135790a59788a5ae3bafe56ededba3e
 extraFiles: [
   "README.adoc",
   "spring-cloud-generator/scripts/resources/googleapis_modification_string.txt"


.github/release-please.yml
```diff
@@ -1,6 +1,7 @@
 primaryBranch: main
 releaseType: java-yoshi
 handleGHRelease: true
+last-release-sha: 103846dbe135790a59788a5ae3bafe56ededba3e
```
chingor13, line 4: This won't work in the app's config. It has to be in `release-please-config.json`. The other config options would need to port to that file too.

Then set `manifest: true` in this app config to tell it to use the `release-please-config.json`. We highly recommend using that config file because then you can also test your settings using the `release-please` CLI as well

burkedavison: Okay - makes sense. Thank you for the clarification.

After discussing with @suztomo , we've reset the v5.0.0 release to a commit on the main branch, so this should allow release please to correct itself with the current configuration.

PR #2746, opened 2024-03-28
docs: publisher executor config properties

### LINKED ISSUE #2745: Document new publisher executor configuration settings
Document the settings introduced in this PR: 
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/2738/files

burkedavison:
Resolves #2745

### DIFF
--- a/docs/src/main/asciidoc/pubsub.adoc
+++ b/docs/src/main/asciidoc/pubsub.adoc
@@ -92,6 +92,17 @@ Enables message ordering. | No | false
 The publisher endpoint.
 Example: `"us-east1-pubsub.googleapis.com:443"`.
 This is useful in conjunction with enabling message ordering because sending messages to the same region ensures they are received in order even when multiple publishers are used. | No | pubsub.googleapis.com:443
+| `spring.cloud.gcp.pubsub.publisher.executor-accept-tasks-after-context-close` |
+Sets the underlying `ThreadPoolTaskScheduler` 's https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/concurrent/ExecutorConfigurationSupport.html#setAcceptTasksAfterContextClose(boolean)[acceptTasksAfterContextClose] property.
+If true, tasks will be accepted after the `ContextClosedEvent` at the expense of a longer shutdown phase. | No | false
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -92,6 +92,17 @@ Enables message ordering. | No | false
...
+| `spring.cloud.gcp.pubsub.publisher.executor-await-termination-millis` |
+Sets the underlying `ThreadPoolTaskScheduler` 's https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/concurrent/ExecutorConfigurationSupport.html#setAwaitTerminationMillis(long)[awaitTerminationMillis] property.
+Determines the maximum number of milliseconds the scheduler will block on shutdown while waiting for remaining tasks to complete.
+This is particularly useful if the remaining tasks need access to other resources that are also managed by the container. | No | 0
+| `spring.cloud.gcp.pubsub.publisher.executor-wait-for-tasks-to-complete-on-shutdown` |
+Sets the underlying `ThreadPoolTaskScheduler` 's https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/concurrent/ExecutorConfigurationSupport.html#setWaitForTasksToCompleteOnShutdown(boolean)[waitForTasksToCompleteOnShutdown] property.
+Determines whether to wait for scheduled tasks to complete on shutdown, not interrupting running tasks and executing all tasks in the queue.
+When true, replicates the default behavior before Spring 6.1.x. | No | false
```
meltsufin, line 105: I wonder if we should default to `true` for backwards compatibility.

meltsufin (approved):

docs/src/main/asciidoc/pubsub.adoc
burkedavison: I suggest leaving it.

This was introduced as we moved to a new major version, it just happens to be one of the breaking changes we didn't introduce directly but instead indirectly via our Spring version bumps. Customers will not be exposed to this change in behavior w/o choosing to update to a new major.

I think it would also result in an inconsistent message about which Spring behavior changes we chose to override and which we chose to keep.

PR #2726, opened 2024-03-21
chore: update codegen to use single googleapis commitish.

zhumin8:
First step to solve #2722.
This should be the minimum set of changes to unblock release with hermetic build changes. 
Further cleanup are needed to simplify this workflow. Will raise separate pr.

PR #2764, opened 2024-04-01
chore(codegen): adopt hermetic build configs and simplify logic

zhumin8:
follow up to #2726. This pr tries to simplify logic now that googleapis commitish is consistent for all mono repo, it is not a blocker for current release cycle. 

- replace old logic to parse proto path from "OwlBot.yaml" file, to utilize the new "generation_config.yaml" and parse GAPIC.proto_path. Keep current logic to only grab latest stable version for now.
  - This step is done inside the loop for each library directory, look for library config by `api_shortname`. This is sufficient for current selection (stable libs). But noted  `api_shortname` is not unique identifier for libraries in config. Need followup on this.
- remove logic to checkout versioned proto folders, and set googleapis_committish directly from config.

### DIFF
--- a/spring-cloud-generator/scripts/generate-library-list.sh
+++ b/spring-cloud-generator/scripts/generate-library-list.sh
@@ -18,21 +18,10 @@ if [[ -z "$commitish" ]]; then
   exit 1
 fi
 
-# download the monorepo, need to loop through metadata there
-git clone https://github.com/googleapis/google-cloud-java.git
-
-# switch to the specified release commitish
-cd ./google-cloud-java
-git checkout $commitish
-
-# read googleapis committish used in hermetic build
-googleapis_committish=$(yq -r ".googleapis_commitish" generation_config.yaml)
-echo "googleapis_committish: ${googleapis_committish}"
-
 cd ${SPRING_GENERATOR_DIR}
 # start file, always override is present
 filename=${SPRING_GENERATOR_DIR}/scripts/resources/library_list.txt
-echo "# api_shortname, googleapis-folder, distribution_name:version, googleapis_committish, monorepo_folder" > "$filename"
+echo "# api_shortname, googleapis-folder, distribution_name:version, monorepo_folder" > "$filename"
 
 # loop through folders
 count=0
@@ -72,10 +61,7 @@ for d in ./google-cloud-java/*java-*/; do
   proto_paths_stable=$(echo "$library" | yq -r '.GAPICs[] | select(.proto_path | test("/v[0-9]+$")) | .proto_path')
   proto_paths_latest=$(echo "$proto_paths_stable" | sort -d -r | head -n 1)
 
-  echo "$api_shortname, $proto_paths_latest, $distribution_name, $googleapis_committish, $monorepo_folder" >> $filename
+  echo "$api_shortname, $proto_paths_latest, $distribution_name, $monorepo_folder" >> $filename
   count=$((count+1))
 done
 echo "Total in-scope client libraries: $count"
-
-# clean up
-rm -rf google-cloud-java/
--- a/spring-cloud-generator/scripts/generate-library-list.sh
+++ b/spring-cloud-generator/scripts/generate-library-list.sh
@@ -66,21 +66,13 @@ for d in ./google-cloud-java/*java-*/; do
     continue
   fi
 
-  # get monorepo-name as pattern ./google-cloud-java/<monorepo_name>/
-  monorepo_name=$(echo $d | sed 's#^./google-cloud-java/##' | sed 's#/$##')
+  # parse proto path from generation_config.yaml, find by api_shortname
+  # then sort and keep latest stable version
+  library=$(yq -r '.libraries[] | select(.api_shortname == "'"$api_shortname"'")' ./google-cloud-java/generation_config.yaml)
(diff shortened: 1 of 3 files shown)


spring-cloud-generator/scripts/generate-library-list.sh
```diff
@@ -66,24 +55,13 @@ for d in ./google-cloud-java/*java-*/; do
     continue
   fi
 
-  # get monorepo-name as pattern ./google-cloud-java/<monorepo_name>/
-  monorepo_name=$(echo $d | sed 's#^./google-cloud-java/##' | sed 's#/$##')
+  # parse proto path from generation_config.yaml, find by api_shortname
+  # then sort and keep latest stable version
+  library=$(yq -r '.libraries[] | select(.api_shortname == "'"$api_shortname"'")' ./google-cloud-java/generation_config.yaml)
+  proto_paths_stable=$(echo "$library" | yq -r '.GAPICs[] | select(.proto_path | test("/v[0-9]+$")) | .proto_path')
```
JoeWang1127, line 61: What happened if no stable proto path exists in a library?

spring-cloud-generator/scripts/generate-steps.sh
```diff
@@ -27,14 +27,22 @@ function compute_monorepo_version() {
...
-# See generate-library-list.sh and https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1390 for details
+# Get googleapis committish used for specified release and assign to global GOOGLEAPIS_COMMITTISH.
+# For each library in google-cloud-java, parse information from ".repo-metadata.json"
+# and hermetic build config file: generation_config.yaml
 # Assumes current working directory is spring-cloud-generator
 #
 # $1 - Monorepo version tag (or committish)
 function generate_libraries_list(){
   monorepo_commitish=$1
+  git clone https://github.com/googleapis/google-cloud-java.git
```
diegomarquezp, line 38: nit: we could use something like sparse_checkout ([example](https://github.com/googleapis/sdk-platform-java/blob/56775a61a054fb67a15023481cd3d0a0cf2b0496/library_generation/utils/utilities.sh#L313-L327)) to speed this up

spring-cloud-generator/scripts/generate-library-list.sh
zhumin8: We only generate spring starter for stable libraries ([logic](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/9f26aedc2877d304fbcc45e78c149dee7ca51e4e/spring-cloud-generator/scripts/generate-library-list.sh#L48-L51)).
I would expect if a library is stable (according to repo-metadata.json), it should have at least one stable (not alpha/beta) proto path. Correct me if I am wrong?
Btw, this logic is the same prior to this change, only parsing proto path directly from "OwlBot.yaml"

spring-cloud-generator/scripts/generate-steps.sh
zhumin8: `sparse-checkout` may not help as much in this case, because in the following steps, I need to loop in each module and look for the repo-metadata.json. However, doing a shallow clone with only 1 or 2 commit should speed things up noticeably.

diegomarquezp (approved):

spring-cloud-generator/scripts/generate-library-list.sh
zhumin8: @JoeWang1127 following up on this, let me know if the above answers your question? If so, I am going to merge this in.

spring-cloud-generator/scripts/generate-steps.sh
```diff
@@ -27,14 +27,22 @@ function compute_monorepo_version() {
...
+# and hermetic build config file: generation_config.yaml
 # Assumes current working directory is spring-cloud-generator
 #
 # $1 - Monorepo version tag (or committish)
 function generate_libraries_list(){
   monorepo_commitish=$1
+  git clone --depth=1 https://github.com/googleapis/google-cloud-java.git
+  pushd google-cloud-java || { echo "Failure: google-cloud-java folder does not exists."; exit 1; }
+  # read googleapis committish used in hermetic build
+  GOOGLEAPIS_COMMITTISH=$(yq -r ".googleapis_commitish" generation_config.yaml)
```
blakeli0, line 41: Is it still possible to keep the library level `googleapis_committish`? We do support it a different `googleapis_committish` at level level in `generation_config.yaml`. Vast majority of time every library in google-cloud-java should be generated with the same committish, but we did have a use case during last freeze that `java-language` wanted to release a new version, so in that case `java-language` would have a different committish.  It is a very rare case and could be addresses in a different PR.

zhumin8: IUCC, this library level "googleapis_committish" is optional, if it's present, we should parse if from the libraries and it takes precedence of this repo level config?
If so, the current `generation_config.yaml` does not have libraries with this setting, and as you said this is a rare case. I would like to address it in a followup pr, so we don't block the OwlBot.yaml change for the release cycle.

zhumin8:
rebase to trigger un-responsive tests.

lqiu96:
Seems like CI is still stuck. Going to close and re-open

blakeli0 (approved):

PR #2765, opened 2024-04-02
fix: BigQueryTemplate now closes its BigQueryJsonDataWriter

### LINKED ISSUE #2750: BigQueryJsonDataWriter never closes when using BigQueryTemplate.writeJsonStream()
Hello 👋 
I'm building a `CommandLineRunner` spring boot application that sends data into BigQuery. 
I want the application to shutdown when the process is finished. However, since the BigQueryJsonDataWriter never closes its `streamWriter`, a thread remains alive and the application hangs indefinitely. 

The workaround I found is to inherit the `BigQueryTemplate` class, override the `getWriteApiResponse()` method and instantiate the `BigQueryJsonDataWriter` into a try with resources. It works but it does not seem very elegant.

A cleaner solution would be to call the `close()` method of the `BigQueryJsonDataWriter` class when finalising the stream. Is there a reason the `close()` method is not called ? 
Also, [the `BigQueryJsonDataWriter` is instantiated every time a call to `BigQueryTemplate.writeJsonStream()` is made](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java#L288). As I make several calls to  `BigQueryTemplate.writeJsonStream()` in my application, wouldn't it be more efficient to instantiate it once and give the control to the user to close it when needed ?  

Thanks

burkedavison:
Fixes #2750

### DIFF
--- a/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
+++ b/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
@@ -322,7 +322,7 @@ public class BigQueryTemplate implements BigQueryOperations {
         }
 
       } catch (Exception e) {
-        throw new BigQueryException("Failed to append records. \n" + e);
+        throw new BigQueryException("Failed to append records.", e);
       }
 
       // Finalize the stream before committing it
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java
```diff
@@ -294,58 +294,55 @@ public WriteApiResponse getWriteApiResponse(String tableName, InputStream jsonIn
...
-          currentBatchSize = 0;
         }
-      }
 
-      if (jsonBatch.length()
-          != 0) { // there might be records less than JSON_STREAM_WRITER_BATCH_SIZE, append those as
-        // well
-        writer.append(jsonBatch, offset);
+      } catch (Exception e) {
+        throw new BigQueryException("Failed to append records. \n" + e);
```
meltsufin, line 325: Can we switch to using the `BigQueryException(String, Throwable)` constructor instead of converting the causing exception to a string?

burkedavison: Done.

meltsufin (approved):

Issue #2774, opened 2024-04-08
Spring Boot+datastore throws "Type must not be Null" when JsonNode is used as attribute on a Datastore Entity


jevanio:
**Describe the bug**
Hey folks,
I been working on a spring boot project which use Datastore as backend. On this project, I'm trying to update jackson dependencies from the current version it has (2.13.4) to a newer one (2.16.2) but I'm getting an issue on a Datastore Entity which use JsonNode as attribute.

This is the error:
![image](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/assets/1214408/51fc959e-d933-4f06-bd3a-8f537492a483)

I have seen that, after upgrading to jackson 2.14.X or higher, I'm getting an error on bean creation which only says "Type must not be null". After a lot of debugging, and creating a SpringBoot+Datastore project from scratch, I found that if I remove a JsonNode attribute from a Datastore Entity the service starts and beans can be creater. I must mention that this JsonNode attribute has been included from long time ago and it was working with Jackson 2.13.4.

The entity itself it just have 2 attributes with getter/setter
![image](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/assets/1214408/017b68d8-a67b-4db9-ae78-5f17552131da)

I feel that there's some configuration or dependencies which could be required but, as of now, I haven't been able to identify which change is missing. I have check Jackson Releases to see if there's any breaking change or mention but so far no luck.

Can you guide me if there's a configuration required to work with JsonNodes on GCP Cloud DataStore Entities?
Thanks in advance for any insights you can give me

**Sample**
This is sample project with Java 17, the issue can be reproduce changing jackson version to 2.14.X or higher. It's a base spring boot project with only 4 classes and no custom logic.
https://github.com/jevanio/demoProject

Note: To start the project on local, it's required to include as Environment Variable **JDK_JAVA_OPTIONS=--add-opens java.base/java.math=ALL-UNNAMED** to fix a math library error
![image](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/assets/1214408/de17f4a6-3cb6-41f3-8212-56739bcfb3d3)

suztomo:
@jevanio Thank you for reporting the issue. Would you add more stacktrace?

![image](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/assets/28604/31c6d553-9359-4213-8e27-6d7b3fd6440c)

[...]

PR #2796, opened 2024-04-26
fix: [datastore] do not create PersistentEntity for custom types

meltsufin:
Also guard against "Type must not be Null"

Fixes: #2774.


meltsufin:
It seems like not having a `PersistentEntity` for complex types is causing other problems.

PR #2797, opened 2024-04-26
fix(datastore): guard against "Type must not be Null"

meltsufin:
Fixes: #2774.


JoeWang1127:
Do we need to include a unit test?

burkedavison (approved):

meltsufin:
> Do we need to include a unit test?

Yes, added!

JoeWang1127 (approved):

PR #2839, opened 2024-05-02
chore: generate autoconfigs

diegomarquezp:
We needed to update the way we compute GOOGLEAPIS_COMMITISH to consider MONOREPO_TAG when cloning google-cloud-java. In this way, the generated autoconfigs use the protos matching the googleapis committish from the config yaml at the specified tag.


spring-cloud-generator/scripts/generate-steps.sh
```diff
@@ -35,8 +35,9 @@ function compute_monorepo_version() {
 # $1 - Monorepo version tag (or committish)
 function generate_libraries_list(){
   monorepo_commitish=$1
-  git clone --depth=1 https://github.com/googleapis/google-cloud-java.git
+  git clone --no-single-branch --depth=1 https://github.com/googleapis/google-cloud-java.git
   pushd google-cloud-java || { echo "Failure: google-cloud-java folder does not exists."; exit 1; }
+  git checkout "${MONOREPO_TAG}"
```
JoeWang1127, line 40: How to setup this variable?

diegomarquezp: It's explained in the main script. This is an environment variable that will have a default value if not set

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/7157231d8827809b82461c49ccbe9f74e6e4e108/spring-cloud-generator/scripts/generate.sh#L7-L12

JoeWang1127 (approved):

blakeli0 (approved):

Issue #2832, opened 2024-05-01
Flaky spanner integration test


burkedavison:
Example:
* https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/8900902682/job/24443531808

Investigate and resolve.
```
com.google.cloud.spanner.SpannerException: FAILED_PRECONDITION: Operation with name "projects/spring-cloud-gcp-ci/instances/reactivetest/databases/testdb/operations/r6cab2ab5_8f3a_458a_a3de_b82aaeeb7746" failed with status = GrpcStatusCode***transportCode=FAILED_PRECONDITION*** and message = Duplicate name in schema: card.
	at com.google.cloud.spanner.SpannerExceptionFactory.newSpannerExceptionPreformatted(SpannerExceptionFactory.java:291)
	at com.google.cloud.spanner.SpannerExceptionFactory.fromApiException(SpannerExceptionFactory.java:311)
	at com.google.cloud.spanner.SpannerExceptionFactory.newSpannerException(SpannerExceptionFactory.java:174)
	at com.google.cloud.spanner.SpannerExceptionFactory.newSpannerException(SpannerExceptionFactory.java:110)
	at com.google.cloud.spanner.DatabaseAdminClientImpl.lambda$updateDatabaseDdl$11(DatabaseAdminClientImpl.java:470)
	at com.google.api.core.ApiFutures$ApiFunctionToGuavaFunction.apply(ApiFutures.java:396)
	at com.google.common.util.concurrent.AbstractCatchingFuture$CatchingFuture.doFallback(AbstractCatchingFuture.java:237)
	at com.google.common.util.concurrent.AbstractCatchingFuture$CatchingFuture.doFallback(AbstractCatchingFuture.java:225)
	at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:135)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:31)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1298)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:1059)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:809)
	at com.google.common.util.concurrent.AbstractTransformFuture.run(AbstractTransformFuture.java:129)
[...]

Issue #2314, opened 2023-10-30
Flaky test: SpannerR2dbcDialectDateTimeBindingIntegrationTest


zhumin8:
Observed this test failing in ci with below error message twice:

```
Error:  Errors: 
Error:    SpannerR2dbcDialectDateTimeBindingIntegrationTest » Spanner FAILED_PRECONDITION: Operation with name "projects/spring-cloud-gcp-ci/instances/reactivetest/databases/testdb/operations/r9504ea9e_e735_4a46_acbc_bcff5f490623" failed with status = GrpcStatusCode***transportCode=FAILED_PRECONDITION*** and message = Duplicate name in schema: card.
Error:    SpannerR2dbcDialectJsonIntegrationTest » Spanner FAILED_PRECONDITION: Operation with name "projects/spring-cloud-gcp-ci/instances/reactivetest/databases/testdb/operations/r54858ab1_fdfa_4001_a0ab_b04fe61367ff" failed with status = GrpcStatusCode***transportCode=FAILED_PRECONDITION*** and message = Duplicate name in schema: PERSON.
[INFO] 
Error:  Tests run: 5, Failures: 0, Errors: 2, Skipped: 0
```
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/6670097939/job/18129301265?pr=2227
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/6696751266/job/18195151261?pr=2227

bjchris32:
hi @zhumin8 and @blakeli0 ,
Would you mind I take a look at this flaky test?
Thank you!

PR #2874, opened 2024-05-12
test: fix Spanner R2DBC flaky integration tests

meltsufin:
Create a new database for each test for isolation.

Fixes: #2832.
Fixes: #2314.


spring-cloud-spanner-spring-data-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/springdata/it/AbstractBaseSpannerR2dbcIntegrationTest.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  void initializeTestEnvironment(String tableDdl) {
+    initializeTestEnvironment(null, tableDdl);
+  }
+
+  @AfterEach
+  void cleanupTestInstanceAfterTest() {
+    log.info("Deleting database {}", testDatabase);
+    this.spanner.getDatabaseAdminClient().dropDatabase(TEST_INSTANCE, testDatabase);
```
burkedavison, line 74: If something goes wrong and this cleanup isn't able to run, we'll have an orphaned database that never gets removed. Example of resolving this situation in google-cloud-java, by checking at the beginning of a test whether any stale resources exist from prior runs: https://github.com/googleapis/google-cloud-java/blob/main/java-compute/google-cloud-compute/src/test/java/com/google/cloud/compute/v1/integration/Util.java#L41-L57

spring-cloud-spanner-spring-data-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/springdata/it/AbstractBaseSpannerR2dbcIntegrationTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.context.annotation.UserConfigurations;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+
+/**
+ * Abstract base for Spanner R2DBC integration tests of {@link SpannerR2dbcDialect}.
+ */
+abstract class AbstractBaseSpannerR2dbcIntegrationTest {
```
burkedavison, line 34: Test logs show a very large number of exceptions due to:
```
SEVERE: *~*~*~ Previous channel ManagedChannelImpl***logId=467, target=spanner.googleapis.com:443*** was garbage collected without being shut down! ~*~*~*
    Make sure to call shutdown()/shutdownNow()
```
With long stack traces that pollute the results. Can we shut down the clients to avoid these errors?

spring-cloud-spanner-spring-data-r2dbc/src/test/java/com/google/cloud/spanner/r2dbc/springdata/it/AbstractBaseSpannerR2dbcIntegrationTest.java
meltsufin: `@AfterEach` is guaranteed to run even if there is a test failure. Unlike Compute instances, Spanner databases don't incur additional costs, but there is a limit of 100 per instance. I would rather hold off the proactive additional cleanup that will slow down the tests, until it's clear that we need it.

burkedavison: @AfterEach doesn't run if the machine fails - the exact problem we had with google-cloud-java. I leave it up to you.

burkedavison (approved):

PR #2894, opened 2024-05-16
feat: configurable expiration policy on auto-created subscriptions

### LINKED ISSUE #2876: Allow setting Expiration Policy on Subscriptions created via Pub Sub Binder
When Pub Sub subscriptions are created, if no [expiration policy](https://cloud.google.com/pubsub/docs/subscription-properties#expiration_period) is specified they expire by default in 31 days. In the setup I am working with, it is common for there to be little activity for long periods of times on some streams. For some of these streams, we spin up consumers reactively once a batch of messages are queued in the subscription (as opposed to having them idle year-round). However, because the subscription is quiet for so long, sometimes it is automatically deleted in between batches. Since there is no way to set the expiration policy on the Binder's auto-created subscriptions, I must create them externally.

I would like to be able to configure the expiration policy for auto-created subscriptions so I don't have to rely on external setup.

If desired, I am happy to contribute this change. I have already prepared it for my own use: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/compare/main...jmitash:spring-cloud-gcp:feat/pubsub-binder-expiration-config.

jmitash:
Adds a way to configure the expiration policy on auto-created subscriptions.

The expiration policy contains a single field, `ttl` which is a Duration.

When the `ttl` is not set, the Pub Sub backend sets its own default of 31 (no expirationPolicy set on API request).

When the `ttl` is set to a >0 duration, that duration is configured for auto-created Subscriptions (expirationPolicy with ttl set on API request).

When the `ttl` is set to a <= 0 duration, the `ttl` is interpreted as "Never", and the auto-created Subscriptions will not expire (expirationPolicy with no ttl set on API request).

Fixes #2876

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/properties/PubSubConsumerProperties.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/com/google/cloud/spring/stream/binder/pubsub/properties/PubSubConsumerProperties.java
@@ -34,7 +34,8 @@ public class PubSubConsumerProperties extends PubSubCommonProperties {
   /**
    * Policy for how soon the subscription should be deleted after no activity.
    * <p>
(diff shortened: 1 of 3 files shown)


burkedavison:
Merging into in-repo branch to trigger integration tests.

Issue #2214, opened 2023-10-02
Showcase CI check fails due to removed `BUILD.bazel` in `sdk-platform-java`


diegomarquezp:
https://github.com/googleapis/sdk-platform-java/pull/1935 removed the showcase BUILD file, and now the [script](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/29229f726218228a9ed424e35644a1305b25b545/spring-cloud-generator/scripts/generate-showcase.sh#L66-L67) that manipulates such BUILD file is failing.

diegomarquezp:
This section assumes the existence of `BUILD.bazel` in `sdk-platform-java/showcase`
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/54ebcde41c41a0ea13756fa6ed42f29553292774/spring-cloud-generator/scripts/generate-showcase.sh#L79-L81

PR #2895, opened 2024-05-16
test: restore showcase autogen IT

diegomarquezp:
Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/2214 🔨 

### Context
So far, this test used bazel + a few string manipulations in sdk-platform-java's showcase BUILD file and WORKSPACE file. This wasn't working since we migrated showcase generation to our hermetic build scripts. This PR transforms the test into using protoc directly

### Approach
It's very similar to [`generate_library.sh`](https://github.com/googleapis/sdk-platform-java/blob/94ef4f42cfda9300f60acb9e6c14e0ca02109716/library_generation/generate_library.sh) in the sense that we infer the protoc version and then use the spring-cloud-generator jar as a plugin to protoc to compile the showcase protos, which are downloaded in the script.

### Other changes
 * Update showcase generated autoconfigs. Includes changes from updating gapic-showcase to a new version. See https://github.com/googleapis/sdk-platform-java/pull/2737 .
 * Changes in the autoconfig composer, specifically about explicitly setting the `endpoint` to `[Service]Settings.getDefaultEndpoint()`. The beans would fail to load otherwise. cc: @lqiu96 
 * Include the generation of a `jar-with-dependencies` for spring-cloud-generator. Apparently bazel was able to figure out how to inject dependencies in the generation runtime. Now we need to create a jar with all dependencies as we run protoc directly with the jar.
 * Added a wrapper executable script for the autoconfig generator. This is due to how `protoc` takes plugins.

### DIFF
--- a/spring-cloud-generator/pom.xml
+++ b/spring-cloud-generator/pom.xml
@@ -110,6 +110,28 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+              <artifactId>maven-assembly-plugin</artifactId>
+              <configuration>
+                <archive>
+                  <manifest>
+                    <mainClass>com.google.cloud.generator.spring.Main</mainClass>
+                  </manifest>
+                </archive>
+                <descriptorRefs>
+                  <descriptorRef>jar-with-dependencies</descriptorRef>
+                </descriptorRefs>
+              </configuration>
+              <executions>
+                <execution>
+                  <id>make-assembly</id> <!-- this is used for inheritance merges -->
+                  <phase>package</phase> <!-- bind to the packaging phase -->
+                  <goals>
+                    <goal>single</goal>
+                  </goals>
+                </execution>
+              </executions>
+            </plugin>
         </plugins>
     </build>
 
--- a/spring-cloud-generator/pom.xml
+++ b/spring-cloud-generator/pom.xml
@@ -124,8 +124,8 @@
               </configuration>
               <executions>
                 <execution>
-                  <id>make-assembly</id> <!-- this is used for inheritance merges -->
-                  <phase>package</phase> <!-- bind to the packaging phase -->
+                  <id>make-assembly</id>
+                  <phase>package</phase>
                   <goals>
                     <goal>single</goal>
                   </goals>
@@ -134,5 +134,4 @@
             </plugin>
         </plugins>
     </build>
-
 </project>
--- a/spring-cloud-generator/pom.xml
+++ b/spring-cloud-generator/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <artifactId>spring-cloud-gcp</artifactId>
         <groupId>com.google.cloud</groupId>
-        <version>5.2.2-SNAPSHOT</version><!-- {x-version-update:spring-cloud-gcp:current} -->
+        <version>5.3.0</version><!-- {x-version-update:spring-cloud-gcp:current} -->
     </parent>
 
     <groupId>com.google.cloud</groupId>
--- a/spring-cloud-generator/scripts/generate-showcase.sh
+++ b/spring-cloud-generator/scripts/generate-showcase.sh
@@ -1,12 +1,12 @@
 set -ex
 
-# To VERIFY: ./scripts/generate-showcase.sh
-# To UPDATE: /scripts/generate-showcase.sh -u
-UPDATE=0
+# To verify: ./scripts/generate-showcase.sh
+# To update: /scripts/generate-showcase.sh -u
+update="false"
 while getopts u flag
 do
     case "${flag}" in
-        u) UPDATE=1;;
+        u) update="true";;
     esac
 done
 
@@ -14,33 +14,33 @@ done
 source ./scripts/generate-steps.sh
 
 # If not set, assume working directory is spring-cloud-generator
-if [[ -z "$SPRING_GENERATOR_DIR" ]]; then
-  SPRING_GENERATOR_DIR=`pwd`
+if [[ -z "${SPRING_GENERATOR_DIR}" ]]; then
+  SPRING_GENERATOR_DIR=$(pwd)
 fi
-SPRING_ROOT_DIR=${SPRING_GENERATOR_DIR}/..
-SHOWCASE_STARTER_OLD_DIR=${SPRING_GENERATOR_DIR}/showcase/showcase-spring-starter
-SHOWCASE_STARTER_NEW_DIR=${SPRING_GENERATOR_DIR}/showcase/showcase-spring-starter-generated
+spring_root_dir=${SPRING_GENERATOR_DIR}/..
+showcase_starter_old_dir=${SPRING_GENERATOR_DIR}/showcase/showcase-spring-starter
+showcase_starter_new_dir=${SPRING_GENERATOR_DIR}/showcase/showcase-spring-starter-generated
 
 # Verifies newly generated showcase-spring-starter against goldens
 #
 # $1 - directory containing existing showcase-spring-starter (golden)
 # $2 - directory containing newly generated showcase-spring-starter
 function verify(){
-  OLD_DIR=$1
-  NEW_DIR=$2
-  SHOWCASE_STARTER_DIFF=$(diff -r ${NEW_DIR}/src/main ${OLD_DIR}/src/main)
-  SHOWCASE_STARTER_POM_DIFF=$(diff -r ${NEW_DIR}/pom.xml ${OLD_DIR}/pom.xml)
-  if [ "$SHOWCASE_STARTER_DIFF" != "" ] || [ "$SHOWCASE_STARTER_POM_DIFF" != "" ]
+  old_dir=$1
+  new_dir=$2
+  showcase_starter_diff=$(diff -r ${new_dir}/src/main ${old_dir}/src/main)
+  showcase_starter_pom_diff=$(diff -r ${new_dir}/pom.xml ${old_dir}/pom.xml)
+  if [ "${showcase_starter_diff}" != "" ] || [ "${showcase_starter_pom_diff}" != "" ]
   then
       echo "Differences detected in generated showcase starter module: "
       echo "Diff from src/main: "
-      echo $SHOWCASE_STARTER_DIFF
+      echo "${showcase_starter_diff}"
       echo "Diff from pom.xml: "
-      echo $SHOWCASE_STARTER_POM_DIFF
+      echo "${showcase_starter_pom_diff}"
       exit 1;
   else
       echo "No differences found in showcase-spring-starter"
-      rm -r ${NEW_DIR}
+      rm -r "${new_dir}"
   fi
 }
 
@@ -48,15 +48,15 @@ function verify(){
 #
 # $1 - target directory for generated starter
 function generate_showcase_spring_starter(){
-  SHOWCASE_STARTER_DIR=$1
+  showcase_starter_dir=$1
 
   # Compute the parent project version.
-  cd ${SPRING_ROOT_DIR}
-  export PROJECT_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
-  cd ${SPRING_GENERATOR_DIR}
-  GAPIC_GENERATOR_JAVA_VERSION=$(mvn help:evaluate -Dexpression=gapic-generator-java-bom.version -q -DforceStdout)
+  pushd "${spring_root_dir}"
+  export project_version=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+  cd "${SPRING_GENERATOR_DIR}"
+  gapic_generator_java_version=$(mvn help:evaluate -Dexpression=gapic-generator-java-bom.version -q -DforceStdout)
 
-  if [[ -z "$GAPIC_GENERATOR_JAVA_VERSION" ]]; then
+  if [[ -z "${gapic_generator_java_version}" ]]; then
     echo "Missing sdk-platform-java commitish to checkout"
     exit 1
   fi
@@ -66,7 +66,7 @@ function generate_showcase_spring_starter(){
     git clone https://github.com/googleapis/sdk-platform-java.git
   fi
   pushd sdk-platform-java
-  git checkout "v${GAPIC_GENERATOR_JAVA_VERSION}"
+  git checkout "v${gapic_generator_java_version}"
 
   # We will use the generation tools from library_generation
   pushd library_generation/utils
@@ -78,10 +78,10 @@ function generate_showcase_spring_starter(){
   # For local development, we cleanup any traces of previous runs
   rm -rdf output
   # mvn clean install
-  GAPIC_SHOWCASE_CLIENT_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+  gapic_showcase_client_version=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
 
   pushd gapic-showcase
-  GAPIC_SHOWCASE_SERVER_VERSION=$(mvn help:evaluate -Dexpression=gapic-showcase.version -q -DforceStdout)
+  gapic_showcase_server_version=$(mvn help:evaluate -Dexpression=gapic-showcase.version -q -DforceStdout)
   popd #showcase/gapic-showcase
 
   # Alternative: if showcase client library is available on Maven Central,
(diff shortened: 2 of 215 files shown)


spring-cloud-generator/pom.xml
```diff
@@ -110,7 +110,28 @@
...
+            <plugin>
+              <artifactId>maven-assembly-plugin</artifactId>
+              <configuration>
+                <archive>
+                  <manifest>
+                    <mainClass>com.google.cloud.generator.spring.Main</mainClass>
+                  </manifest>
+                </archive>
+                <descriptorRefs>
+                  <descriptorRef>jar-with-dependencies</descriptorRef>
```
diegomarquezp, line 122: This is in order to run the spring generator jar using dependencies such as protobuf classes. This was something bazel took care about with the previous setup.

spring-cloud-generator/showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/EchoSpringAutoConfiguration.java
```diff
@@ -130,6 +130,7 @@ public EchoSettings echoSettings(
     clientSettingsBuilder
         .setCredentialsProvider(this.credentialsProvider)
         .setTransportChannelProvider(defaultTransportChannelProvider)
+        .setEndpoint(EchoSettings.getDefaultEndpoint())
```
diegomarquezp, line 133: Not sure if this is what we want when creating client settings. We can follow up with proper changes in gax. 
cc: @lqiu96

lqiu96: Do you have the error message if you don't add this line to set the endpoint? Ideally, we shouldn't have to set the endpoint ourselves but for this PR, I think it's fine we can't resolve the issue here.

diegomarquezp: > Do you have the error message if you don't add this line to set the endpoint?

The error message:
```
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'echoClient' defined in com.google.showcase.v1beta1.spring.EchoSpringAutoConfiguration: Failed to instantiate [com.google.showcase.v1beta1.EchoClient]: Factory method 'echoClient' threw exception with message: Invalid host or port: .googleapis.com 443
	at org.springframework.beans.factory.support.ConstructorResolver.instantiate(ConstructorResolver.java:648)
	at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:636)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1335)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1165)
```

It fails to resolve the endpoint if we don't explicitly set it: `Invalid host or port: .googleapis.com 443`

> Or if this isn't because of an error, is this because we're using an old version of the generator. Or perhaps this is just something spring generator does?

I don't think this is very different from a normal use case of instantiating a client with a few settings.

Here we create the settings:
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/0f94e07471093fd5a2165907a0a4bd3c006ffd0d/spring-cloud-generator/showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/EchoSpringAutoConfiguration.java#L115-L121

Here we create the client with these settings:
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/0f94e07471093fd5a2165907a0a4bd3c006ffd0d/spring-cloud-generator/showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/EchoSpringAutoConfiguration.java#L339-L343

spring-cloud-generator/scripts/generate-showcase.sh
```diff
@@ -1,111 +1,174 @@
...
+  cp -r schema "${output_folder}"
+  popd #gapic-showcase
+
+  # We download googleapis and prepare the protos in output_folder
+  if [[ ! -d "./googleapis" ]]; then
+    sparse_clone https://github.com/googleapis/googleapis.git "google/iam/v1 google/cloud/location google/api google/longrunning google/rpc google/type"
+  fi
+  pushd googleapis
+  cp -r google "${output_folder}"
+  popd #gapic-showcase
```
burkedavison, line 114: popd #googleapis

diegomarquezp: Fixed

spring-cloud-generator/showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/EchoSpringAutoConfiguration.java
lqiu96: > It fails to resolve the endpoint if we don't explicitly set it: Invalid host or port: .googleapis.com 443

Ah, I think this is an edge case due to the the showcase protos. The showcase protos have a `default_host` value that won't match a "normal" google endpoint. For example, the echo proto's `service_host` value: https://github.com/googleapis/gapic-showcase/blob/4e7c30a5d5df8aac30e8b8f1e6f71809fa9d20f6/schema/google/showcase/v1beta1/echo.proto#L45
 where as a normal endpoint would be something like `cloudassets.googleapis.com`.

The error above is complaining that we can't populate the `serviceName` part of the endpoint and that's because the showcase protos don't have one. That's why you'd have to explicitly call `setEndpoint()` yourself.

This looks like it's changing the code inside the generator, which I believe would impact all of spring code gen. The endpoint normally *should* be constructed internally for non-showcase services and I believe having the generator explicitly call `setEndpoint()` may be imcompatible with enabling certain features.

In the showcase ITs, we have this same problem. But we solve it by explicitly calling `setEndpoint()` in the handwritten layers: https://github.com/googleapis/sdk-platform-java/blob/5799827a864be627bac03969cc178efc9d6170aa/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/util/TestClientInitializer.java#L67

I'm not too familiar with the showcase autogen ITs. Are there any handwritten parts that we can do something similar or are these the only tests that are generated + run?

blakeli0: > Are there any handwritten parts that we can do something similar or are these the only tests that are generated + run?

This is the [handwritten test](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/5b9fb60ca29e43d5ccac954dfd8fefca80715468/spring-cloud-generator/showcase/showcase-spring-starter/src/test/java/com/google/showcase/v1beta1/spring/EchoAutoConfigurationTests.java).Compare to GAPIC showcase tests, this test though does not create a client, instead, it fetches the client from existing Spring beans. So it's hard to override the endpoint in the tests since the whole point of SpringCodeGen is to create a client for customers.

I think this change might be unavoidable for now, and should not impact any existing features, because SpringCodeGen does not expose a property for customers to set endpoint or universe domain either. We can revisit this decision once we get to implement universe domain in spring-cloud-gcp. WDYT @lqiu96 ?

lqiu96: I see. I'm not too familiar with spring so I don't know of a way to do this, but we are going to have to remove the explicit setting for the default endpoint at some point (i.e. whenever we enable universe domain). This will require a workaround for the showcase clients in the future or more configurations generated by the generator.

I think for now we can proceed with explicitly setting the default endpoint.

spring-cloud-generator/scripts/generate-showcase.sh
```diff
@@ -1,111 +1,175 @@
...
-  cd sdk-platform-java && git checkout "v${GAPIC_GENERATOR_JAVA_VERSION}"
+  if [[ ! -d "./sdk-platform-java" ]]; then
+    git clone https://github.com/googleapis/sdk-platform-java.git
+  fi
+  pushd sdk-platform-java
+  git checkout "v${gapic_generator_java_version}"
+
+  # We will use the generation tools from library_generation
+  pushd library_generation/utils
+  source utilities.sh
```
blakeli0, line 73: How many utility methods in `utilities.sh` are we relying on? If it's not too much, I would prefer just duplicate the logic, because this creates dependencies of our hermetic build scripts, and we may accidentally break SpringCodeGen scripts.

spring-cloud-generator/scripts/generate-showcase.sh
```diff
@@ -1,111 +1,175 @@
...
+  popd #googleapis
+
+  # Now we call protoc with a series of arguments we obtain from
+  # sdk-platform-java's utilities.sh and others that are hardcoded (and stable).
+  # Note that --java_gapic_spring_opt uses `get_gapic_opts` which will work
+  # since the BUILD rules take similar arguments
+
+  proto_path="schema/google/showcase/v1beta1"
+  proto_files=$(find "${proto_path}" -type f  -name "*.proto" | LC_COLLATE=C sort)
+  gapic_additional_protos="google/iam/v1/iam_policy.proto google/cloud/location/locations.proto"
```
blakeli0, line 124: I don't think we need these additional protos, same thing for `rest_numeric_enums`, `transport`, `gapic_yaml`, `include_samples` etc. Can you try to remove them and see if the checks are still passing?

spring-cloud-generator/scripts/generate-showcase.sh
```diff
@@ -1,111 +1,175 @@
...
+  pushd library_generation/utils
+  source utilities.sh
+  popd #library_generation/utils
 
   # Install showcase client libraries locally
-  cd showcase && mvn clean install
-  GAPIC_SHOWCASE_CLIENT_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+  pushd showcase
+  # For local development, we cleanup any traces of previous runs
+  rm -rdf output
```
burkedavison, line 79: Everything below uses `${output_folder}`, but this uses literal value `output`.

burkedavison:
LGTM

spring-cloud-generator/scripts/generate-showcase.sh
diegomarquezp: @blakeli0 when removing the additional protos, the generated code has missing `iam` methods. For example:

```
diff -r /usr/local/google/home/diegomarquezp/Desktop/spring-cloud-gcp/spring-cloud-generator/scripts/../showcase/showcase-spring-starter-generated/src/main/java/com/google/showcase/v1beta1/spring/SequenceServiceSpringAutoConfiguration.java /usr/local/google/home/diegomarquezp/Desktop/spring-cloud-gcp/spring-cloud-generator/scripts/../showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/SequenceServiceSpringAutoConfiguration.java
98a99,101
>     if (this.clientProperties.getUseRest()) {
>       return SequenceServiceSettings.defaultHttpJsonTransportProviderBuilder().build();
>     }
123c126,134
<     SequenceServiceSettings.Builder clientSettingsBuilder = SequenceServiceSettings.newBuilder();
---
>     SequenceServiceSettings.Builder clientSettingsBuilder;
>     if (this.clientProperties.getUseRest()) {
>       clientSettingsBuilder = SequenceServiceSettings.newHttpJsonBuilder();
>       if (LOGGER.isTraceEnabled()) {
>         LOGGER.trace("Using REST (HTTP/JSON) transport.");
>       }
>     } else {
>       clientSettingsBuilder = SequenceServiceSettings.newBuilder();
>     }
186a198,224
>       RetrySettings listLocationsRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.listLocationsSettings().getRetrySettings(), serviceRetry);
>       clientSettingsBuilder.listLocationsSettings().setRetrySettings(listLocationsRetrySettings);
> 
>       RetrySettings getLocationRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.getLocationSettings().getRetrySettings(), serviceRetry);
>       clientSettingsBuilder.getLocationSettings().setRetrySettings(getLocationRetrySettings);
> 
>       RetrySettings setIamPolicyRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.setIamPolicySettings().getRetrySettings(), serviceRetry);
>       clientSettingsBuilder.setIamPolicySettings().setRetrySettings(setIamPolicyRetrySettings);
> 
>       RetrySettings getIamPolicyRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.getIamPolicySettings().getRetrySettings(), serviceRetry);
>       clientSettingsBuilder.getIamPolicySettings().setRetrySettings(getIamPolicyRetrySettings);
> 
>       RetrySettings testIamPermissionsRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.testIamPermissionsSettings().getRetrySettings(), serviceRetry);
>       clientSettingsBuilder
>           .testIamPermissionsSettings()
>           .setRetrySettings(testIamPermissionsRetrySettings);
> 
254a293,346
>       }
>     }
>     Retry listLocationsRetry = clientProperties.getListLocationsRetry();
>     if (listLocationsRetry != null) {
>       RetrySettings listLocationsRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.listLocationsSettings().getRetrySettings(), listLocationsRetry);
>       clientSettingsBuilder.listLocationsSettings().setRetrySettings(listLocationsRetrySettings);
>       if (LOGGER.isTraceEnabled()) {
>         LOGGER.trace("Configured method-level retry settings for listLocations from properties.");
>       }
>     }
>     Retry getLocationRetry = clientProperties.getGetLocationRetry();
>     if (getLocationRetry != null) {
>       RetrySettings getLocationRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.getLocationSettings().getRetrySettings(), getLocationRetry);
>       clientSettingsBuilder.getLocationSettings().setRetrySettings(getLocationRetrySettings);
>       if (LOGGER.isTraceEnabled()) {
>         LOGGER.trace("Configured method-level retry settings for getLocation from properties.");
>       }
>     }
>     Retry setIamPolicyRetry = clientProperties.getSetIamPolicyRetry();
>     if (setIamPolicyRetry != null) {
>       RetrySettings setIamPolicyRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.setIamPolicySettings().getRetrySettings(), setIamPolicyRetry);
>       clientSettingsBuilder.setIamPolicySettings().setRetrySettings(setIamPolicyRetrySettings);
>       if (LOGGER.isTraceEnabled()) {
>         LOGGER.trace("Configured method-level retry settings for setIamPolicy from properties.");
>       }
>     }
>     Retry getIamPolicyRetry = clientProperties.getGetIamPolicyRetry();
>     if (getIamPolicyRetry != null) {
>       RetrySettings getIamPolicyRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.getIamPolicySettings().getRetrySettings(), getIamPolicyRetry);
>       clientSettingsBuilder.getIamPolicySettings().setRetrySettings(getIamPolicyRetrySettings);
>       if (LOGGER.isTraceEnabled()) {
>         LOGGER.trace("Configured method-level retry settings for getIamPolicy from properties.");
>       }
>     }
>     Retry testIamPermissionsRetry = clientProperties.getTestIamPermissionsRetry();
>     if (testIamPermissionsRetry != null) {
>       RetrySettings testIamPermissionsRetrySettings =
>           RetryUtil.updateRetrySettings(
>               clientSettingsBuilder.testIamPermissionsSettings().getRetrySettings(),
>               testIamPermissionsRetry);
>       clientSettingsBuilder
>           .testIamPermissionsSettings()
>           .setRetrySettings(testIamPermissionsRetrySettings);
>       if (LOGGER.isTraceEnabled()) {
>         LOGGER.trace(
>             "Configured method-level retry settings for testIamPermissions from properties.");

```

diegomarquezp: Removing `rest_numeric_enums` causes issues too:
```
diff -r /usr/local/google/home/diegomarquezp/Desktop/spring-cloud-gcp/spring-cloud-generator/scripts/../showcase/showcase-spring-starter-generated/src/main/java/com/google/showcase/v1beta1/spring/TestingSpringAutoConfiguration.java /usr/local/google/home/diegomarquezp/Desktop/spring-cloud-gcp/spring-cloud-generator/scripts/../showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/TestingSpringAutoConfiguration.java
96a97,99
>     if (this.clientProperties.getUseRest()) {
>       return TestingSettings.defaultHttpJsonTransportProviderBuilder().build();
>     }   
118c121,129
<     TestingSettings.Builder clientSettingsBuilder = TestingSettings.newBuilder();
---
>     TestingSettings.Builder clientSettingsBuilder;
>     if (this.clientProperties.getUseRest()) {
>       clientSettingsBuilder = TestingSettings.newHttpJsonBuilder();
>       if (LOGGER.isTraceEnabled()) {
>         LOGGER.trace("Using REST (HTTP/JSON) transport.");
>       }
>     } else {
>       clientSettingsBuilder = TestingSettings.newBuilder();
>     }   

```

I'll keep checking on the other params

spring-cloud-generator/scripts/generate-showcase.sh
diegomarquezp: I'm moving them to `generate-showcase-utilities`

spring-cloud-generator/scripts/generate-showcase.sh
diegomarquezp: `include_samples` and `gapic_yaml` were unnecessary. The other ones produce changes if removed.

blakeli0: SG, we can keep the ones produce changes.

blakeli0 (approved):

PR #2985, opened 2024-06-24
chore: fix autoconfig module generation

diegomarquezp:
This fixes the autoconfig generation by excluding two libraries that do not have a `java_gapic_library` rule in their BUILD.bazel files ([example](https://github.com/googleapis/googleapis/blob/master/google/cloud/alloydb/connectors/v1/BUILD.bazel) for alloydbconnectors).

### DIFF
--- a/spring-cloud-generator/scripts/generate-library-list.sh
+++ b/spring-cloud-generator/scripts/generate-library-list.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -ex
+
 # If not set, assume working directory is spring-cloud-generator
 if [[ -z "$SPRING_GENERATOR_DIR" ]]; then
   SPRING_GENERATOR_DIR=`pwd`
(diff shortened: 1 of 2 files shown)


spring-cloud-generator/scripts/resources/manual_modules_exclusion_list.txt
```diff
@@ -11,3 +11,5 @@ google-cloud-secretmanager
 google-cloud-resourcemanager
 google-cloud-storage
 google-cloud-vision
+google-cloud-alloydb-connectors
```
diegomarquezp, line 14: These two libraries had compilation errors and I'm excluding them temporarily

spring-cloud-generator/scripts/generate-library-list.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -ex
```
diegomarquezp, line 3: `-e` will reduce the log output and help us isolate the error by exiting earlier.
`-o` will save us from a local run due to lack of output in the runner log.

diegomarquezp:
Had to do a `--force` push of a rebase since `git checkout origin/main -- spring-cloud-previews` (with a previous `git fetch`) did reset the folders but they were still showing as changed in the PR

JoeWang1127 (approved):

PR #3004, opened 2024-06-27
Test only - Minimal Spring Boot 3.3.x updates

burkedavison:


burkedavison:
Investigation complete. Minimal edits to maintain trace's compatibility with Spring Boot 3.2.x is not possible.

PR #3014, opened 2024-06-28
ci: add Spring Boot 3.2.5 testing

burkedavison:


.github/workflows/integrationTests.yaml
```diff
@@ -118,3 +118,111 @@ jobs:
...
+          - multisample
+          - pubsub
+          - pubsub-bus
+          - pubsub-docs
+          - pubsub-emulator
+          - pubsub-integration
+          - secretmanager
+          - spanner
+          - storage
+          # trace excluded
```
meltsufin, line 151: Should we include it, but explicitly skip it so it still shows up, but as skipped?

meltsufin (approved):

.github/workflows/integrationTests.yaml
burkedavison: I'm guessing the conditional execution check occurs at the job level, not the individual strategy level. If so, the strategy variable might not be in scope... but giving it a try.

PR #3019, opened 2024-07-01
ci: skip trace test rather than excluding

burkedavison:
Skips the trace test from 3.2.5 testing rather than implicitly excluding it.


burkedavison:
Using 'exclude' does not show a 'skipped' status in CI check list.

Issue #2696, opened 2024-03-13
PubSub message conversion exception is silently swallowed


minogin:
`com.google.cloud:spring-cloud-gcp-pubsub:5.0.4`

When using `com.google.cloud.spring.pubsub.core.subscriber.PubSubSubscriberTemplate#pullAndConvertAsync` and an exception happens during message conversion the future just never completes and you never get neither exception nor the message.

```
@Override
  public <T> CompletableFuture<List<ConvertedAcknowledgeablePubsubMessage<T>>> pullAndConvertAsync(
      String subscription, Integer maxMessages, Boolean returnImmediately, Class<T> payloadType) {
    final CompletableFuture<List<ConvertedAcknowledgeablePubsubMessage<T>>> completableFuture =
        new CompletableFuture<>();

    this.pullAsync(subscription, maxMessages, returnImmediately)
        .whenComplete(
            (ackableMessages, exception) -> {
                if (exception != null) {
                  completableFuture.completeExceptionally(exception);
                  return;
                }
                completableFuture.complete(
                    this.toConvertedAcknowledgeablePubsubMessages(payloadType, ackableMessages));
            });

    return completableFuture;
  }
```

The problem is here:
```
completableFuture.complete(
                    this.toConvertedAcknowledgeablePubsubMessages(payloadType, ackableMessages));
```

when `toConvertedAcknowledgeablePubsubMessages` throws an exception.

ldetmer:
Fixed in release 5.5.1

PR #3091, opened 2024-07-30
fix: check for exception when parsing pub sub response and return it (#2696)

ldetmer:
* fix: check for exception when parsing pub sub response and return it (#2696 )

### DIFF
--- a/spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
+++ b/spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
@@ -17,6 +17,7 @@
 package com.google.cloud.spring.pubsub.core.subscriber;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.same;
 import static org.mockito.Mockito.doAnswer;
@@ -59,8 +60,10 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
```diff
@@ -530,6 +531,21 @@ void testPullAndConvertAsync()
...
+  @Test
+  void testPullAndConvertAsync_publishesConvertException()
+          throws InterruptedException, ExecutionException, TimeoutException {
+    when(this.messageConverter.fromPubSubMessage(this.pubsubMessage, BigInteger.class)).thenThrow(new NullPointerException());
+    CompletableFuture<List<ConvertedAcknowledgeablePubsubMessage<BigInteger>>> asyncResult =
+            this.pubSubSubscriberTemplate.pullAndConvertAsync("sub2", 1, true, BigInteger.class);
+
+    try {
+      asyncResult.get();
+      Assertions.fail();
```
suztomo, line 543: Add a message that appears when this assertion fails.

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
```diff
@@ -530,6 +531,21 @@ void testPullAndConvertAsync()
...
 
+  @Test
+  void testPullAndConvertAsync_publishesConvertException()
+          throws InterruptedException, ExecutionException, TimeoutException {
+    when(this.messageConverter.fromPubSubMessage(this.pubsubMessage, BigInteger.class)).thenThrow(new NullPointerException());
+    CompletableFuture<List<ConvertedAcknowledgeablePubsubMessage<BigInteger>>> asyncResult =
+            this.pubSubSubscriberTemplate.pullAndConvertAsync("sub2", 1, true, BigInteger.class);
+
+    try {
+      asyncResult.get();
```
suztomo, line 542: Does the test fail if there's no change in spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplate.java?

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
ldetmer: done

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
```diff
@@ -530,6 +531,21 @@ void testPullAndConvertAsync()
...
+      Assertions.fail("NullPointerException should have been thrown and was not!");
```
burkedavison, line 543: No change requested, just a consideration: Another way to write this check is to use the `assertThrows` function which has the benefit of providing an automatic clear test failure synopsis.

```java
ExecutionException e = assertThrows(ExecutionException.class, () -> asyncResult.get());
assertThat(e.getCause()).isInstanceOf(NullPointerException.class);
```

burkedavison:
LGTM once Tomo gives the thumbs up.

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
ldetmer: Good catch, updated to add a timeout to fail in the case that the response hangs (no change to PubSubSubscriberTemplate)

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
ldetmer: nice that's cleaner! applied suggestion

spring-cloud-gcp-pubsub/src/test/java/com/google/cloud/spring/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
```diff
@@ -530,6 +533,17 @@ void testPullAndConvertAsync()
     assertThat(result.get(0).getProjectSubscriptionName().getSubscription()).isEqualTo("sub2");
   }
 
+  @Test
+  @Timeout(value = 5)
```
suztomo, line 537: Would you add a unit? Without a unit, I had to look up the API reference (https://junit.org/junit5/docs/5.5.1/api/org/junit/jupiter/api/Timeout.html).

suztomo (approved):
Would you add a unit and them merge this?

burkedavison (approved):

Issue #3139, opened 2024-08-19
PubSubDeadLetterTopicSampleAppIntegrationTest failed with cast class exception


ldetmer:
Error running [PubSubDeadLetterTopicSampleAppIntegrationTest: testSample_deadLetterHandling](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/10449601755/job/28932260344#step:9:11800)

```
08:24:25.893 [global-gcp-pubsub-subscriber1] WARN  c.g.c.s.p.i.i.PubSubInboundChannelAdapter - error occurred in message handler [org.springframework.cloud.stream.function.FunctionConfiguration$FunctionToDestinationBinder$1@68347216]
08:24:29.895 [global-gcp-pubsub-subscriber3] ERROR o.s.i.handler.LoggingHandler - org.springframework.messaging.MessageHandlingException: error occurred in message handler [org.springframework.cloud.stream.function.FunctionConfiguration$FunctionToDestinationBinder$1@68347216], failedMessage=GenericMessage [payload=byte[127], headers=***id=47898e41-b4de-a94b-a74d-d86064f5ffe0, contentType=application/json, gcp_pubsub_original_message=PushedAcknowledgeablePubsubMessage***projectId='spring-cloud-gcp-ci', subscriptionName='anonymous.my-main-topic.32f1ddf2-791f-410f-bf55-dfd4ffd5b58b', message=data: "***\"body\":\"test message 5d63cca5-980e-4e83-ae2e-4eb14f459c65\",\"username\":\"testUserName\",\"createdAt\":[2024,8,19,8,24,6,337013596]***"
attributes ***
  key: "contentType"
  value: "application/json"
***
attributes ***
  key: "googclient_deliveryattempt"
  value: "3"
***
message_id: "11994779617011169"
publish_time ***
  seconds: 1724055846
  nanos: 468000000
***
***, googclient_deliveryattempt=3, timestamp=1724055869894***]
	at org.springframework.integration.support.utils.IntegrationUtils.wrapInHandlingExceptionIfNecessary(IntegrationUtils.java:191)
	at org.springframework.integration.handler.AbstractMessageHandler.doHandleMessage(AbstractMessageHandler.java:108)
	at org.springframework.integration.handler.AbstractMessageHandler.handleMessage(AbstractMessageHandler.java:73)
[...]

PR #3145, opened 2024-08-21
fix: CastClassException in PubSubDeadLetterTopicSampleAppIntegrationTest #3139

ldetmer:
Added object mapper bean to [fix issue with spring cloud stream 4.1.3 failing to load objects in application context when loading their ObjectMapper](https://github.com/spring-cloud/spring-cloud-stream/issues/2977) causing PubSubDeadLetterTopicSampleAppIntegrationTest when trying to deserialize the object from the stream.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-dead-letter-sample/src/main/java/com/example/SinkExample.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-dead-letter-sample/src/main/java/com/example/SinkExample.java
@@ -67,6 +67,9 @@ public class SinkExample {
             userMessage.getBody());
   }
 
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-dead-letter-sample/src/main/java/com/example/SinkExample.java
```diff
@@ -63,4 +66,9 @@ public Consumer<UserMessage> deadLetterMessages() {
             userMessage.getUsername(),
             userMessage.getBody());
   }
+
+  @Bean
+  public JsonMessageConverter customJsonMessageConverter(ObjectMapper objectMapper) {
```
burkedavison, line 71: Can we add a comment here pointing to either our bug or https://github.com/spring-cloud/spring-cloud-stream/issues/2977 ?

Issue #3142, opened 2024-08-20
spring-cloud-gcp-data-datastore-sample fails with "com.google.cloud.datastore.DatastoreException: The project XXX does not exist."


Kehrlann:
Running the `spring-cloud-gcp-data-datastore-sample` project fails with: `com.google.cloud.datastore.DatastoreException: The project XXX does not exist."`

## Setup

Running:
- Java 21
- This project at commit 7733fe54158ea14af5e27952de0d3efcd0a79004
- The following config:

```
spring.cloud.gcp.datastore.credentials.location=classpath:credentials.json
spring.cloud.gcp.datastore.database-id=my-db
```

- GCP API key with role `Cloud Datastore User`
    - Edit: even when using ACD with my account, I get the same error

## Bug:

Running the following command:

```
./mvnw -pl spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample spring-boot:run
```

Fails with:

```
com.google.cloud.datastore.DatastoreException: The project <REDACTED> does not exist.
	at com.google.cloud.datastore.spi.v1.HttpDatastoreRpc.translate(HttpDatastoreRpc.java:140) ~[google-cloud-datastore-2.21.1.jar:2.21.1]
	at com.google.cloud.datastore.spi.v1.HttpDatastoreRpc.translate(HttpDatastoreRpc.java:125) ~[google-cloud-datastore-2.21.1.jar:2.21.1]
	at com.google.cloud.datastore.spi.v1.HttpDatastoreRpc.runQuery(HttpDatastoreRpc.java:202) ~[google-cloud-datastore-2.21.1.jar:2.21.1]
	at com.google.cloud.datastore.DatastoreImpl.lambda$runQuery$0(DatastoreImpl.java:242) ~[google-cloud-datastore-2.21.1.jar:2.21.1]
	at com.google.api.gax.retrying.DirectRetryingExecutor.submit(DirectRetryingExecutor.java:102) ~[gax-2.51.0.jar:2.51.0]
	at com.google.cloud.RetryHelper.run(RetryHelper.java:76) ~[google-cloud-core-2.41.0.jar:2.41.0]
	at com.google.cloud.RetryHelper.runWithRetries(RetryHelper.java:50) ~[google-cloud-core-2.41.0.jar:2.41.0]
[...]

PR #3156, opened 2024-08-23
fix: Descendants entities losing database ID from parent (#3142)

ldetmer:
If you override the default database ID and are using `@Descendants` annotation the database ID wasn't getting copied over when migrating keys from ancestor to child


spring-cloud-gcp-data-datastore/src/main/java/com/google/cloud/spring/data/datastore/core/util/KeyUtil.java
```diff
@@ -46,6 +46,7 @@ public static Key getKeyWithoutAncestors(Key entityKey) {
           Key.newBuilder(entityKey.getProjectId(), entityKey.getKind(), entityKey.getId());
     }
     ancestorLookupKey.setNamespace(entityKey.getNamespace());
+    ancestorLookupKey.setDatabaseId(entityKey.getDatabaseId());
```
burkedavison, line 49: It's too bad the ancestors can't be explicitly cleared in the builder, or we'd be able to use the copy constructors to avoid direct knowledge of all the individual fields.
```java
public static Key getKeyWithoutAncestors(Key entityKey) {
  return Key.newBuilder(entityKey).withoutAncestors().build();
}

burkedavison (approved):

Issue #2785, opened 2024-04-15
TransportChannelProvider bean overwrites maxInboundMetadataSize with null


ivan-zaitsev:
**Describe the bug**
[TransportChannelProvider](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java#L437) bean overwrites `maxInboundMetadataSize` with null by not setting it. It causes `io.grpc.netty.shaded.io.netty.handler.codec.http2.Http2Exception: Header size exceeded max allowed size (10240)` error because it fallbacks to default 10KiB instead of 4MiB.

**Sample**
The library has default 4MB:
https://github.com/googleapis/java-pubsub/blob/main/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java#L487

diegomarquezp:
Hi @ivan-zaitsev, thanks for reporting this issue. Would you mind sharing how you are using the bean and any properties file you are working with? Thanks!

@burkedavison @meltsufin shouldn't a config file override this default setting?

ivan-zaitsev:
Hello @diegomarquezp, default config.

```
spring:
  cloud:
    gcp:
      project-id: ...
      credentials:
        encoded-key: ...
```

You can trace bean [TransportChannelProvider](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java#L437) usage and will find that it will overwrite default 4MB value with null.

meltsufin:
@diegomarquezp I agree that we should keep the defaults set by Pub/Sub, unless they are overwritten by configuration or we intentionally want to change them in Spring Cloud GCP. In this particular case, we are unintentionally overriding the defaults and should fix it.

PR #3157, opened 2024-08-23
fix: set `maxInboundMetadataSize` in pubsub 

JoeWang1127:
Fix: #2785


zhumin8 (commented):
Do you mind adding a comment citing the pubsub client library default? 
Also, do you think this default behavior is worth mentioning somewhere in the documentation?

JoeWang1127:
> Also, do you think this default behavior is worth mentioning somewhere in the documentation?

Do you mean in SCGCP docs?

zhumin8:
> Do you mean in SCGCP docs?

Yes, I am referring to SCGCP docs. In light of "secure by defaults", although we are hardcoding in this case, IMO it probably still worth explicitly spell out that we are following PubSub client library defaults for these settings.

Another question, should we also set "maxInboundMessageSize" as client library does [here](https://github.com/googleapis/java-pubsub/blob/69bf4e6abc7ea52f40da0a7ccaae2f2eab446630/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java#L486)? And what about `publisherTransportChannelProvider` setting? At a glance, it looks like SCGCP is not keeping consistent with client library defaults [here](https://github.com/googleapis/java-pubsub/blob/69bf4e6abc7ea52f40da0a7ccaae2f2eab446630/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Publisher.java#L732)?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -436,6 +436,9 @@ public SubscriptionAdminClient subscriptionAdminClient(
   @ConditionalOnMissingBean(name = "subscriberTransportChannelProvider")
   public TransportChannelProvider subscriberTransportChannelProvider() {
     return SubscriberStubSettings.defaultGrpcTransportProviderBuilder()
+        // default value specified by pubsub client library,
+        // see https://github.com/googleapis/java-pubsub/blob/main/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java#L487.
+        .setMaxInboundMetadataSize(4 * 1024 * 1024)
```
mpeddada1, line 441: Is it possible to reference the default variable here instead of hardcoding the default value? Just so that we are aligned with the default in the client in case it changes in the future.

mpeddada1: Thanks for checking, That is a little unfortunate that it can't be referenced. It might be worth creating a ticket in the client repo (similar to https://github.com/googleapis/java-pubsub/issues/373)

mpeddada1:
> Also, do you think this default behavior is worth mentioning somewhere in the documentation?

+1 how about we document this in https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#publishersubscriber-configuration?

JoeWang1127:
> +1 how about we document this in https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#publishersubscriber-configuration?

I didn't add this property in [GcpPubSubProperties](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/e04f755f75c98619c11fde42106cff81e95f1540/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubProperties.java#L28) and I don't see `TransportChannelProvider` can be custom configured in the existing pubsub properties.

I think we shouldn't surface this property in the doc, what do you think?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
JoeWang1127: I think the builder is autogenerated by the generator: https://github.com/googleapis/java-pubsub/blob/main/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/SubscriptionAdminSettings.java#L246 and https://github.com/googleapis/java-pubsub/blob/main/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/stub/SubscriberStubSettings.java#L425

mpeddada1:
> > +1 how about we document this in https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#publishersubscriber-configuration?
> 
> I didn't add this property in [GcpPubSubProperties](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/e04f755f75c98619c11fde42106cff81e95f1540/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/GcpPubSubProperties.java#L28) and I don't see `TransportChannelProvider` can be custom configured in the existing pubsub properties.
> 
> I think we shouldn't surface this property in the doc, what do you think?

Agreed that this particular property can't be configured so adding it as part of the configurations table might be confusing but I think users can still customize the `subscriberTransportChannelProvider` by providing their own bean. It might still be worth adding "Note" section in the documentation which mentions this default value? As an example, we have a note that mentions the default thread names we [use](https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#publishersubscriber-configuration):

```
By default, subscription-specific threads are named after fully-qualified subscription name, ex: gcp-pubsub-subscriber-projects/project-id/subscriptions/subscription-name. This can be customized, by registering a SelectiveSchedulerThreadNameProvider bean.
```

JoeWang1127:
> It might still be worth adding "Note" section in the documentation which mentions this default value

Added a note in the pubsub doc.

mpeddada1 (approved):

PR #3158, opened 2024-08-27
feat: add properties to set universe domain and endpoint in bigquery

mpeddada1:
This PR introduces the following changes:

- Introduces `spring.cloud.gcp.bigquery.universe-domain` property to customize the universe domain in `BigqueryOptions` (option to customize endpoint is not provided in the client library) and `BigQueryWriteClient` if these values aren't `null`. 
- Introduces `spring.cloud.gcp.bigquery.endpoint` property to customize the endpoint and host in `BigQueryWriteClient` and if this value isn't `null`. Since Biqguery accepts `host` instead of `endpoint`, the endpoint is formatted to the `https://service.universeDomain` format before being set to Bigquery. 
- Adds unit tests in `GcpBigQueryAutoConfigurationTests` verify that the client settings have been correctly set. 

End-to-end verification with test environment endpoints have been documented in a separate test plan document.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
@@ -162,14 +162,10 @@ public class GcpBigQueryAutoConfiguration {
   }
 
   private String resolveToHost(String endpoint) throws URISyntaxException {
-    if (!endpoint.contains("://")) {
-      endpoint = "https://" + endpoint;
+    int portIndex = endpoint.indexOf(":");
+    if (portIndex != -1){
+      return "https://" + endpoint.substring(0,portIndex) + "/";
     }
-    URI uri = new URI(endpoint);
-
-    // Construct the new URL with https:// and no port
-    String newUrl = new URI("https", uri.getUserInfo(), uri.getHost(), -1,
-        uri.getPath(), uri.getQuery(), uri.getFragment()).toString();
-    return newUrl;
+    return "https://" + endpoint + "/";
   }
 }
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
@@ -26,7 +26,6 @@ import com.google.cloud.spring.bigquery.core.BigQueryTemplate;
 import com.google.cloud.spring.core.DefaultCredentialsProvider;
 import com.google.cloud.spring.core.GcpProjectIdProvider;
 import com.google.cloud.spring.core.UserAgentHeaderProvider;
-import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
@@ -57,6 +57,9 @@ public class GcpBigQueryAutoConfiguration {
 
   private int threadPoolSize;
 
+  private String universeDomain;
+  private String endpoint;
+
   GcpBigQueryAutoConfiguration(
       GcpBigQueryProperties gcpBigQueryProperties,
       GcpProjectIdProvider projectIdProvider,
@@ -78,6 +81,9 @@ public class GcpBigQueryAutoConfiguration {
     this.jsonWriterBatchSize = gcpBigQueryProperties.getJsonWriterBatchSize();
 
     this.threadPoolSize = getThreadPoolSize(gcpBigQueryProperties.getThreadPoolSize());
+
+    this.universeDomain = gcpBigQueryProperties.getUniverseDomain();
+    this.endpoint = gcpBigQueryProperties.getEndpoint();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
```diff
@@ -57,6 +57,11 @@ public class GcpBigQueryAutoConfiguration {
 
   private int threadPoolSize;
 
+  private String universeDomain;
+  private String jsonWriterEndpoint;
+
+  private String host;
```
mpeddada1, line 63: Calling out the addition of this property since this is a deviation from our original design of exposing universe domain and endpoint.  Apiary libraries such as Bigquery and Storage only allow for customizing `host` (which is essentially endpoint without the port) instead of `endpoint` so we are doing the same in the Spring Cloud GCP layer. For context: https://github.com/googleapis/sdk-platform-java/pull/2329

@burkedavison @blakeli0

burkedavison (approved):

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
```diff
@@ -97,25 +106,35 @@ private int getThreadPoolSize(int threadPoolSize) {
...
             .setQuotaProjectId(this.projectId)
-            .setHeaderProvider(new UserAgentHeaderProvider(GcpBigQueryAutoConfiguration.class))
-            .build();
-    return BigQueryWriteClient.create(bigQueryWriteSettings);
+            .setHeaderProvider(new UserAgentHeaderProvider(GcpBigQueryAutoConfiguration.class));
+    if (this.universeDomain != null) {
+      bigQueryWriteSettingsBuilder.setUniverseDomain(this.universeDomain);
+    }
+    if (this.jsonWriterEndpoint != null) {
+      bigQueryWriteSettingsBuilder.setEndpoint(this.jsonWriterEndpoint);
```
blakeli0, line 135: Looks like `jsonWriterEndpoint` is only used for `BigQueryWriteClient`, which is actually part of bigquery-storage, not bigquery. Ideally I think we should have different auto-configuration for them, but since they are in one auto-configuration, can we either expose `host` or `endpoint`? And construct the other dynamically?

mpeddada1: That's right, it's only used for BigqueryWriteClient (bigquery storage). We can expose `endpoint` and strip the port and reformat it to the pattern that [`DEFAULT_HOST`](https://github.com/googleapis/sdk-platform-java/blob/1a988df22f7e3d15ce6b121bf26897c59ab468e4/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L86) currently has when setting it to BigqueryOptions. 

However, the concern I have with this is that it would prevent us from making endpoint/host configurable on a per-service basis. In fact, looking at the code again, I'm wondering if it was a mistake to make universe domain shared between bigquery and bigquerystorage.

blakeli0: Do we share anything else between Bigquery and Bigquerystorage client? Ideally they should be separate as they are two different services, but looks like we are treating them the same in spring-cloud-gcp.

mpeddada1: That's a good question - you're right, they do seem pretty interconnected in spring-cloud-gcp.
 `BigqueryTemplate` [wraps both Bigquery and BigQueryWriteClient](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/f7fc095a55966a910693718deaa74d9bd2a0d2f8/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java#L102) and both services share the `spring.cloud.gcp.bigquery.threadPoolSize` property and `spring.cloud.gcp.bigquery.datasetName` property. Dataset name is used by the [`BigtableTemplate#writeJsonStream`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blame/f7fc095a55966a910693718deaa74d9bd2a0d2f8/spring-cloud-gcp-bigquery/src/main/java/com/google/cloud/spring/bigquery/core/BigQueryTemplate.java#L218C46-L218C55) method which uses Bigquery to create a table before invoking the BigQuery Storage Write API. 

`spring.cloud.gcp.bigquery.jsonWriterBatchSize` is the only property that is unique to bigquery storage.

blakeli0: I think we have two options
1. Continue treating bigquery and bigquerystorage the same product. Then we only introduce `universe-domain`, and either `host` or `endpoint`.
2. Start treating them two different product. Then we would introduce `universe-domain` and `host` for bigquery, `universe-domain` and `endpoint` for bigquerystorage.
Any preferences? cc: @burkedavison

mpeddada1: As discussed offline, considering that both bigquery and bigquerystorage are wrapped as one product in spring-cloud-gcp, we will expose global endpoint and universe domain properties. The port will be omitted from the endpoint when set to the bigquery client.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfigurationTests.java
```diff
@@ -60,6 +61,107 @@ void testSettingBigQueryOptions() {
...
+  @Test
+  void testBigQueryWrite_endpoint() {
+    this.contextRunner
+        .withPropertyValues(
+            "spring.cloud.gcp.bigquery.endpoint=bigquerystorage.example.com:123")
+        .run(
+            ctx -> {
+              BigQueryWriteClient client = ctx.getBean(BigQueryWriteClient.class);
+              assertThat(client.getSettings().getEndpoint())
+                  .isEqualTo("bigquerystorage.example.com:123");
```
burkedavison, line 135: Please verify that it's intended to have getEndpoint not return the URI scheme, but getResolvedApiaryHost does return the URI scheme.

LGTM

burkedavison (approved):

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfigurationTests.java
mpeddada1: Yes, returning endpoint with this pattern is intended for Bigquerystorage which will otherwise throw an exception if it doesn't follow a specific convention. For example, passing in `bigquerystorage.example.com` without the port will result in:
```
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.google.cloud.bigquery.storage.v1.BigQueryWriteClient]: Factory method 'bigQueryWriteClient' threw exception with message: invalid endpoint, expecting "<host>:<port>"
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:178)
	at org.springframework.beans.factory.support.ConstructorResolver.instantiate(ConstructorResolver.java:644)
	... 31 more
Caused by: java.lang.IllegalArgumentException: invalid endpoint, expecting "<host>:<port>"
```
On the other hand bigquery accepts either accepts user-provided endpoint as-is if one is provided or builds the host in the `https://service.universeDomain` format  if only the universe domain is provided:
https://github.com/googleapis/sdk-platform-java/blob/1a988df22f7e3d15ce6b121bf26897c59ab468e4/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L868 but [doesn't contain an explicit check to verify this schema](https://github.com/googleapis/java-bigquery/blob/bb2cd471fda5a2218580c41839c09edc5fa6b483/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java#L110). And in this PR, we reformat the provided endpoint to follow the same pattern that getResolvedApiaryHost() defaults to.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/bigquery/GcpBigQueryAutoConfiguration.java
```diff
@@ -141,4 +160,16 @@ public BigQueryTemplate bigQueryTemplate(
...
         bigQuery, bigQueryWriteClient, bqInitSettings, bigQueryThreadPoolTaskScheduler);
   }
+
+  private String resolveToHost(String endpoint) throws URISyntaxException {
+    if (!endpoint.contains("://")) {
+      endpoint = "https://" + endpoint;
+    }
+    URI uri = new URI(endpoint);
+
+    // Construct the new URL with https:// and no port
```
blakeli0, line 170: Can we just do String manipulation is host is basically endpoint without ports? Is there any edge cases can not be handled by String manipulation?

mpeddada1: Done. Simplified to use String manipulation.

blakeli0 (approved):

PR #3184, opened 2024-09-03
chore: fix javadoc job by updating to the new javadoc output path

mpeddada1:
The default javadoc output path was changed from target/site to target/reports in https://github.com/apache/maven-reporting-impl/pull/26 and was introduced in maven-javadoc-plugin:3.10.0 in https://github.com/apache/maven-javadoc-plugin/pull/204.

The maven-javadoc-plugin was updated from 3.8.0 to 3.10.0 in v5.6.0 ([PR](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3164)). This fix aims to address the current failure in the publish_javadoc job:
```

# Move into generated docs directory
pushd target/site/apidocs/
github/spring-cloud-gcp/.kokoro/publish_javadoc.sh: line 29: pushd: target/site/apidocs/: No such file or directory


[ID: 8846228] Command finished after 191 secs, exit value: 1
```

### DIFF
--- a/.kokoro/publish_javadoc.sh
+++ b/.kokoro/publish_javadoc.sh
@@ -19,7 +19,7 @@ python3 -m pip install --require-hashes -r .kokoro/requirements.txt
 mvn clean javadoc:aggregate -Drelease=true
 
 # Move into generated docs directory
-pushd target/site/apidocs/
+pushd target/reports/apidocs/
 
 python3 -m docuploader create-metadata \
      --name spring-cloud-gcp \


zhumin8 (commented):
Noticed the change was in "default value of output directory". I wonder if we can explicitly set this path in `maven-javadoc-plugin`, so we don't suffer similar issue in the future.

meltsufin (approved):

mpeddada1:
> Noticed the change was in "default value of output directory". I wonder if we can explicitly set this path in `maven-javadoc-plugin`, so we don't suffer similar issue in the future.

That's a great idea! There is an option to specify the `<outputDirectory>` in `maven-javadoc-plugin`. However, modifying this [plugin](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/38cfb06612cb6a7dd5cc4bf76000b3216501fa28/pom.xml#L760-L777) to include the following snippet doesn't seem to address the error seen above (at least when I tested it locally):
```
<configuration>
	<outputDirectory>${project.build.directory}/site</outputDirectory>
</configuration>
```

Perhaps we can explore this long term solution as a follow-up?

zhumin8 (approved):

suztomo:
> The default javadoc output path was changed from target/site to target/reports in https://github.com/apache/maven-reporting-impl/pull/26 and was introduced in maven-javadoc-plugin:3.10.0 in https://github.com/apache/maven-javadoc-plugin/pull/204.

That's great of you finding the source of the behavior change in upstream projects.

Issue #3290, opened 2024-10-06
Spanner: Flaky Test: SpannerStatementQueryExecutor.getChildrenRowsQuery


williamlin0825:
A test in the getChildrenRowsQueryTest checks for the result from getSQL() from SpannerStatementQueryExecutor.getChildrenRowsQuery. However, it checks the result in a specific order, which is not guaranteed. I purpose to simply check whether the selected field names are correct instead of checking in a specific order.

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/05ad3efab8b70f2235e63f98dec3bc7e35f00954/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/repository/query/SpannerQueryLookupStrategyTests.java#L176-L179

```
[INFO] 
INFO: Adding excluded groups to newly created one
INFO: Adding NonDex argLine to existing argLine specified by the project
CONFIG: nondexFilter=.*
nondexMode=FULL
nondexSeed=933178
nondexStart=0
nondexEnd=9223372036854775807
nondexPrintstack=false
nondexDir=/home/jhihwei2/spring-cloud-gcp/spring-cloud-gcp-data-spanner/.nondex
nondexJarDir=/home/jhihwei2/spring-cloud-gcp/spring-cloud-gcp-data-spanner/.nondex
nondexExecid=fBBMREC82yfCbEu7MM71JCaMC87SgB9c1lrSr1Wi5nY=
nondexLogging=CONFIG
test=
[INFO] Using auto detected provider org.apache.maven.surefire.junitplatform.JUnitPlatformProvider
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running com.google.cloud.spring.data.spanner.repository.query.SpannerQueryLookupStrategyTests
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.198 s <<< FAILURE! -- in com.google.cloud.spring.data.spanner.repository.query.SpannerQueryLookupStrategyTests
[ERROR] com.google.cloud.spring.data.spanner.repository.query.SpannerQueryLookupStrategyTests.getChildrenRowsQueryTest -- Time elapsed: 3.119 s <<< FAILURE!
org.opentest4j.AssertionFailedError: 

[...]

PR #3297, opened 2024-10-11
test(fix): refactor SQL query assertions in SpannerStatementQueryExecutor.getChildrenRowsQuery to prevent flakiness

williamlin0825:
Related to issue: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3290

A test in the API `getChildrenRowsQueryTest` checks for the result from `getSQL()` from `SpannerStatementQueryExecutor.getChildrenRowsQuery`. However, it checks the result in a specific order, which is not guaranteed. Here is the affected test:

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/05ad3efab8b70f2235e63f98dec3bc7e35f00954/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/repository/query/SpannerQueryLookupStrategyTests.java#L176-L179

This PR propose to simply check whether the selected field names are correct instead of checking in a specific order.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/repository/query/SpannerQueryLookupStrategyTests.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/repository/query/SpannerQueryLookupStrategyTests.java
@@ -174,9 +174,13 @@ class SpannerQueryLookupStrategyTests {
             new SpannerWriteConverter(),
             this.spannerMappingContext);
     assertThat(statement.getSql())
-        .isEqualTo(
-            "SELECT deleted, id3, id, id_2 FROM child_test_table WHERE ((id = @tag0 AND id_2 ="
-                + " @tag1)) AND (deleted = false)");
+        .contains("SELECT ")
+        .contains("deleted")
+        .contains("id3")
+        .contains("id_2")
+        .contains("id")
+        .contains(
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/repository/query/SpannerQueryLookupStrategyTests.java
```diff
@@ -174,9 +174,13 @@ void getChildrenRowsQueryTest() {
...
             this.spannerMappingContext);
     assertThat(statement.getSql())
-        .isEqualTo(
-            "SELECT deleted, id3, id, id_2 FROM child_test_table WHERE ((id = @tag0 AND id_2 ="
-                + " @tag1)) AND (deleted = false)");
+        .contains("SELECT ")
+        .contains("deleted")
+        .contains("id3")
+        .contains("id_2")
+        .contains("id")
```
mpeddada1, line 181: This _may_ cause us to under-assert on the sql string returned as it appears to be verifying whether or not "deleted" or "id3" are present but doesn't enforce a query structure.

mpeddada1:
Thank you for putting up a PR! However, given our current observations in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3290#issuecomment-2770517740, we're going to hold off on making changes to the unit test.

PR #3312, opened 2024-10-18
fix(test): add mock method to mock credentials.

zhumin8:
This change is to avoid runtime error when creating the client bean https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3307#issuecomment-2417603861
Adding a mock return for `getMetricsCredentialType()` because in gax we do not allow for null values (should not happen because default is set)

Also added together in this pr so ci could pass:
- libraries bom upgrade (ci would still fail due to 1) removal of rpcs in this version 2) auth upgrade in this version that needs this fix)
- generation workflow (to fix for 1))

BEGIN_COMMIT_OVERRIDE
fix(deps): update dependency com.google.cloud:libraries-bom to v26.49.0

fix(test): add mock method to mock credentials. 
END_COMMIT_OVERRIDE


spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/config/PubSubBinderConfigurationTests.java
```diff
@@ -165,7 +167,10 @@ void test_healthRegistryEnabled() {
   private static class TestConfiguration {
     @Bean
     public CredentialsProvider googleCredentials() {
-      return () -> mock(Credentials.class);
+      Credentials mockCredential = mock(Credentials.class);
+      when(mockCredential.getMetricsCredentialType())
+          .thenReturn(CredentialTypeForMetrics.DO_NOT_SEND);
```
burkedavison, line 172: If mock credentials were placed in core, could you avoid the duplication here?

zhumin8: The mock credentials are placed under "test" folder, not accessible here.
`spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/TestUtils.java`
Created another util class in this module to avoid duplicating logic.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/TestUtils.java
```diff
@@ -0,0 +1,38 @@
...
+package com.google.cloud.spring.autoconfigure;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import com.google.auth.CredentialTypeForMetrics;
+import com.google.auth.Credentials;
+
+public class TestUtils {
+  public static final Credentials MOCK_CREDENTIALS = createMockCredentials();
```
JoeWang1127, line 26: private?

burkedavison (approved):

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/TestUtils.java
zhumin8: unfortunately, this is needed from classes in sub-packages within `com.google.cloud.spring.autoconfigure`.

PR #3314, opened 2024-10-18
chore: fix generate workflow to compile on non-autgen libraries

zhumin8:
This workflow step "Compile non-autogen libraries" but is trying to install all modules. This leads to failures https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3304#issuecomment-2423226206 because certain rpcs are removed (https://github.com/googleapis/google-cloud-java/pull/11237).

duplicating the profiles might not be the best way, but `-pl '!spring-cloud-previews' ` is not working for me. This is to  unblock release.


burkedavison (approved):

burkedavison:
Create #3323 to review configurations for improvement at a later time.

PR #3356, opened 2024-11-01
fix: need to update to use new trampoline_release.py script as well as update to configure docker auth with new image

ldetmer:


.kokoro/trampoline.sh
```diff
@@ -23,4 +23,11 @@ function cleanup() {
 trap cleanup EXIT
 
 $(dirname $0)/populate-secrets.sh # Secret Manager secrets.
-python3 "${KOKORO_GFILE_DIR}/trampoline_v1.py"
\ No newline at end of file
+TRAMPOLINE_HOST=$(echo "${TRAMPOLINE_IMAGE}" | cut -d/ -f1)
+if [[ ! "${TRAMPOLINE_HOST}" =~ "gcr.io" ]]; then
+  # If you need to specify a host other than gcr.io, you have to run on an update version of gcloud.
```
JoeWang1127, line 28: Can you write the rational of this setting? Do we treat images without `gcr.io` unsafe?

ldetmer: done

burkedavison (approved):

Issue #3331, opened 2024-10-24
Secret Manager: Support for Regional Secret Operations in Spring Cloud GCP Secret Manager


suvidha-malaviya:
### Description

We would like to request a feature enhancement for the Spring Cloud GCP library to support regional secret operations in the Secret Manager service.

Google Cloud Platform has recently introduced support for regional secrets, allowing users to create and manage secrets in specific regions for better performance and compliance. Currently, the Spring Cloud GCP library supports global secret operations, but it lacks functionality for the regional secret management capabilities.

Implementing support for regional secret operations would greatly enhance the flexibility and usability of the library, enabling users to take advantage of the regional features provided by GCP.

### Proposed Changes:

- Enhance the current Secret Manager service integration to support the creation, retrieval, and management of regional secrets. This can be achieved by introducing an additional property called `location` in `GcpSecretManagerProperties`, which other classes can use to provide the same functionality as the existing global secret operations.
- Update the documentation to reflect the new capabilities and provide sample of how to use regional secrets.

We believe it will be beneficial for many users looking to leverage regional secrets in their applications. 

More information about regional secrets: https://cloud.google.com/secret-manager/regional-secrets/data-residency

abheda-crest:
Opened the PR to add support for Regional Secret Operations in Spring Cloud GCP Secret Manager Service: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3365

PR #3365, opened 2024-11-04
feat: added regional secret support for secret-manager

abheda-crest:
Issue: [#3331](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3331)

**Description:**

Added the support for regional secret creation/updation, deletion and fetch for secret manager service.

- Updated the autoconfigure/secretmanager to create the regional client and secretmanager module to support the regional secret operation.
- Added the optional property location in GcpSecretManagerProperties.java which will take region from application.properties file. Whenever the location is available, it will use to perform the operation on regional secret. If not provided the global stack will be served.
- Updated the documentation for this additional property in the docs/src/main/asciidoc/secretmanager.adoc
- Added the sample application for regional secret operations.
- Added/Updated the unit and integration tests.

**Note:** Fixed the integration test `testUpdateSecrets` in the file `secretmanager/it/SecretManagerTemplateIntegrationTests.java`

Performed the below mentioned manual unit tests to validate the working of the global and regional secret operations.

- Create secret with only secretId and payload
- Create secret with existing secretId and payload
- Create secret with secretId, payload and valid projectID
- Create secret with secretId, payload and invalid projectID (project on which user doesn't have access)
- Read an existing secret with secretId
- Read a non-existing secret with secretId
- Read a secret with secretId and existing version
- Read a secret with secretId and disabled version
- Read a secret with secretId and destroyed version
- Read a secret with secretId and non-existing version
- Read a secret with secretId and valid project
- Read a secret with secretId and invalid project (project on which user doesn't have access)
- Read a secret with secretId and existing version and valid project
- Read a secret with secretId and existing version and invalid project
- Read a secret with secretId and non-existing version and valid project
- Read a secret with secretId and non-existing version and invalid project
- Update an existing secret with only secretId and payload
- Update an existing secret with secretId, payload and valid projectID
- Update an existing secret with secretId, payload and invalid projectID (project on which user doesn't have access)
- Delete an existing secret with secretId
- Delete a non-existing secret with secretId
- Delete a secret with secretId and valid projectId
- Delete a secret with secretId and invalid projectId
- Enable an existing secret and valid version
- Enable an existing secret without version
- Enable an existing secret and invalid version
- Enable a non-existing secret
- Disable an existing secret and valid version
- Disable an existing secret without version
- Disable an existing secret and invalid version
- Disable a non-existing secret
- Check if secret exists for existing secret
- Check if secret exists for non-existing secret
- Check if secret exists for existing secret & valid projectId
- Check if secret exists for existing secret & invalid projectId
- Read secret with secretId and other project using service account
- Inject secret with the `@Value` annotation

More information about regional secrets: https://cloud.google.com/secret-manager/regional-secrets/data-residency


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLoaderUnitTests.java
```diff
@@ -1,14 +1,19 @@
 package com.google.cloud.spring.autoconfigure.secretmanager;
 
 import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
mieseprem, line 4: We, in my company, always try to avoid having AssertJ AND JUnit assertions mixed in one class.
At least in my team we prefere to only use AssertJ as it let you chain assertions on same object.

Not sure if there are coding guidlines in Google projects, but maybe you consider to only use one assertions library to keep the mental load low.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerProperties.java
```diff
@@ -43,6 +44,13 @@ public class GcpSecretManagerProperties implements CredentialsSupplier {
    */
   private String projectId;
 
+  /**
+   * Defines the region of the secrets when Regional Stack is used.
+   *
+   * <p>When not specified, the secret manager will use the Global Stack.
```
mieseprem, line 50: Is it syntactically correct if there is no closing `</p>` tag?

```suggestion
   * <p>When not specified, the secret manager will use the Global Stack.</p>
```

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLoaderUnitTests.java
abheda-crest: I've replaced the JUnit Assertion with the AssertJ assertion.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLoaderUnitTests.java
```diff
@@ -21,18 +26,30 @@ class SecretManagerConfigDataLoaderUnitTests {
...
+    when(properties.getLocation()).thenReturn(Optional.ofNullable(location));
     SecretManagerConfigDataResource resource = new SecretManagerConfigDataResource(
-        ConfigDataLocation.of("fake"));
-    assertThatCode(() -> loader.load(loaderContext, resource)).doesNotThrowAnyException();
+        ConfigDataLocation.of(resourceName));
+    assertThatCode(() -> {
+      ConfigData configData = loader.load(loaderContext, resource);
+      SecretManagerPropertySource propertySource =
+          (SecretManagerPropertySource) configData.getPropertySources().get(0);
+      assertThat(Optional.ofNullable(location)).isEqualTo(propertySource.getLocation());
```
mieseprem, line 52: First, I'm not from Google, so my opinion is truly not more than my opinion. Google may see things different 😉 

Now, in assertions you normally test assumtions/expectations on an object. Having this on mind the assertions should be more in this way:
```text
assertThat(objectToTest)
  .verificationOne(...)
  .verificationTwo(...);
```

The result is the same, but now it is in a more logical order:

```java
assertThat(propertySource)
  .returns(Optional.ofNullable(location), SecretManagerPropertySource::getLocation);
```
You could add more verifications if suitable.

spring-cloud-gcp-secretmanager/src/test/java/com/google/cloud/spring/secretmanager/SecretManagerPropertyUtilsTests.java
```diff
@@ -102,4 +102,109 @@ void testLongProperty_projectSecretVersion() {
...
+        SecretManagerPropertyUtils.getSecretVersionName(
+            property,
+            DEFAULT_PROJECT_ID_PROVIDER,
+            "us-central1"
+        );
+
+    assertThat(secretIdentifier.getProject()).isEqualTo("defaultProject");
+    assertThat(secretIdentifier.getLocation()).isEqualTo("us-central1");
+    assertThat(secretIdentifier.getSecret()).isEqualTo("the-reg-secret");
+    assertThat(secretIdentifier.getSecretVersion()).isEqualTo("latest");
```
mieseprem, line 144: In reference to my other comment, this is how AssertJ chaining performs several checks on the same object:

```java
assertThat(secretIdentifier)
  .returns("defaultProject", SecretVersionName::getProject)
  .returns("us-central1", SecretVersionName::getLocation)
  .returns("the-reg-secret", SecretVersionName::getSecret)
  .returns("latest", SecretVersionName::getSecretVersion);
```

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLoaderUnitTests.java
abheda-crest: I followed the existing convention used throughout the codebase, where assertions are written in the same order as the current pattern. However, I understand your perspective, and if the team prefers the alternative style you suggested, I’m happy to make the adjustment.

spring-cloud-gcp-secretmanager/src/test/java/com/google/cloud/spring/secretmanager/SecretManagerPropertyUtilsTests.java
abheda-crest: As mentioned in the previous comment, I opted to maintain consistency with other assertions used across the project.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerProperties.java
```diff
@@ -43,6 +44,13 @@ public class GcpSecretManagerProperties implements CredentialsSupplier {
...
+   */
+  private Optional<String> location = Optional.empty();
```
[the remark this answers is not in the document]
abheda-crest: The use of `Optional` here is to explicitly signal that location may or may not be present, which I believe makes the intent clearer than a plain String. While it's true we're not using `.map()` or chaining operations now, `Optional` forces consumers to consider the absence of location explicitly (rather than relying on null checks). This can help avoid future bugs or assumptions about the field always being set.

I understand your suggestion regarding variable naming, but I believe the intent is already clear from the type itself (`Optional<String>`), and I prefer to keep the variable name focused on the domain rather than the technical implementation detail.

Let me know your thoughts, and I'm happy to make further adjustments if needed!

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerPropertySource.java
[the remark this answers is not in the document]
abheda-crest: The `getProperty()` method will still require special handling to account for different cases when retrieving the property value. Specifically, I have added an additional check for location, which is used to form the version name based on the location value. This logic will still be necessary, even if we keep the location as a String.

abheda-crest:
@mieseprem I've addressed the review comments. Could you please review these and let me know if those looks good?

mieseprem:
Hej,

I am pleased that my opinion is so valued. But the following also applies:

> First, I'm not from Google, so my opinion is truly not more than my opinion

So, I had already made my opinion known and I think now someone from Google would either have to give their approval or make some comments

abheda-crest:
@burkedavison Could you please review this PR?

docs/src/main/asciidoc/secretmanager.adoc
```diff
@@ -42,6 +42,7 @@ This can be overridden using the authentication properties.
 | `spring.cloud.gcp.secretmanager.credentials.location` | OAuth2 credentials for authenticating to the Google Cloud Secret Manager API. | No | By default, infers credentials from https://cloud.google.com/docs/authentication/production[Application Default Credentials].
 | `spring.cloud.gcp.secretmanager.credentials.encoded-key` | Base64-encoded contents of OAuth2 account private key for authenticating to the Google Cloud Secret Manager API. | No | By default, infers credentials from https://cloud.google.com/docs/authentication/production[Application Default Credentials].
 | `spring.cloud.gcp.secretmanager.project-id` | The default Google Cloud project used to access Secret Manager API for the template and property source. | No | By default, infers the project from https://cloud.google.com/docs/authentication/production[Application Default Credentials].
+| spring.cloud.gcp.secretmanager.location | Defines the region of the Secret Manager where your secrets are stored, specifically when using a regional stack. This option is particularly useful for applications that need to access secrets from a specific geographical location. | No |  By default, the global stack will be utilized.
```
jinseopkim0, line 45: nit: s/spring.cloud.gcp.secretmanager.location/`spring.cloud.gcp.secretmanager.location`/

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerPropertyUtils.java
```diff
@@ -48,6 +52,7 @@ static SecretVersionName getSecretVersionName(
     if (usedPrefix.isEmpty() || isAttemptingFullStringMatch) {
       return null;
     }
+
```
jinseopkim0, line 55: nit: remove this empty line

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerPropertyUtils.java
```diff
@@ -33,9 +33,13 @@ final class SecretManagerPropertyUtils {
...
   private SecretManagerPropertyUtils() {}
 
-
   static SecretVersionName getSecretVersionName(
       final String input, GcpProjectIdProvider projectIdProvider) {
+    return getSecretVersionName(input, projectIdProvider, null);
+  }
+
+  static SecretVersionName getSecretVersionName(
+      final String input, GcpProjectIdProvider projectIdProvider, String location) {
```
jinseopkim0, line 42: What do you think of using Optional<String> location as input here?

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerPropertyUtils.java
```diff
@@ -92,10 +97,15 @@ static SecretVersionName getSecretVersionName(
...
-        .setSecret(secretId)
-        .setSecretVersion(version)
-        .build();
+    return getSecretVersionName(projectId, secretId, version, location);
+  }
+
+  static SecretVersionName getSecretVersionName(
+          String projectId, String secretId, String version, String location) {
+    if (location != null) {
+      return SecretVersionName.newProjectLocationSecretSecretVersionBuilder().setLocation(location).setProject(projectId).setSecret(secretId).setSecretVersion(version).build();
```
jinseopkim0, line 106: To confirm, location needs to be String here, right? What do you think of having location to be `Optional<String>` everywhere else until we reach this line?

jinseopkim0 (changes requested):
Thank you for the PR. Left a few comments. PTAL.

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerPropertyUtils.java
abheda-crest: Yes, `location` needs to be a `String` here, but I have modified the code to use `Optional<String>` even after this point for consistency. Let me know if you have any further suggestions!

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerTemplate.java
```diff
@@ -233,27 +236,34 @@ private void createNewSecretVersion(String secretId, ByteString payload, String
...
-    ProjectName projectName = ProjectName.of(projectId);
-
-    Secret secret =
-        Secret.newBuilder()
-            .setReplication(
-                Replication.newBuilder().setAutomatic(Replication.Automatic.getDefaultInstance()))
-            .build();
+    String parent;
+    Secret.Builder secret = Secret.newBuilder();
+    if (location.isPresent()) {
```
jinseopkim0, line 241: Is it possible to use ifPresentOrElse here? Also, would it be better to use getLocation() instead of location?

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerTemplate.java
```diff
@@ -233,27 +236,34 @@ private void createNewSecretVersion(String secretId, ByteString payload, String
...
-        .setProject(this.projectIdProvider.getProjectId())
-        .setSecret(secretId)
-        .setSecretVersion(LATEST_VERSION)
-        .build();
+    return SecretManagerPropertyUtils.getSecretVersionName(
+        this.projectIdProvider.getProjectId(), secretId, LATEST_VERSION, getLocation());
+  }
+
+  private SecretName getSecretName(String projectId, String secretId) {
+    if (location.isPresent()) {
```
jinseopkim0, line 263: Is it possible to use ifPresentOrElse here? Also, would it be better to use getLocation() instead of location?

abheda-crest: Yes updated

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerTemplate.java
abheda-crest: Updated

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerPropertyUtils.java
```diff
@@ -92,10 +96,15 @@ static SecretVersionName getSecretVersionName(
...
-        .setProject(projectId)
-        .setSecret(secretId)
-        .setSecretVersion(version)
-        .build();
+    return getSecretVersionName(projectId, secretId, version, location);
+  }
+
+  static SecretVersionName getSecretVersionName(
+          String projectId, String secretId, String version, Optional<String> location) {
+    if (location.isPresent()) {
```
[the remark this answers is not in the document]
abheda-crest: Updated

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerPropertySource.java
```diff
@@ -38,13 +41,41 @@ public SecretManagerPropertySource(
...
+  public Optional<String> getLocation() {
+    return location;
+  }
+
   @Override
   public Object getProperty(String name) {
-    SecretVersionName secretIdentifier =
-        SecretManagerPropertyUtils.getSecretVersionName(name, this.projectIdProvider);
+    AtomicReference<SecretVersionName> secretIdentifier = new AtomicReference<>();
+    getLocation().ifPresentOrElse(
```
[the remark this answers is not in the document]
abheda-crest: Yes I have removed it

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerPropertyUtils.java
```diff
@@ -33,9 +33,13 @@ final class SecretManagerPropertyUtils {
 
   private SecretManagerPropertyUtils() {}
 
-
   static SecretVersionName getSecretVersionName(
```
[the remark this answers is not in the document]
abheda-crest: Removed it

meltsufin (changes requested):
1) I don't think we should use `Optional` for location.
2) How do you retrieve secrets from two different projects with different locations? I think we should add support for secret URIs like `sm://projects/<project-id>/locations/<location>/secrets/<secret-id>/versions/<version-id>}`. (See [Section 21.2](https://googlecloudplatform.github.io/spring-cloud-gcp/6.0.0/reference/html/index.html#secret-manager) in the documentation.)

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLoaderUnitTests.java
zhumin8: We do not have a strong preference on the assertion style. 
However, in your current code, for better readability please revert actual and expected. [`isEqualTo()`](https://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractAssert.html#isEqualTo(java.lang.Object)) takes the expected value.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerRegionalCompatibilityTests.java
```diff
@@ -0,0 +1,107 @@
+package com.google.cloud.spring.autoconfigure.secretmanager;
```
zhumin8, line 1: Missing header

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-regional-sample/src/main/java/com/example/SecretManagerRegionalApplication.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2017-2024 the original author or authors.
```
zhumin8, line 2: For new classes, can you switch to headers like [this one](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/67d994e7f39372be7d0e9010e234f2fe46e68fd9/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/alloydb/AlloyDbProperties.java#L1-L15), with 2025 for the year? We have this header discrepancy due to historical migration reasons. Sorry for the confusion. 

```suggestion
 * Copyright 2025 Google LLC
```

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerProperties.java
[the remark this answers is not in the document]
zhumin8: Additionally, from source quoted above, behavior for Optional properties is 
"if you do declare an `Optional` property and it has no value, `null` rather than an empty `Optional` will be bound."
ref: https://github.com/spring-projects/spring-boot/commit/7b3c0a9e096501daf55271e1beb73550ced915c2

zhumin8 (changes requested):

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-regional-sample/src/main/java/com/example/SecretManagerRegionalApplication.java
abheda-crest: No worries. Updated

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerRegionalCompatibilityTests.java
abheda-crest: Added

suvidha-malaviya:
> 1. I don't think we should use `Optional` for location.
> 2. How do you retrieve secrets from two different projects with different locations? I think we should add support for secret URIs like `sm://projects/<project-id>/locations/<location>/secrets/<secret-id>/versions/<version-id>}`. (See [Section 21.2](https://googlecloudplatform.github.io/spring-cloud-gcp/6.0.0/reference/html/index.html#secret-manager) in the documentation.)

In current enhancement, we have implemented the regional support in such a way that, it will support all of the syntax to specify the secret mentioned in the documentation ([Section 21.2](https://googlecloudplatform.github.io/spring-cloud-gcp/6.0.0/reference/html/index.html#secret-manager-config-data-resource)). By default it will be rendering the secrets from the global stack. If we specify the spring.cloud.gcp.secretmanager.location field in the application.properties file, it will render the regional secret for mentioned location. At a time single region can be accessible.

I have taken the reference from the KMS for multi-region support as mentioned in the review comment. The reason we cannot implement it is :
- In case of KMS, the default client creation can access the regional keys as it accept the name in form of projects/*/locations/*/keyRings/* (See [KMS Rest API documentation)](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings) which can able to access the regional and global with default client creation.
- While in case of Secret Manager, it accept the name in form of projects/*/secrets/* (See [Secret Manager Rest API documentation](https://cloud.google.com/secret-manager/docs/reference/rest/v1/projects.secrets)) which can not access the secrets from across the region/globally.

abheda-crest:
> 1. I don't think we should use `Optional` for location.
> 2. How do you retrieve secrets from two different projects with different locations? I think we should add support for secret URIs like `sm://projects/<project-id>/locations/<location>/secrets/<secret-id>/versions/<version-id>}`. (See [Section 21.2](https://googlecloudplatform.github.io/spring-cloud-gcp/6.0.0/reference/html/index.html#secret-manager) in the documentation.)

In current enhancement, we have implemented the regional support in such a way that, it will support all of the syntax to specify the secret mentioned in the documentation ([Section 21.2](https://googlecloudplatform.github.io/spring-cloud-gcp/6.0.0/reference/html/index.html#secret-manager-config-data-resource)). By default it will be rendering the secrets from the global stack. If we specify the `spring.cloud.gcp.secretmanager.location` field in the application.properties file, it will render the regional secret for mentioned location. At a time single region can be accessible.

I have taken the reference from the KMS for multi-region support as mentioned in the review comment. The reason we cannot implement it is :
- In case of KMS, the default client creation can access the regional keys as it accept the name in form of `projects/*/locations/*/keyRings/*` (See [KMS Rest API documentation)](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings) which can able to access the regional and global with default client creation.
- While in case of Secret Manager, it accept the name in form of `projects/*/secrets/*` (See [Secret Manager Rest API documentation](https://cloud.google.com/secret-manager/docs/reference/rest/v1/projects.secrets)) which can not access the secrets from across the region/globally.

meltsufin:
> > What do you think about this point about the use of `Optional`?
> 
> I have referred [documentation](https://github.com/spring-projects/spring-boot/commit/7b3c0a9e096501daf55271e1beb73550ced915c2) and as it is not recommended to use `java.util.Optional` with `@ConfigurationProperties` I have change the type to `String`
> 

I think there are still some unused `import java.util.Optional;` statements in the file though.

> > I see that the API doesn't support this reference format natively, but we parse it ourselves in SecretManagerPropertyUtils. So, can't you implement parsing of location from the secret URI there? This will allow accessing secrets from multiple regions in the same application.
> 
> We can implement the parsing logic for secret URI, but to access the regional secrets we will need the client created for that region.
> 
> * For example if we create the default client it will allow us to access the global secrets.
> * For regional secret access we need to pass the regional endpoint `secretmanager.{region}.rep.googleapis.com:443` while client creation. Also the secret manager client bean is created on application startup based on the configuration mentioned in the `application.properties` file. The client created for one region cannot access the secret from another region. We parse the secret URI and pass it to access the secret it will give error while trying to access the secret of another region. Please refer to the error screenshot I have attached for more clarity.
> 
> Screenshot 2025-02-28 at 12 39 30 PM

If the client is bound to a particular region, why do you still need to specify the same region in the secret requests? It seems redundant.

abheda-crest:
> I think there are still some unused `import java.util.Optional;` statements in the file though.

I have removed it, It was unused import from the tests file.

> If the client is bound to a particular region, why do you still need to specify the same region in the secret requests? It seems redundant.

The endpoint to get the regional secret is as follow: `https://secretmanager.{location}.rep.googleapis.com/v1/projects/{project}/locations/{location}/secrets/{secret}`. The Google SDK itself is implemented in such a way that we need to pass the endpoint while client creation and pass location when using regional secrets.

- It can be observed that to access the regional secret the different endpoint will be used, which is part of the client creation for the regional secret. 
- Other than this, the regional secret have the format `projects/{project-id}/locations/{location-id}/secrets/{secret}` which needs to be passed to the Google SDK methods of the secret. If the location is not passed while performing any operation on the regional secret, it will throw error which is mentioned in the below screenshot.

Screenshot 2025-03-03 at 12 36 08 PM

abheda-crest:
> > I think there are still some unused `import java.util.Optional;` statements in the file though.
> 
> I have removed it, It was unused import from the tests file.
> 
> > If the client is bound to a particular region, why do you still need to specify the same region in the secret requests? It seems redundant.
> 
> The endpoint to get the regional secret is as follow: `https://secretmanager.{location}.rep.googleapis.com/v1/projects/{project}/locations/{location}/secrets/{secret}`. The Google SDK itself is implemented in such a way that we need to pass the endpoint while client creation and pass location when using regional secrets.
> 
> * It can be observed that to access the regional secret the different endpoint will be used, which is part of the client creation for the regional secret.
> * Other than this, the regional secret have the format `projects/{project-id}/locations/{location-id}/secrets/{secret}` which needs to be passed to the Google SDK methods of the secret. If the location is not passed while performing any operation on the regional secret, it will throw error which is mentioned in the below screenshot.
> 
> Screenshot 2025-03-03 at 12 36 08 PM
> [Editing the reply, after checking the code again] I have check out the code again and it seems that we can support the string format `sm://projects/<project-id>/locations/<location>/secrets/<secret-id>/versions/<version-id>` by using the `SecretManagerTemplate`. We can override each methods which takes `locationId` field.
> 
> I still need to validate the use case of the secret resolution in `application.properties` file and using Value injection for multi-location or the combination of global/regional secrets.
> 
> I'll update here once I completed with the development.

I have updated the code to support the regional secret operation. Introduced the below forms for regional secrets. I have used `ConcurrentHashMap` for maintain the `SecretManagerServiceClient`. Please re-review the code and let me know your thoughts.

1. Long form - specify project ID, location ID, secret ID, and version
`sm@projects/<project-id>/locations/<location-id>/secrets/<secret-id>/versions/<version-id>`

2. Long form - specify project ID, location ID, secret ID, and use latest version
`sm@projects/<project-id>/locations/<location-id>/secrets/<secret-id>`

3. Short form - specify project ID, location ID, secret ID, and version
`sm@<project-id>/<location-id>/<secret-id>/<version-id>`

4. Short form - specify location ID, secret ID,
version and use default project
`sm@locations/<location-id>/<secret-id>/<version>`

5. Shortest form - specify location ID, secret ID, and use default project and latest version
`sm@locations/<location-id>/<secret-id>`

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/DefaultSecretManagerServiceClientFactory.java
```diff
@@ -0,0 +1,83 @@
...
+      } catch (IOException e) {
+        throw new RuntimeException(
+            "Failed to create SecretManagerServiceClient for location: " + loc, e);
+      }
+    });
+  }
+
+  @Override
+  public SecretManagerServiceClient getClient() {
+    return clientCache.computeIfAbsent("", loc -> {
```
meltsufin, line 70: Can you just re-use `getClient(String)` with some modification to avoid code duplication?

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerTemplate.java
```diff
@@ -47,17 +48,18 @@ public class SecretManagerTemplate implements SecretManagerOperations {
...
+  private final SecretManagerServiceClientFactory clientFactory;
   private final GcpProjectIdProvider projectIdProvider;
   /**
    * Define the behavior when accessing a non-existed secret string/bytes.
    */
   private boolean allowDefaultSecretValue;
 
+
   public SecretManagerTemplate(
-      SecretManagerServiceClient secretManagerServiceClient,
```
meltsufin, line 58: This is a breaking change. Consider just adding an additional constructor.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java
```diff
@@ -60,22 +58,14 @@ public GcpSecretManagerAutoConfiguration(
 
   @Bean
   @ConditionalOnMissingBean
-  public SecretManagerServiceClient secretManagerClient()
```
meltsufin, line 63: Customers may already be using this one or defining a custom `secretManagerClient` bean and expecting it to be used. Consider leaving this bean and passing it down into `clientFactory`. I understand that it might not be used for apps that only use regional secrets, but it would be more backwards-compatible.

meltsufin (changes requested):

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/DefaultSecretManagerServiceClientFactory.java
abheda-crest: Updated

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java
abheda-crest: Updated. Also checked sample application by using the `SecretManagerServiceClient` with the `SecretManagerTemplate`.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java
```diff
@@ -60,22 +62,33 @@ public GcpSecretManagerAutoConfiguration(
...
+  public SecretManagerServiceClientFactory clientFactory(SecretManagerServiceClient client) {
+    return new DefaultSecretManagerServiceClientFactory(this.credentialsProvider, client);
+  }
+
+
+  @Bean
+  @ConditionalOnMissingBean
+  public SecretManagerTemplate secretManagerTemplate(
+      SecretManagerServiceClient client, SecretManagerServiceClientFactory clientFactory) {
+    if (clientFactory != null) {
```
meltsufin, line 85: Are you sure Spring Boot will not throw an an exception if `clientFactory` bean does not exist? I believe you have to use `ObjectProvider` in the method signature.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolverUnitTests.java
```diff
@@ -70,22 +62,45 @@ void resolveReturnsConfigDataLocation(String prefix) {
...
+
+    when(properties.getCredentials()).thenReturn(credentials);
 
     defaultBootstrapContext.register(GcpSecretManagerProperties.class,
         BootstrapRegistry.InstanceSupplier.of(properties));
     defaultBootstrapContext.register(CredentialsProvider.class,
         BootstrapRegistry.InstanceSupplier.of(credentialsProvider));
     defaultBootstrapContext.register(SecretManagerServiceClient.class,
         BootstrapRegistry.InstanceSupplier.of(secretManagerServiceClient));
+    defaultBootstrapContext.register(SecretManagerServiceClientFactory.class,
```
meltsufin, line 102: Can we also have a test without this bean registered?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java
abheda-crest: Yes, I have changed it with `ObjectProvider`

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolverUnitTests.java
abheda-crest: Added the test cases

meltsufin (approved):
@jinseopkim0  @zhumin8  Please do a second pass.

abheda-crest:
@zhumin8 I have resolved the merge conflict. Can you please review the PR?

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerTemplate.java
```diff
@@ -46,22 +48,51 @@ public class SecretManagerTemplate implements SecretManagerOperations {
    */
   public static final String LATEST_VERSION = "latest";
 
+  /**
+   * Default value for the latest version of the secret.
+   */
+  public static final String GLOBAL_LOCATION = "global";
```
zhumin8, line 54: GLOBAL_LOCATION seems only used within this class, does it needs to be public? We try to restrict public access when possible  to avoid future breaking changes.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/DefaultSecretManagerServiceClientFactory.java
```diff
@@ -0,0 +1,75 @@
...
+ * UserAgentHeaderProvider} that adds the Spring Cloud GCP agent header to the client.
+ *
+ */
+@Component
+public class DefaultSecretManagerServiceClientFactory implements SecretManagerServiceClientFactory {
+
+  /**
+   * Default value for the latest version of the secret.
+   */
+  public static final String GLOBAL_LOCATION = "global";
```
zhumin8, line 45: GLOBAL_LOCATION seems only used within this class, does it needs to be public? We try to restrict public access when possible to avoid future breaking changes.

Also see this defined separately in SecretManagerTemplate below

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/DefaultSecretManagerServiceClientFactory.java
```diff
@@ -0,0 +1,75 @@
...
+
+  private final CredentialsProvider credentialsProvider;
+  private final Map<String, SecretManagerServiceClient> clientCache = new ConcurrentHashMap<>();
+
+  public DefaultSecretManagerServiceClientFactory(CredentialsProvider credentialsProvider, SecretManagerServiceClient client) {
```
zhumin8, line 50: Does this method need to be accessed from outside of this package? Let's limit the scope if possible

docs/src/main/asciidoc/secretmanager.adoc
```diff
@@ -76,6 +76,22 @@ sm@<secret-id>/<version>
 
 # 5. Shortest form - specify secret ID, use default project and latest version.
 sm@<secret-id>
+
```
zhumin8, line 79: Probably worth highlighting 1-5 is for global secrets, and 6-10 for regional secrets. maybe even with a link to https://cloud.google.com/secret-manager/regional-secrets/data-residency

zhumin8 (changes requested):
Looks good in general, commented on some smaller points. Will do another pass.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfigurationUnitTests.java
```diff
@@ -61,9 +60,9 @@ void testProjectIdWithGcpProperties() {
   }
 
   @Test
-  void testSecretManagerServiceClientExists() {
```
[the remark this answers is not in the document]
abheda-crest: Added

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLoaderUnitTests.java
```diff
@@ -1,14 +1,19 @@
 package com.google.cloud.spring.autoconfigure.secretmanager;
 
+import static org.assertj.core.api.Assertions.assertThat;
```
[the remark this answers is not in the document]
abheda-crest: Removed unused import

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
```diff
@@ -62,6 +63,10 @@ public class SecretManagerWebController {
   @Value("${sm@application-secret:DEFAULT}")
   private String appSecretFromValue;
 
+  // Application secrets can be accessed using @Value syntax.
```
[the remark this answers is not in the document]
abheda-crest: Updated

spring-cloud-gcp-secretmanager/src/main/java/com/google/cloud/spring/secretmanager/SecretManagerTemplate.java
abheda-crest: Used it in `DefaultSecretManagerServiceClientFactory.java` and removed the duplicate constant in that file.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/DefaultSecretManagerServiceClientFactory.java
abheda-crest: Used constant from the `SecretManagerTemplate`

docs/src/main/asciidoc/secretmanager.adoc
abheda-crest: Updated

jinseopkim0 (approved):

zhumin8 (approved):

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
```diff
@@ -104,8 +110,15 @@ private static void registerSecretManagerBeans(ConfigDataLocationResolverContext
     // Register the Core properties.
     registerBean(context, GcpProperties.class, getGcpProperties(context));
     // Register the Secret Manager properties.
-    registerBean(
-        context, GcpSecretManagerProperties.class, getSecretManagerProperties(context));
+    registerBean(context, GcpSecretManagerProperties.class, getSecretManagerProperties(context));
+    // Register the CredentialsProvider.
+    registerBean(context, CredentialsProvider.class, getCredentialsProvider(context));
+    // Register the Secret Manager client factory.
```
zhumin8, line 116: @abheda-crest Sorry this was missed in earlier reviews.
Merging this PR caused IT tests failures across modules. I have reverted this change. More context [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3734)

 I believe the root cause is here: This SecretManagerConfigDataLocationResolver is used regardless if secretmanager dependency is present. When not present, for example if you run the vision sample app, this code `createSecretManagerServiceClientFactory()` calls DefaultSecretManagerServiceClientFactory that eventually gives "NoClassDefFoundError" for "com/google/cloud/spring/secretmanager/SecretManagerServiceClientFactory"

Note that secretmanager is optional dependency for autoconfig. You will need to guard this logic to only run when secretmanager is present.

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
```diff
@@ -62,6 +63,10 @@ public class SecretManagerWebController {
   @Value("${sm@application-secret:DEFAULT}")
   private String appSecretFromValue;
 
+  // Application secrets with regions can be accessed with this syntax.
+  @Value("${sm@locations/us-central1/application-secret:DEFAULT}")
```
zhumin8, line 67: Can you add step to [README](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/README.adoc) on setting up this regional secret (after step 4)?

Also, I am not familiar with this, but can global and regional secret co-exist in same project?  From [this page](https://cloud.google.com/secret-manager/regional-secrets/global-regionalized-service-comparison), it sounds like a choice per org?

zhumin8 (commented):
@abheda-crest Sorry for the inconvenience, but do you mind opening a new pr for this change once done with issue below? I had to revert this after merging in.

cc. @jinseopkim0 @meltsufin

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
abheda-crest: I have fixed the issue

abheda-crest:
I have raised another PR after fixing the issue: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3746

Issue #3181, opened 2024-08-30
SecretManager autoconfiguration ignores and superseeds spring.cloud.gcp.project-id


mieseprem:
This issue is regarding the projectId value returned by GcpProjectIdProvider when SecretManager is enabled.

I know, as an example, theses two closed issues, but I think they doesn't cover the root cause:
* https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/2698
* https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1793

Regarding [the documentation](https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#project-id), the `GcpProjectIdProvider.getProjectId()` method will always return the value of the property `spring.cloud.gcp.project-id` if set. There is no limitation regarding the use of SecretManager.

In addition, in [the SecretManager section of the documentation](https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#configuration-11) the project-id is (if not set as property): "By default, infers the project from [Application Default Credentials](https://cloud.google.com/docs/authentication/production)"

I wasn't able to to determine the method who creates the `GcpProjectIdProvider` bean for the [SecretManagerAutoConfiguration](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java#L51), but it definitivley get its values not from the `spring.cloud.gcp.project-id` property.
Because of this presence of a `GcpProjectIdProvider` bean there is no propper one created in [GcpContextAutoConfiguration](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/core/GcpContextAutoConfiguration.java#L65-L78).

I created a simple repository to demonstrate the issue: https://github.com/mieseprem/projectIdIssue
Please keep in mind that this issue is not on how to access secrets or other services in GCP. This is only about how the project-id value is determined and provided to all beans that relay on the `GcpProjectIdProvider`

In my example project, there is a profile named "xxx" and the project-id is set to a specific value. If you run the application with active profile "xxx" you will see 2 things:
* the `GcpProjectIdProvider` is created via 'GcpContextAutoConfiguration'
[...]

PR #3383, opened 2024-11-16
fix: GcpProjectIdProvider when using Secret Manager

PatrickGotthard:
When using the Secret Manager module the GcpProjectIdProvider behaves incorrectly.
It will return the property **spring.cloud.gcp.secretmanager.project-id** or a DefaultGcpProjectIdProvider instead of **spring.cloud.gcp.project-id**. Also the SecretManagerConfigDataLocationResolver currently does not take the property **spring.cloud.gcp.project-id** into account.

This PR solves #3181


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
```diff
@@ -94,10 +104,21 @@ private static GcpSecretManagerProperties getSecretManagerProperties(
...
+    
+    GcpSecretManagerProperties secretManagerProperties = bootstrapContext.get(GcpSecretManagerProperties.class);
+    if(secretManagerProperties.getProjectId() != null) {
+        return secretManagerProperties::getProjectId;
+    }
+    
+    GcpProperties gcpProperties = bootstrapContext.get(GcpProperties.class);
+    if(gcpProperties.getProjectId() != null) {
+        return gcpProperties::getProjectId;
+    }
```
mieseprem, line 118: As written in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3181#issuecomment-2488507623 this needs a little change:
```suggestion
    if (secretManagerProperties.getProjectId() != null) {
      return secretManagerProperties::getProjectId;
    }
    
    GcpProperties gcpProperties = bootstrapContext.get(GcpProperties.class);
    if (gcpProperties.getProjectId() != null) {
      return gcpProperties::getProjectId;
    }
```

mieseprem (approved):

zhumin8 (commented):
Thanks for contributing!
This looks generally reasonable to me. However, I am not too familiar with the bootstrap configs. 
Also, we will need some test coverage for this change.

I will take another look. In the meantime, adding @JoeWang1127 who is more familiar with this code for a second opinion.

zhumin8:
closing in favor of #3664 created from branch to allow full ci tests.

Issue #3402, opened 2024-12-06
Provide better error information in PubSubInboundChannelAdapter


burkedavison:
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/inbound/PubSubInboundChannelAdapter.java

The 'logWarning' method in this class logs the `RuntimeException`'s top-level exception message, but not the nested exception messages. This results in the logger containing only mildly-helpful information akin to "error occurred in message handler". (See b/374291633#comment17 for a specific example.) This means important debug information is lost about the root cause.

This issue is not prescriptive about what changes should be made to the logged error information, but it should be improved to ensure customers are able to understand the root cause of the exception.

PR #3418, opened 2024-12-13
feat: log stacktrace

JoeWang1127:
Fix: #3402

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -179,6 +179,7 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
     } else {
       LOGGER.warn(re.getMessage());
     }
+    LOGGER.warn(re.getStackTrace());
   }
 
   private void addToHealthRegistry() {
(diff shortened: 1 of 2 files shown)


JoeWang1127:
Native tests failed:
```
Error: Classes that should be initialized at run time got initialized during image building:
 com.google.protobuf.RuntimeVersion was unintentionally initialized at build time. To see why com.google.protobuf.RuntimeVersion got initialized use --trace-class-initialization=com.google.protobuf.RuntimeVersion
To see how the classes got initialized, use --trace-class-initialization=com.google.protobuf.RuntimeVersion
```

This is irrelevant to this feature.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -179,6 +181,10 @@ private void logWarning(
     } else {
       LOGGER.warn(re.getMessage());
     }
+    // Log the stacktrace for troubleshoot.
+    StringWriter stringWriter = new StringWriter();
+    re.printStackTrace(new PrintWriter(stringWriter));
```
burkedavison, line 186: This is not threadsafe -- we probably don't want this here. Logging is the goal.

spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -179,6 +181,10 @@ private void logWarning(
...
+    LOGGER.warn(stringWriter.toString());
```
burkedavison, line 187: This would duplicate the log entry since we're already sending a log above. Can we adjust the log messages above, instead? For example, `LOGGER.warn("Message", exception)`

JoeWang1127: Ah, thanks, I wasn't aware there's a `LOGGER.warn(String, Exception)` method.

burkedavison (approved):

PR #3425, opened 2024-12-18
chore: update libraries-bom in 3.x

JoeWang1127:


diegomarquezp:
The test started failing since https://github.com/googleapis/java-pubsub/pull/2271

The test confirms that `private` fields have a certain value, but these fields were changed to `java.time.Duration`.

This does not imply any breakage in the public interface of java-pubsub

burkedavison (approved):

Issue #3494, opened 2025-01-22
Nightly IT failures are consistent


diegomarquezp:
Example: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/12903857310/job/35979910068

Pubsub fails consistently.
Other modules may fail as well.

zhumin8:
To add context: I briefly looked last week, pubsub-bus has been consistently failing. 
Others are more of flaky.

diegomarquezp:
Last failure was 100% green

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/13280797293

Closing for now as this was mean to track _sequentially consistent_ failures. Flakes should have their own bug.

PR #3501, opened 2025-01-24
feat(pubsub)!: set default max ack extension period to 60 minutes

diegomarquezp:
* Max ack extension period: we do not use a `final` variable storing the default value. We will instead build the library with a null value if not provided via beans, which will default to the core library's default value


diegomarquezp:
Pubsub-bus failure is tracked in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3494

meltsufin (changes requested):
Instead of setting the default, can we just not set the property and automatically expose the default already configured in Pub/Sub client library?

diegomarquezp:
> Instead of setting the default, can we just not set the property and automatically expose the default already configured in Pub/Sub client library?

@meltsufin See https://github.com/googleapis/java-pubsub/blob/855653f1ae37c4c45600d2ef0c34e1804939e75a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java#L96-L105
The clients don't have these defaults publicly exposed. Should we propose this change in pubsub?

meltsufin:
Why do you need access to the default property in Pub/Sub? You shouldn't need to use it directly. If you don't set it in `DefaultSubscriberFactory.createSubscriber`, the default will be used. The way to make sure it's not set, is to keep our value `null`. `null` -> default.

diegomarquezp:
@meltsufin PTAL. We are now passing null values to build the subscriber. However, I found that the global executor bean became unnecessary since it defaults to the static values we are getting rid of. Not sure if there is a use case where a custom global executor with other values satisfies a special need?

diegomarquezp:
@meltsufin PTAL. We are now passing null values to build the subscriber. However, I found that the global executor bean became unnecessary since it defaults to the static values we are getting rid of. Not sure if there is a use case where a custom global executor with other values satisfies a special need?

meltsufin (changes requested):
There should be documentation updates corresponding to this too.

It looks like you included changes for `DEFAULT_EXECUTOR_THREADS` as well. Would it make sense to do it in a separate PR? Was it part of the original request?

diegomarquezp:
> There should be documentation updates corresponding to this too.
> 
> It looks like you included changes for `DEFAULT_EXECUTOR_THREADS` as well. Would it make sense to do it in a separate PR? Was it part of the original request?

@meltsufin I double checked. There seems to not be a hard requirement on default threads (i.e. "the really important is max ack..."). I restricted this PR to only include the default max ack extension period updates and changed the docs accordingly.

meltsufin (approved):

PR #3525, opened 2025-02-04
chore: make parallel checks required in 5.x

diegomarquezp:


diegomarquezp:
Looks like this file doesn't have effects. I modified the required checks directly in Settings. QQ: What does this file affect then? @suztomo maybe?

PR #3411, opened 2024-12-09
feat!: Spring Cloud 2024.0 and Spring Boot 3.4.1

diegomarquezp:
This pull request introduces an upgrade to Spring Cloud 2024 and Spring Boot 3.4

### Changes per module

 * **spring-cloud-gcp-secretmanager**: We will introduce support for resolving values of the form `sm@secret_id` in addition to the already existing `sm://secret_id` format. Users should move away from `sm://` and use `sm@` instead. Usage of `sm://` will raise a warning about this being removed in the future. This has been tested using `spring-framework:6.2.2-SNAPSHOT`.
 * **spring-cloud-gcp-config-sample (breaking change)**: removed autoconfiguration and sample code. This is to adjust to what's stated in [the docs](https://googlecloudplatform.github.io/spring-cloud-gcp/5.9.0/reference/html/index.html#cloud-runtime-configuration-api).
 * **spring-cloud-gcp-kotlin-sample (breaking change)**: Adjust to [hibernate upgrade](https://github.com/spring-projects/spring-boot/issues/43100) which [introduced](https://github.com/hibernate/hibernate-orm/commit/522269e9a93f71af593f1d94039df5ade4f93617) a change in default value handling (see [tracker](https://hibernate.atlassian.net/browse/HHH-1661)). (More info in [SO](https://stackoverflow.com/questions/79228209/spring-boot-3-4-0-lets-integration-tests-with-jpa-hibernate-fail))

### Internal changes
 * **spring-cloud-gcp-pubsub-stream**: Adjust unit tests to prevent duplicate bean detection ([example logs](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/actions/runs/12542129567/job/34971518975)). 
 * **spring-cloud-gcp-data-datastore**: Upgrade [deprecated](https://github.com/spring-projects/spring-data-commons/commit/096836ee808d2cf82caf2efd48cb56271e6ee7c8) `QueryMethodEvaluationContextProvider` to `ValueExpressionDelegate`
 * **spring-cloud-gcp-data-spanner**: Upgrade [deprecated](https://github.com/spring-projects/spring-data-commons/commit/096836ee808d2cf82caf2efd48cb56271e6ee7c8) `QueryMethodEvaluationContextProvider` to `ValueExpressionDelegate`

### Pending changes
[...]

PR #3534, opened 2025-02-06
docs: update secret manager documentation

diegomarquezp:
This PR explains the introduction of the `sm@` syntax introduced by https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3411

From its description:
We will introduce support for resolving values of the form `sm@secret_id` in addition to the already existing `sm://secret_id` format. Users should move away from `sm://` and use `sm@` instead. Usage of `sm://` will raise a warning about this being removed in the future. This has been tested using `spring-framework:6.2.2-SNAPSHOT`.

The introduction of this syntax was suggested in a discussion of a regression caused by spring-framework:6.2.1 - see https://github.com/spring-projects/spring-framework/issues/34124


zhumin8 (commented):
LGTM. Do you mind adding some context to the pr description? So if we come back to it and forget, we still have a reminder why this recommendation was made. If you got any links from Spring for recommending this new syntax will be great.

diegomarquezp:
> LGTM. Do you mind adding some context to the pr description? So if we come back to it and forget, we still have a reminder why this recommendation was made. If you got any links from Spring for recommending this new syntax will be great.

@zhumin8 thanks for the review!
I have modified the description explaining how the new support for Spring Cloud 2024 has most of the context, plus the discussion with the spring team where they suggest this syntax.

zhumin8 (approved):

PR #3627, opened 2025-03-04
fix(test): change logback default flush lev

JoeWang1127:
In this PR:
- Change logback default flush level, see https://github.com/googleapis/java-logging-logback/pull/1441 for more information.

### DIFF
--- a/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
+++ b/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
@@ -44,8 +44,10 @@ class LoggingAppenderTests {
     LoggingAppender loggingAppender = new LoggingAppender();
     loggingAppender.setCredentialsFile("src/test/resources/fake-project-key.json");
     loggingAppender.setLogDestinationProjectId("my-log-destination-project");
-    assertThat(loggingAppender.getLoggingOptions().getCredentials()).isNotNull();
-    assertThat(loggingAppender.getLoggingOptions().getProjectId()).isEqualTo("my-log-destination-project");
+    assertThat(loggingAppender.getLoggingOptions().getCredentials())
+        .isNotNull();
+    assertThat(loggingAppender.getLoggingOptions().getProjectId())
+        .isEqualTo("my-log-destination-project");
     assertThat(loggingAppender.getLoggingOptions().getUserAgent())
         .isNotNull()
         .contains("spring-cloud-gcp-logging")
--- a/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
+++ b/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
@@ -44,7 +44,11 @@ class LoggingAppenderTests {
     LoggingAppender loggingAppender = new LoggingAppender();
     loggingAppender.setCredentialsFile("src/test/resources/fake-project-key.json");
     loggingAppender.setLogDestinationProjectId("my-log-destination-project");
-    assertThat(loggingAppender.getLoggingOptions().getProjectId())
-        .isEqualTo("my-log-destination-project");
+    assertThat(loggingAppender.getLoggingOptions().getCredentials()).isNotNull();
+    assertThat(loggingAppender.getLoggingOptions().getProjectId()).isEqualTo("my-log-destination-project");
+    assertThat(loggingAppender.getLoggingOptions().getUserAgent())
+        .isNotNull()
+        .contains("spring-cloud-gcp-logging")
+        .contains("Spring");;
   }
 }
--- a/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
+++ b/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
@@ -17,11 +17,7 @@
 package com.google.cloud.spring.logging;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.mock;
 
-import com.google.cloud.logging.Logging;
-import com.google.cloud.logging.LoggingOptions;
-import com.google.cloud.spring.core.UserAgentHeaderProvider;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
```diff
@@ -67,8 +68,9 @@ void testGetLoggingOptions() {
         .contains("Spring");
 
     loggingAppender.start();
-    // java-logging-logback appender default is ERROR
-    verify(logging).setFlushSeverity(com.google.cloud.logging.Severity.ERROR);
+    // java-logging-logback appender default is OFF
+    // See: https://github.com/googleapis/java-logging-logback/pull/1441
+    assertThat(logging.getFlushSeverity()).isNull();
```
zhumin8, line 73: `logging` is a mock ([here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blame/c6b6e457addf5d561287dc8a885d8db5568c94dd/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java#L38)), I don't think this assert is useful.

zhumin8 (changes requested):
I also found this related fix, perhaps we can revert now that logging-logback's default is also OFF? Or perhaps keeping it is no harm. I don't know enough about stackdriver appender to make the call, can you look into it a bit?

You might also want to update [this test](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blame/c6b6e457addf5d561287dc8a885d8db5568c94dd/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java#L74-L81), now that with the OFF default, it passes with or without [this line](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blame/c6b6e457addf5d561287dc8a885d8db5568c94dd/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java#L78)

spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
JoeWang1127: Do you think we should remove this test and update this [test](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blame/c6b6e457addf5d561287dc8a885d8db5568c94dd/spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java#L75)?

zhumin8: Yes, I think removing this assertion line makes sense. The remaining part of the test is still valid.

spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
```diff
@@ -65,32 +62,14 @@ void testGetLoggingOptions() {
         .isNotNull()
         .contains("spring-cloud-gcp-logging")
         .contains("Spring");
```
JoeWang1127, line 64: Looking at the implementation of `LoggingAppender.java`, it only override `getLoggingOptions` method. I think the proper tests should verify whether the logging options are correctly set.

spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
```diff
@@ -65,32 +62,14 @@ void testGetLoggingOptions() {
...
-    loggingAppender.start();
-    verify(logging, never()).setFlushSeverity(any());
   }
 
   @Test
   void testSetLogDestinationProjectId() {
     LoggingAppender loggingAppender = new TestLoggingAppender();
     loggingAppender.setCredentialsFile("src/test/resources/fake-project-key.json");
     loggingAppender.setLogDestinationProjectId("my-log-destination-project");
-    assertThat(loggingAppender.getLoggingOptions().getCredentials()).isNotNull();
```
zhumin8, line 89: Can I ask why removing this check? It seems to be a valid check to see if credentials file is respected by loggingAppender.

JoeWang1127: This test seems overlap with the previous one.

spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
```diff
@@ -65,32 +62,14 @@ void testGetLoggingOptions() {
...
-  }
-
-  @Test
-  void testDisablingFlushLevel() {
-    LoggingAppender loggingAppender = new TestLoggingAppender();
-    loggingAppender.setLogDestinationProjectId("my-log-destination-project");
-    loggingAppender.setFlushLevel(Level.OFF);
-
-    loggingAppender.start();
-    verify(logging, never()).setFlushSeverity(any());
```
zhumin8, line 81: since we are removing all occurrences `verify(logging)`, I believe we can delete the TestLoggingAppender class and revert back to use the base class for tests in this class. Basically reverting most of changes made to this class from https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/2922/files#diff-cd11f343096de66a41d821803b737c68495d3e1f52f1c86dfe153b4fa8daba06

spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
```diff
@@ -65,32 +62,14 @@ void testGetLoggingOptions() {
...
   }
 
   @Test
   void testSetLogDestinationProjectId() {
     LoggingAppender loggingAppender = new TestLoggingAppender();
     loggingAppender.setCredentialsFile("src/test/resources/fake-project-key.json");
     loggingAppender.setLogDestinationProjectId("my-log-destination-project");
-    assertThat(loggingAppender.getLoggingOptions().getCredentials()).isNotNull();
-    assertThat(loggingAppender.getLoggingOptions().getProjectId()).isEqualTo("my-log-destination-project");
-    assertThat(loggingAppender.getLoggingOptions().getUserAgent())
```
zhumin8, line 91: This seems to be valid check for [this line](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/72ee57fb075a2866ba6e80cf3312f2bf911e8b07/spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/LoggingAppender.java#L43), if we setup this test with `LoggingAppender loggingAppender = new LoggingAppender();` instead

zhumin8 (changes requested):

spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
JoeWang1127: Removed

spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
JoeWang1127: OK, I restored the original assertions.

spring-cloud-gcp-logging/src/test/java/com/google/cloud/spring/logging/LoggingAppenderTests.java
zhumin8: That's fair. I was not looking at that one.

zhumin8:
nit: perhaps "fix(test)" for title? Technically, we are not changing logback default here, only observing behavior change due to upstream changes.

Issue #1270, opened 2022-09-25
Datastore: not equals filter support


dosomder:
It seems like datastore itself does support not equals filter: https://cloud.google.com/datastore/docs/concepts/queries

If I try to use a not equal filter in my Repository
```
  List<MyObject> findAllByMyIntegerPropertyNot(int myIntegerProperty);
```

I get the following exception
```
Caused by: org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataException: Unsupported predicate keyword: NEGATING_SIMPLE_PROPERTY (1): [IsNot, Not]
	at org.springframework.cloud.gcp.data.datastore.repository.query.PartTreeDatastoreQuery.lambda$applySelectWithFilter$1(PartTreeDatastoreQuery.java:355)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:575)
	at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
	at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:616)
	at org.springframework.cloud.gcp.data.datastore.repository.query.PartTreeDatastoreQuery.applySelectWithFilter(PartTreeDatastoreQuery.java:365)
	at org.springframework.cloud.gcp.data.datastore.repository.query.PartTreeDatastoreQuery.applyQueryBody(PartTreeDatastoreQuery.java:296)
	at org.springframework.cloud.gcp.data.datastore.repository.query.PartTreeDatastoreQuery.runQuery(PartTreeDatastoreQuery.java:193)
```

Is it possible to add NOT_EQUALS support?

JoeWang1127:
@dosomder Hi, from the [docs](https://cloud.google.com/datastore/docs/concepts/queries#datastore-datastore-basic-query-java) you mentions, I can see there is a `NOT_EQUAL` filter, anything wrong with that?

JoeWang1127:
[...]

PR #3635, opened 2025-03-06
feat: add support for `not equals` queries in datastore integration

diegomarquezp:
Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1270 ☕


diegomarquezp:
@jinseopkim0 the failing native tests were due to the branch being stale. Merging the latest of main did the trick.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -693,7 +704,8 @@ void ancestorsTest() {
   void referenceTest() {
     ReferenceEntity parent = saveReferenceEntitiesGraph();
 
-    ReferenceEntity loadedParent = this.datastoreTemplate.findById(parent.id, ReferenceEntity.class);
+    ReferenceEntity loadedParent =
```
diegomarquezp, line 707: some checkstyle problems arised so I formatted the whole file.

diegomarquezp:
Double checked for flakiness by reopening the PR. The datastore tests are not failing.

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/testdomains/ProductRepository.java
```diff
@@ -22,4 +22,6 @@
 import org.springframework.stereotype.Repository;
 
 @Repository
-public interface ProductRepository extends DatastoreRepository<Product, Key> {}
+public interface ProductRepository extends DatastoreRepository<Product, Key> {
```
jinseopkim0, line 25: Is this change necessary?

jinseopkim0 (approved):

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/it/testdomains/ProductRepository.java
diegomarquezp: It is not necessary - removing. Thanks for the catch!

spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQueryTests.java
```diff
@@ -152,6 +152,49 @@ private ProjectionInformation getProjectionInformationMock() {
...
+
+              assertThat(statement).isEqualTo(expected);
+
+              return EMPTY_RESPONSE;
+            });
+
+    when(this.queryMethod.getCollectionReturnType()).thenReturn(List.class);
+
+    this.partTreeDatastoreQuery.execute(params);
+    verify(this.datastoreTemplate).queryKeysOrEntities(any(), any());
```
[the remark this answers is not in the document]
diegomarquezp: Actually that's a great question. 
In short, this is a unit test based on Mocks, so we don't use the production capabilities of this integration in this file (i.e. we can't really execute queries). All other tests in this file use the same approach, so I just merged this new test in.
I believe the reason to restrict this with mocks is because the class under test builds the query and then sends it to the underlying client (I think its `DatastoreReaderWriter` via `DatastoreTemplate`, which is mocked), but using the underlying client converts it into an integration test, so we just restrict the testing to confirming how the queries/filters/etc are built instead.

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/fb27879a45672e9ced2d4250dfa9aef5ce0bbfa4/spring-cloud-gcp-data-datastore/src/test/java/com/google/cloud/spring/data/datastore/repository/query/PartTreeDatastoreQueryTests.java#L108-L120

If we want to actually perform the queries, we have the sample tests:
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/fb27879a45672e9ced2d4250dfa9aef5ce0bbfa4/spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-basic-sample/src/test/java/com/example/DatastoreBookshelfExampleIntegrationTests.java#L146-L154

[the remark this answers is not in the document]
JoeWang1127: Thanks for the explanation.

JoeWang1127 (approved):

PR #3664, opened 2025-03-14
fix: GcpProjectIdProvider when using Secret Manager

### LINKED ISSUE #3181: SecretManager autoconfiguration ignores and superseeds spring.cloud.gcp.project-id

### LINKED ISSUE #3564: GcpSecretManagerAutoConfiguration should have @AutoConfigureAfter(GcpContextAutoConfiguration.class)
When I upgrading to v6.0.0 the order on how `GcpSecretManagerAutoConfiguration` is loaded changed, resulting in an error on some of our integration tests.

I think this class should have `@AutoConfigureAfter(GcpContextAutoConfiguration.class)` since it assumes a bean of type `CredentialsProvider` .

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/84706c7add68ed013be7a8ed537b5b9726ac22be/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java#L42

zhumin8:
fixes #3181 , also fixes #3564

cherry-picked changes from https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3383 to allow ci properly run. Adding tests.
@PatrickGotthard 's original comment on root cause: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3181#issuecomment-2478682585

When using Secret Manager (sm) module, expected behavior for GcpProjectIdProvider should be:
- in Spring Bootstrap Phase (where sm if configured), order of discovery: secretmanager.project-id >>  gcp.project-id >> default (according to DefaultGcpProjectIdProvider)
- in Application Context, GcpProjectIdProvider bean should behave similar to other modules, with no knowledge of secretmanager.project-id, thus: gcp.project-id >> default

Current behavior prior to this fix, the discovery order for both cases are:
- secretmanager.project-id >> default 

Fix is to let SecretManagerConfigDataLocationResolver recognizing GcpProperties and **not** promoting GcpProjectIdProvider to application context.

Thanks @PatrickGotthard for your contribution!

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java
@@ -19,11 +19,13 @@ package com.google.cloud.spring.autoconfigure.secretmanager;
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.cloud.secretmanager.v1.SecretManagerServiceClient;
 import com.google.cloud.secretmanager.v1.SecretManagerServiceSettings;
+import com.google.cloud.spring.autoconfigure.core.GcpContextAutoConfiguration;
 import com.google.cloud.spring.core.GcpProjectIdProvider;
 import com.google.cloud.spring.core.UserAgentHeaderProvider;
 import com.google.cloud.spring.secretmanager.SecretManagerTemplate;
 import java.io.IOException;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
@@ -39,6 +41,7 @@ import org.springframework.context.annotation.Bean;
 @EnableConfigurationProperties(GcpSecretManagerProperties.class)
 @ConditionalOnClass(SecretManagerTemplate.class)
 @ConditionalOnProperty(value = "spring.cloud.gcp.secretmanager.enabled", matchIfMissing = true)
+@AutoConfigureAfter(GcpContextAutoConfiguration.class)
 public class GcpSecretManagerAutoConfiguration {
 
   private final GcpProjectIdProvider gcpProjectIdProvider;
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
@@ -95,7 +95,7 @@ public class SecretManagerConfigDataLocationResolver implements
   private static GcpProperties getGcpProperties(ConfigDataLocationResolverContext context) {
     return context
         .getBinder()
-        .bind(GcpProperties.PREFIX, GcpProperties.class)
+        .bind(GcpProperties.CORE_PROPERTY_PREFIX, GcpProperties.class)
         .orElse(new GcpProperties());
   }
 
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
@@ -84,10 +84,10 @@ public class SecretManagerConfigDataLocationResolver implements
         // lazy register the client solely for unit test.
         BootstrapRegistry.InstanceSupplier.from(() -> createSecretManagerClient(context)));
     // Register the GCP Project ID provider.
-    registerBean(
+    registerAndPromoteBean(
         context,
         GcpProjectIdProvider.class,
-        createProjectIdProvider(context));
+        BootstrapRegistry.InstanceSupplier.of(createProjectIdProvider(context)));
     // Register the Secret Manager template.
     registerAndPromoteBean(
         context,
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
@@ -84,10 +84,7 @@ public class SecretManagerConfigDataLocationResolver implements
         // lazy register the client solely for unit test.
         BootstrapRegistry.InstanceSupplier.from(() -> createSecretManagerClient(context)));
     // Register the GCP Project ID provider.
-    registerAndPromoteBean(
-        context,
-        GcpProjectIdProvider.class,
-        BootstrapRegistry.InstanceSupplier.of(createProjectIdProvider(context)));
+    registerBean(context, GcpProjectIdProvider.class, createProjectIdProvider(context));
     // Register the Secret Manager template.
     registerAndPromoteBean(
         context,
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
@@ -73,8 +73,7 @@ public class SecretManagerConfigDataLocationResolver implements
 
   private static void registerSecretManagerBeans(ConfigDataLocationResolverContext context) {
     // Register the Core properties.
-    registerBean(
-        context, GcpProperties.class, getGcpProperties(context));
+    registerBean(context, GcpProperties.class, getGcpProperties(context));
     // Register the Secret Manager properties.
     registerBean(
         context, GcpSecretManagerProperties.class, getSecretManagerProperties(context));
@@ -96,9 +95,11 @@ public class SecretManagerConfigDataLocationResolver implements
         BootstrapRegistry.InstanceSupplier.of(createSecretManagerTemplate(context)));
   }
 
-  private static GcpProperties getGcpProperties(
-      ConfigDataLocationResolverContext context) {
-    return context.getBinder().bind(GcpProperties.PREFIX, GcpProperties.class).orElse(new GcpProperties());
+  private static GcpProperties getGcpProperties(ConfigDataLocationResolverContext context) {
+    return context
+        .getBinder()
+        .bind(GcpProperties.PREFIX, GcpProperties.class)
+        .orElse(new GcpProperties());
   }
 
   private static GcpSecretManagerProperties getSecretManagerProperties(
@@ -108,23 +109,19 @@ public class SecretManagerConfigDataLocationResolver implements
         .orElse(new GcpSecretManagerProperties());
   }
 
-  private static GcpProjectIdProvider createProjectIdProvider(
-      ConfigDataLocationResolverContext context) {
-
+  @VisibleForTesting
+  static GcpProjectIdProvider createProjectIdProvider(ConfigDataLocationResolverContext context) {
     ConfigurableBootstrapContext bootstrapContext = context.getBootstrapContext();
-    
-    GcpSecretManagerProperties secretManagerProperties = bootstrapContext.get(GcpSecretManagerProperties.class);
+    GcpSecretManagerProperties secretManagerProperties =
+        bootstrapContext.get(GcpSecretManagerProperties.class);
     if (secretManagerProperties.getProjectId() != null) {
       return secretManagerProperties::getProjectId;
     }
-    
     GcpProperties gcpProperties = bootstrapContext.get(GcpProperties.class);
     if (gcpProperties.getProjectId() != null) {
       return gcpProperties::getProjectId;
     }
-    
     return new DefaultGcpProjectIdProvider();
-
   }
 
   @VisibleForTesting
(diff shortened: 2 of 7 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
```diff
@@ -80,30 +84,44 @@ private static void registerSecretManagerBeans(ConfigDataLocationResolverContext
         // lazy register the client solely for unit test.
         BootstrapRegistry.InstanceSupplier.from(() -> createSecretManagerClient(context)));
     // Register the GCP Project ID provider.
-    registerAndPromoteBean(
+    registerBean(
```
zhumin8: re: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3383#discussion_r1973994631
"Promote" is needed for beans that are needed during the ConfigData loading phase and are also required by the application at runtime. Definition [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/26e0deeadf02db31390b40d3e24608499e7274fb/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java#L178-L194)
In this case, we do expose this bean so it is needed.
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/26e0deeadf02db31390b40d3e24608499e7274fb/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/core/GcpContextAutoConfiguration.java#L65-L67

Reverting this change.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java
```diff
@@ -39,6 +41,7 @@
 @EnableConfigurationProperties(GcpSecretManagerProperties.class)
 @ConditionalOnClass(SecretManagerTemplate.class)
 @ConditionalOnProperty(value = "spring.cloud.gcp.secretmanager.enabled", matchIfMissing = true)
+@AutoConfigureAfter(GcpContextAutoConfiguration.class)
```
zhumin8, line 44: I believe this was missed. Adding this allows beans defined in [GcpContextAutoConfiguration.java](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/core/GcpContextAutoConfiguration.java) be accessible in application.
In this case specifically, together with registerAndPromoteBean() in resolver, let application access the correct GcpProjectIdProvider by injecting it.

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/test/java/com/example/SecretManagerSamplePropertyIdTest.java
```diff
@@ -0,0 +1,44 @@
...
+  // When secretmanager.project-id and project gcp.project-id are both set,
+  // secretManagerTemplate should recognize secretmanager.project-id
+  // but gcpProjectIdProvider should only capture gcp.project-id
+  @Test
+  void testProjectIdCorrect() {
+    String projectId = secretManagerTemplate.getProjectId();
+    assertThat(projectId).isEqualTo("spring-cloud-gcp-ci");
+    String gcpProjectId = gcpProjectIdProvider.getProjectId();
+    assertThat(gcpProjectId).isEqualTo("gcp-project-id");
+  }
```
mieseprem, line 43: Do you mind adding `spring.cloud.import=sm@` to the property list?
This property triggered the issue in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3181 and in a sense of regression-prevention test it would be nice to have it in.

mieseprem: I used your branch to run tests in my project and I can verify your changes lead to a valid projectId in global gcpProjectIdProvider.

But, now I know what to add to a new Issue (as said in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3181#issuecomment-2724211664). 
For now, you can leave out to add `spring.cloud.import=sm@`. This will be addressed in the other Issue.

meltsufin (approved):

jinseopkim0 (approved):

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerCompatibilityTests.java
```diff
@@ -34,17 +33,18 @@ class SecretManagerCompatibilityTests {
...
+        Arguments.of("sm@", "spring.cloud.gcp.secretmanager.project-id="));
   }
 
   @BeforeEach
   void init() {
     application = new SpringApplicationBuilder(SecretManagerCompatibilityTests.class)
         .web(WebApplicationType.NONE)
         .properties(
-            "spring.cloud.gcp.secretmanager.project-id=" + PROJECT_NAME,
+            // "spring.cloud.gcp.secretmanager.project-id=" + PROJECT_NAME,
```
JoeWang1127, line 47: Why not removing this line?

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolverUnitTests.java
```diff
@@ -90,4 +97,27 @@ void registerBean() {
     when(context.getBinder()).thenReturn(new Binder());
     when(context.getBootstrapContext()).thenReturn(defaultBootstrapContext);
   }
+
+  @Test
+  void testSecretManagerProjectIdTakesPrecedence() {
+    when(secretManagerProperties.getProjectId()).thenReturn("secret-manager-property-id");
+    GcpProjectIdProvider projectIdProvider =
```
JoeWang1127, line 104: Could you also add a line to setup gcpProperties? To make sure the secret manager property is indeed returned first.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolverUnitTests.java
```diff
@@ -90,4 +97,27 @@ void registerBean() {
...
+        SecretManagerConfigDataLocationResolver.createProjectIdProvider(context);
+    assertThat(projectIdProvider.getProjectId()).isEqualTo("secret-manager-property-id");
+  }
+
+  @Test
+  void testProjectIdUseCoreWhenBoSecretManagerProjectId() {
```
JoeWang1127, line 110: Typo in test name?

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
```diff
@@ -80,30 +84,41 @@ private static void registerSecretManagerBeans(ConfigDataLocationResolverContext
...
   }
 
-  private static GcpProjectIdProvider createProjectIdProvider(
-      ConfigDataLocationResolverContext context) {
-    GcpSecretManagerProperties properties = context.getBootstrapContext()
-        .get(GcpSecretManagerProperties.class);
-    return properties.getProjectId() != null
-        ? properties::getProjectId : new DefaultGcpProjectIdProvider();
+  @VisibleForTesting
+  static GcpProjectIdProvider createProjectIdProvider(ConfigDataLocationResolverContext context) {
```
JoeWang1127, line 110: Can we test the behavior without making it non private?

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerCompatibilityTests.java
zhumin8: good catch. removed.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolverUnitTests.java
zhumin8: done b269abf622e4e06fe1ea6e210bfd883f5f4a9bf5

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
zhumin8: I cannot think of a easy way to test it. Also think package-private is okay and still allow us the flexibility to changes in the future

JoeWang1127 (approved):
Thank you.

Issue #3560, opened 2025-02-15
Startup error when no credentials present on 6.0.0 due to Secret Manager client


jmitash:
**Describe the bug**
I'm not sure if this is actually a bug, or a desired change in default behavior.

On versions 5.11.0 and prior, including `spring-cloud-gcp-starter` can be done without a resolvable project ID or application default credentials. A warning is logged on startup:
> WARN ... c.g.c.s.core.DefaultCredentialsProvider  : No core credentials are set. Service-specific credentials (e.g., spring.cloud.gcp.pubsub.credentials.*) should be used if your app uses services that require credentials.

Despite the warning, the application starts up and can run.

On 6.0.0 and above, the same warning is logged, but is subsequently followed by an error which prevents startup:

> ERROR ... o.s.boot.SpringApplication               : Application run failed
>
> org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'secretManagerClient' defined in class path resource [com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.class]: Failed to instantiate [com.google.cloud.secretmanager.v1.SecretManagerServiceClient]: Factory method 'secretManagerClient' threw exception with message: Your default credentials were not found. To set up Application Default Credentials for your environment, see https://cloud.google.com/docs/authentication/external/set-up-adc.

I observed that in 6.0.0, the dependency on Secret Manager (`spring-cloud-gcp-secretmanager`) switched from optional to required. When autoconfigured, the Secret Manager client fails with the above error. I am not familiar with Secret Manager, but will assume that failing startup when autoconfigured is a desired behavior.

**\>\> Workaround \<\<**
Similar to as mentioned in #2512, Secret Manager can be disabled via an additional property to get around this.

I know not much can be done without credentials. My personal use case is using `spring-cloud-gcp-starter-logging` just to get the correct logback format for Google Cloud Logging, which does not require any credentials, so I often choose to run without them.

But with the required dependency, this changes the _default_ behavior of starting up with no application credentials from "a warning during startup" to "a failure to startup". Perhaps I missed it, but I didn't see this documented and am wondering if it was intentional and if the behavior will remain this way going forward.

**Sample - 5.11.0 POM**
```xml
<!-- pom.xml -->
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #3706, opened 2025-03-31
fix: secretmanager to be an optional dep in autoconfig module

diegomarquezp:
Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3560 ☕ 

It marks the secretmanager dependency as optional. The config loader in the Secret Manager autoconfig was assuming the existence of the now optional Secret Manager integration.
We now check for the class to be present before resolving the properties

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
@@ -53,12 +53,40 @@ public class SecretManagerConfigDataLocationResolver implements
    */
   private static SecretManagerServiceClient secretManagerServiceClient;
 
+  /**
+   * Checks if the property can be resolved by the Secret Manager resolver.
+   * For the check, we rely on the presence of the SecretManagerSyntaxUtils class, which is an
+   * optional dependency.
+   * Since optional dependencies can not be present at runtime, we explicitly check for its
+   * existence before resolving the property.
+   * If it's not present, it means this config resolver (which operates at early stages of Spring
+   * initialization) is not meant to be used.
+   * @return true if it contains the expected `sm@` or `sm://` prefix, false otherwise.
+   */
   @Override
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
```diff
@@ -53,14 +53,42 @@ public class SecretManagerConfigDataLocationResolver implements
    */
   private static SecretManagerServiceClient secretManagerServiceClient;
 
+  /**
+   * Checks if the property can be resolved by the Secret Manager resolver.
+   * For the check, we rely on the presence of the SecretManagerSyntaxUtils class, which is an
+   * optional dependency.
+   * Since optional dependencies can not be present at runtime, we explicitly check for its
```
diegomarquezp, line 60:
```suggestion
   * Since optional dependencies may not be present at runtime, we explicitly check for its
```

jinseopkim0 (approved):

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/SecretManagerConfigDataLocationResolver.java
```diff
@@ -53,14 +53,43 @@ public class SecretManagerConfigDataLocationResolver implements
...
+   * Since optional dependencies may not be present at runtime, we explicitly check for its
+   * existence before resolving the property.
+   * If it's not present, it means this config resolver is not meant to be used.
+   *
+   * @return true if it contains the expected `sm@` or `sm://` prefix, false otherwise.
+   */
   @Override
   public boolean isResolvable(ConfigDataLocationResolverContext context,
       ConfigDataLocation location) {
+    boolean secretManagerSyntaxUtilsPresent = isClassPresent("com.google.cloud.spring.secretmanager.SecretManagerSyntaxUtils");
```
zhumin8, line 69: @diegomarquezp A little late to this, but wonder you have this tested for GraalVM (the native test in ci may not cover this specifc scenario)? Since reflection is used here, wonder if we need to add to GraalVM hints?

Issue #3694, opened 2025-03-25
Only require minimal GCP permissions to publish to a topic


birnbuazn:
**Is your feature request related to a problem? Please describe.**

A have simple Spring Cloud Stream 6.0.1  project where I want to publish messages to a GCP topic. 

Unfortunately, Spring Cloud Stream will always try to check if a topic exists if it is provisioning the producer destinations in `com.google.cloud.spring.stream.binder.pubsub.provisioning.PubSubChannelProvisioner#provisionProducerDestination`. This happens even if `spring.cloud.stream.gcp.pubsub.default.producer.auto-create-resources` is set to `false`.

The GCP call will fail if the Google IAM identity does not have the required permission `pubsub.topics.list` to list all topics. Publishing to a specific topic should only require the `pubsub.topics.publish` permission, but now we have to add the additional list permission for Spring Cloud Stream to work, which should not be required. (see https://cloud.google.com/pubsub/docs/access-control)

Note, that this is similar to https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/2231, but here we are talking about the producer side of things.

**Describe the solution you'd like**

Checking the existence of a topic should be optional (or disabled if `auto-create-resources` is `false`).

**Describe alternatives you've considered**

The only viable alternative now is adding an extra permission to the IAM identity used when running the service. This works, but we really want to use a minimal set of permissions and listing all available topics shouldn't be required.

**Additional context**

Relevant part of the `application.yml`:

```
spring:
  cloud:
    function:
      definition: updateRequest

    stream:
      bindings:
        updateRequest-out-0:
          destination: private-edge.ecst.edge.requests

      gcp.pubsub.default:
        producer:
          auto-create-resources: false
```

Current stack trace when Spring Cloud Stream is trying to make sure that the topic exists:

```
[...]

PR #3717, opened 2025-04-03
fix: #3694: Producer destination: ensure topic exists or create one only if auto create resource is true

SheheryarAamir:
fix: #3694   Producer destination: ensure topic exists or create one only if auto create resource is true


meltsufin (approved):
LGTM

PR #3718, opened 2025-04-04
chore: cleanup release scripts to prepare for migration to internal

meltsufin:
b/308610909

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -448,7 +448,7 @@
 			<id>default</id>
 			<activation>
 				<property>
-					<name>!release</name>
+					<name>!performRelease</name>
 				</property>
 			</activation>
 			<modules>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -815,19 +815,6 @@
 							</excludePackageNames>
 						</configuration>
 					</plugin>
-					<plugin>
```
jinseopkim0, line 818: Was this removed on purpose?

meltsufin: Yes, this due to a very obscure issue with the `-DpreformRelease`. I kept getting a build error complaining about this plugin being defined twice. It turned out that Maven Super POM already has this plugin defined in a profile that is activated with `-DperformRelease`. See: https://maven.apache.org/ref/3.0.4/maven-model-builder/super-pom.html

jinseopkim0: Thank you for the explanation!

jinseopkim0 (approved):

PR #3744, opened 2025-04-26
[pubsub] Add delay threshold in duration form for spring properties

### LINKED ISSUE #3437: [cloud-gcp-pubsub] Add application config to set setDelayThresholdDuration in the millisecond range
I want to set `spring.cloud.gcp.pubsub.publisher.batching.delay-threshold-seconds` in the millisecond range. 
Currently I have a manual (code) configuration that sets the publisher using `setDelayThresholdDuration` to 100ms. I would like to configure this using the application config. But this is currently impossible as it only accepts 'whole' seconds.

Add an config option to directly set `setDelayThresholdDuration` in duration form.

- Add an option to set `spring.cloud.gcp.pubsub.publisher.batching.delay-threshold-seconds` but for milliseconds
- Add an an "free" option setting field.

MV-GH:
Fixes #3437

Currently adds delay threshold in duration, which takes priority over delay threshold in seconds.


MV-GH:
@zhumin8 @lqiu96 Hey, could someone take a look at this PR/issue?

PR #3796, opened 2025-05-20
chore: add jdk 24 to test matrixes

ldetmer:
this is in order to support graalvm upgrade

This required an upgrade in errorprone to support the support JDK 24

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -659,8 +659,7 @@
 							<artifactId>maven-compiler-plugin</artifactId>
 							<version>${maven-compiler-plugin.version}</version>
 							<configuration>
-								<source>${maven.compiler.source}</source>
-								<target>${maven.compiler.target}</target>
+								<release>${maven.compiler.source}</release>
 								<encoding>UTF-8</encoding>
 								<fork>true</fork>
 								<compilerArgs>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -659,11 +659,11 @@
 							<artifactId>maven-compiler-plugin</artifactId>
 							<version>${maven-compiler-plugin.version}</version>
 							<configuration>
-								<source>${maven.compiler.source}</source>
-								<target>${maven.compiler.target}</target>
+								<release>${maven.compiler.source}</release>
 								<encoding>UTF-8</encoding>
 								<fork>true</fork>
 								<compilerArgs>
+									<arg>--should-stop=ifError=FLOW</arg>
```
mpeddada1, line 666: Ah beginner question - what does this argument do?

ldetmer: Its part of the upgrade to errorprone 2.38.0 - they no longer accept the default tag, guidance if if you were using default flag update to use flow tag

https://github.com/google/error-prone/commit/e71db1f369a9367f6f2db34c4fbd006b6d6238fd

installation instructions: https://errorprone.info/docs/installation#maven

ldetmer: let me know if you think its worth it to leave comment

mpeddada1: Got it, thank you. Yes please, that would be great.

mpeddada1 (approved):

PR #3878, opened 2025-06-13
fix: increase timeout

JoeWang1127:


zhumin8 (approved):
Great find! Thank you @JoeWang1127

PR #3880, opened 2025-06-16
fix: use different channel names in pubsub emulator integration tests…

zhumin8:
… (#3878)

b/424814598

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubEmulator.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubEmulator.java
@@ -125,7 +125,7 @@ public class PubSubEmulator implements BeforeAllCallback, AfterAllCallback, Para
    * operation.
    */
   @Override
-  public void afterAll(ExtensionContext extensionContext) throws Exception {
+  public void afterAll(ExtensionContext extensionContext) {
     findAndDestroyEmulator();
   }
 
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -125,7 +125,7 @@ public String resolveParameter(
    * operation.
    */
   @Override
-  public void afterAll(ExtensionContext extensionContext) throws Exception {
```
blakeli0, line 128: Not sure this was discussed in previous PRs, removing `throws Exception` is actually a breaking change. I know this is in a test package and this is overriding a parent method so this change might be unavoidable, but still want to call it out in case we missed it.

JoeWang1127: Thanks for the catch.

Do you think this should be fine since it comes from samples?

Also, this method *never* throws exception through the `findAndDestroyEmulator` method.

JoeWang1127 (approved):

spring-cloud-gcp-pubsub-stream-binder/src/test/java/com/google/cloud/spring/stream/binder/pubsub/PubSubEmulator.java
blakeli0: If we don't expect customers to call `PubSubEmulator#afterAll` directly, then we should be fine. Otherwise they would still have to modify their code since `throws Exception` requires them to `try catch` previously. This is regardless if `findAndDestroyEmulator` throws exception or not.

zhumin8: I do not think PubsubEmulator#afterAll is expected to be called directly.  This class is a Junit extention to simplify test providing easy setup and teardown of a local PubSub emulator, afterAll does the teardown.

PR #4069, opened 2025-08-19
Include `parameter.adoc` in main documentation structure

### LINKED ISSUE #4061: Documentation for Parameter Manager not added to table of contents
I noticed this new feature in the release notes for Spring Cloud GCP 7.1.0:

> **parameter:** added support for remote configuration using the gcp parameter manager (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/3684) ([d73760e](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/commit/d73760e689e08fb4262a3126923b7702f00d2673))

I see in the commit that `docs/src/main/asciidoc/parameter.adoc` was added, and I can indeed open https://googlecloudplatform.github.io/spring-cloud-gcp/7.1.0/reference/html/parameter.html, but it seems this page wasn't added to the table of contents of the [Spring Cloud GCP 7.1.0 documentation](https://googlecloudplatform.github.io/spring-cloud-gcp/7.1.0/reference/html/), so this information is not accessible from there and kind of half-published.

I think `
include::parameter.adoc[]` should be added in `docs/src/main/asciidoc/index.adoc`?

skowrxn:
Fixes #4061

### DIFF
--- a/docs/src/main/asciidoc/index.adoc
+++ b/docs/src/main/asciidoc/index.adoc
@@ -65,6 +65,8 @@ include::secretmanager.adoc[]
 
 include::kms.adoc[]
 
+include::parameter.adoc[]
+
 include::config.adoc[]
 
 include::cloudfoundry.adoc[]


docs/src/main/asciidoc/index.adoc
```diff
@@ -67,6 +67,8 @@ include::kms.adoc[]
 
 include::config.adoc[]
 
+include::parameter.adoc[]
```
zhumin8, line 68: I'd imagine you want to add parameter doc right after kms?

skowrxn: yes, it makes more sense. I've updated it.

PR #4123, opened 2025-09-05
feat: Add support for configuring the Cloud SQL JDBC connection using a DNS name

hessjcg:
This updates the Spring GCP Cloud SQL configuration so that you can use an instance DNS name instead of an instance connection name. 

See [Using domain names to identify instances](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/docs/jdbc.md#using-dns-domain-names-to-identify-instances) in the Cloud SQL Java Connection documentation.

### DIFF
--- a/docs/src/main/asciidoc/sql.adoc
+++ b/docs/src/main/asciidoc/sql.adoc
@@ -167,8 +167,10 @@ However, it doesn't compromise the security of the communication because the con
 | `spring.cloud.gcp.sql.jdbc.enabled` | Enables or disables Cloud SQL auto-configuration for JDBC | No | `true`
 | `spring.cloud.gcp.sql.r2dbc.enabled` | Enables or disables Cloud SQL auto-configuration for R2DBC | No | `true`
 | `spring.cloud.gcp.sql.database-name` | Name of the database to connect to. | Yes |
-| `spring.cloud.gcp.sql.instance-connection-name` | A string containing a Google Cloud SQL instance's project ID, region and name, each separated by a colon. | Yes |
+| `spring.cloud.gcp.sql.instance-connection-name` | A string containing a Google Cloud SQL instance's project ID, region and name, each separated by a colon. | Yes* |
 For example, `my-project-id:my-region:my-instance-name`.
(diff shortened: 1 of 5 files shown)


docs/src/main/asciidoc/sql.adoc
```diff
@@ -186,6 +188,12 @@ Used to authenticate and authorize new connections to a Google Cloud SQL instanc
 | `spring.cloud.gcp.sql.adminQuotaProject` | A project ID for quota and billing. | No | (empty)
 |===
 
+*Valid configurations set exactly one of these properties:
```
jinseopkim0, line 191: Could we have a test case for when both the properties are enabled?

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessorTests.java
```diff
@@ -100,6 +103,25 @@ void testCloudSqlSpringDataSourceUrlPropertyOverride() {
             });
   }
 
+  @Test
+  void testCloudSqlDataSourceWithDnsName() {
```
jinseopkim0: Could we have a test case that uses PostgreSQL?

docs/src/main/asciidoc/sql.adoc
hessjcg: I added a testcase: testCloudSqlDataSourceWithDnsNameAndInstanceName. Technically, this is allowed. If both are set, the connector will use the instance connection name.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/CloudSqlEnvironmentPostProcessorTests.java
hessjcg: This is a good idea. Done.

Issue #4117, opened 2025-09-01
BeanDefinitionOverrideException: Invalid bean definition with name 'clientFactory' defined in class path resource [com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.class]


tompson:
**Describe the bug**  
After upgrading from Spring Cloud GCP **7.2.0** to **7.3.0**, our application fails to start due to a `BeanDefinitionOverrideException`. It seems that both `GcpSecretManagerAutoConfiguration` and `GcpParameterManagerAutoConfiguration` attempt to register a bean with the same name `clientFactory`.

Error trace (excerpt):

```
Caused by: org.springframework.beans.factory.support.BeanDefinitionOverrideException: Invalid bean definition with name 'clientFactory' defined in class path resource [com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.class]: Cannot register bean definition [Root bean: class=null; scope=; abstract=false; lazyInit=null; autowireMode=3; dependencyCheck=0; autowireCandidate=true; primary=false; fallback=false; factoryBeanName=com.google.cloud.spring.autoconfigure.secretmanager.GcpSecretManagerAutoConfiguration; factoryMethodName=clientFactory; initMethodNames=null; destroyMethodNames=[(inferred)]; defined in class path resource [com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.class]] for bean 'clientFactory' since there is already [Root bean: class=null; scope=; abstract=false; lazyInit=null; autowireMode=3; dependencyCheck=0; autowireCandidate=true; primary=false; fallback=false; factoryBeanName=com.google.cloud.spring.autoconfigure.parametermanager.GcpParameterManagerAutoConfiguration; factoryMethodName=clientFactory; initMethodNames=null; destroyMethodNames=[(inferred)]; defined in class path resource [com/google/cloud/spring/autoconfigure/parametermanager/GcpParameterManagerAutoConfiguration.class]] bound.
```

This issue did not occur in version 7.2.0, but appears consistently in 7.3.0.

---

**Sample**  
A minimal demo project reproducing the issue is available here: [troii/gcp-test-fail](https://github.com/troii/gcp-test-fail).  
The error occurs immediately on application startup when using Spring Boot with `spring-cloud-gcp-starter-secretmanager`.

---

**Expected behavior**  
[...]

PR #4127, opened 2025-09-08
fix(issues-4117)- Fixed invalid bean definition with name 'clientFactory'

suvidha-malaviya:
Issue: [#4117](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/4117)

Changes:

- Make Parameter Manager as `optional dependency` in AutoConfigure module.
- Changed the `clientFactory` bean name for both Secret Manager and Parameter Manager.
- Added logic to resolve the registered bean in `spring.factories` only if the class is present in the classpath for Parameter Manager.
- Added parameter manager module and starter module in `spring-cloud-gcp-dependencies`.


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java
```diff
@@ -77,7 +77,7 @@ public SecretManagerServiceClient secretManagerClient() throws IOException {
 
   @Bean
   @ConditionalOnMissingBean
-  public SecretManagerServiceClientFactory clientFactory() {
+  public SecretManagerServiceClientFactory secretManagerClientFactory() {
```
gptSanyam, line 80: This change is backwards incompatible, right?
If anyone is using the auto-configured bean ("clientFactory") then after this change that will result in error.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/parametermanager/GcpParameterManagerAutoConfigurationUnitTests.java
```diff
@@ -43,6 +44,13 @@ void init() {
         .withUserConfiguration(TestConfig.class);
   }
 
+  @Test
+  void testParameterManagerDependencyWithoutConfigImportShouldNotCrash() {
+    contextRunner
+        .withConfiguration(AutoConfigurations.of(GcpSecretManagerAutoConfiguration.class))
+        .run(ctx -> assertThat(ctx).hasSingleBean(ParameterManagerClientFactory.class));
```
gptSanyam, line 52: Let's assert that ctx has a bean for SecretManagerClientFactory.class as well - given that we are autoconfiguring SM beans as well.

suvidha-malaviya: Added test case for this.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/secretmanager/GcpSecretManagerAutoConfiguration.java
suvidha-malaviya: Yes, this change is backward-incompatible. However, the impact should be minimal, as the auto-configured clientFactory bean was mainly introduced to support global with multi-regional functionality and is not expected to be directly used by most users.

gptSanyam: @jinseopkim0 What do you think about the bean name changes?

Issue #4186, opened 2025-10-10
Flaky Test: `ConverterAwareMappingSpannerEntityReaderTests.readUnmatachableTypesTest()`


pem70:
## Description
This test `readUnmatachableTypesTest()` fails intermittently because it asserts the **exact exception message** (`"...: id"`).  
The underlying reader processes entity fields in a **non-deterministic order** (e.g., reflection order), so the first field that triggers the exception can change between runs.  
As a result, the message may not always end with `: id`, causing **flaky test failures** even though the exception type is correct.

## How to Reproduce
Run the test with the [NonDex](https://github.com/TestingResearchIllinois/NonDex) plugin to randomize iteration order:

```
mvn -pl spring-cloud-gcp-data-spanner edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=ConverterAwareMappingSpannerEntityReaderTests#readUnmatachableTypesTest
```

## Additional
I have a fix ready for this test, and I can make the PR if permitted.
Also, there is a typo in the test name.

pem70:
I have a PR that fixes this issue: #4187

PR #4187, opened 2025-10-10
fix: Flaky Test `ConverterAwareMappingSpannerEntityReaderTests.readUnmatachableTypesTest()`

pem70:
## Summary
This PR fixes a flaky test `readUnmatachableTypesTest` in  
`spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/convert/ConverterAwareMappingSpannerEntityReaderTests.java`.

The original test asserted an **exact exception message**:
```java
.hasMessage("Unable to read column from Cloud Spanner results: id");
```
However, the order in which entity fields are processed is **non-deterministic** (due to reflection order).  
As a result, the first failing field—and thus the message suffix—can vary between runs, leading to intermittent test failures.

## Root Cause
- `SpannerEntityReader` uses reflection over entity fields, which does not guarantee order.
- The test relied on a **fixed exception message**, making it brittle to iteration changes.

## Fix
The test has been updated to:
- Provide explicit read parameters for deterministic field mapping.
- Assert **message content** rather than full equality.
- Retain a stable suffix check for `: id`.

## Additional
There is a typo from the original test name, which is not modified due to consistency.


pem70:
This PR addressed issue #4186

Issue #2697, opened 2024-03-13
doc: Readme documentation and links are not up-to-date


zhumin8:
documentation and java doc links on README page is not up-to-date.

For 5.x latests, release please was able to pick up first occurrence of version, not others. 
For 3.x & 4.x branches, manual updates needed on each release.

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/adb57c6165a384b0236d8599598f8c699fccc6d3/README.adoc?plain=1#L13-L17

PR #4270, opened 2026-01-04
docs: update README documentation links to 7.4.2 #2697

mehdi0933:
updated the documentation links and the BOM version in the README.adoc file to point to the latest stable version (7.4.2)


zhumin8 (approved):
Thank you for noticing and fixing!
I removed the issue in description as that is referring to the process and the need for manual updates in each release.

PR #4280, opened 2026-01-13
fix: Use ADC in NativeTests

jinseopkim0:
Removes manual key handling in NativeTests.yaml and pom.xml. Relies on google-github-actions/auth to set GOOGLE_APPLICATION_CREDENTIALS. 

Fixes b/475547380.

### DIFF
--- a/spring-cloud-generator/showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/ComplianceSpringAutoConfiguration.java
+++ b/spring-cloud-generator/showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/ComplianceSpringAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Google LLC
+ * Copyright 2026 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-cloud-generator/showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/ComplianceSpringProperties.java
+++ b/spring-cloud-generator/showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/ComplianceSpringProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Google LLC
+ * Copyright 2026 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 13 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-starter-firestore-sample/src/test/java/com/example/FirestoreSampleAppNativeIntegrationTests.java
```diff
@@ -36,18 +36,10 @@
...
   @Test
-  void credentialsPropertiesAreAccessibleTest() {
-    logger.info(
-        "This test is only needed for graalvm compilation test"
-            + " to verify the encoded-key property is accessible."
-            + " Turn on by it.firestore.native=true");
-    String encodedKeyFromAutoConfig =
-        applicationContext.getBean(GcpFirestoreProperties.class).getCredentials().getEncodedKey();
-    assertNotNull(encodedKeyFromAutoConfig);
+  void contextLoads() {
```
zhumin8, line 43: I don't think this is equivelant to the specific test? The original test was added to verify a specific scenario: property set via spring.cloud.gcp.credentials.encoded-key is accessible (not null).  ADC credential resolution would succeed in this case still, without encoded-key.

I'd admit the original test setup was a bit lazy. Maybe a better way is to setup a dummy key to cloud.gcp.firestore.credentials.encoded-key? But then the test would fail at Spring Context loads, before getting to assertions.

spring-cloud-generator/showcase/showcase-spring-starter/src/main/java/com/google/showcase/v1beta1/spring/ComplianceSpringAutoConfiguration.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Google LLC
+ * Copyright 2026 Google LLC
```
zhumin8, line 2: FYI - there is a [recent fix](https://github.com/googleapis/sdk-platform-java/pull/4036) in generator, once the new libraries-bom is out and upgraded in spring-cloud-gcp, we can set these copyright years in tests to a fix var to avoid update each year.

spring-cloud-gcp-samples/spring-cloud-gcp-starter-firestore-sample/src/test/java/com/example/FirestoreSampleAppNativeIntegrationTests.java
jinseopkim0: Yes, this is not equivelant to the previous test. I removed the configuration from `spring-cloud-gcp-samples/pom.xml` in this PR, so the `spring.cloud.gcp.firestore.credentials.encoded-key` property is no longer set (it is effectively null). This new test is effectively a test to ensure ADC works.

zhumin8: ack

PR #4318, opened 2026-02-12
Chore: Migrate testing infrastructure and SpEL evaluation to Spring Boot 4.0 / Spring 7

prathmeshcc:
This PR contains the unit test fixes required for the Spring Boot 4.0 and Spring Framework 7 migration. These changes ensure that the test suite is compatible with the new modular architecture of Spring Boot 4, the upgraded Jakarta EE 11 baseline, and the refactored SpEL evaluation engine in Spring Data 4 (2025.1).

1. Mockito Infrastructure (Spring Test 7.0)
  * Annotation Migration: Replaced @MockBean and @SpyBean (deprecated in Boot 4.0) with @MockitoBean and @MockitoSpyBean from the new org.springframework.test.context.bean.override.mockito package.
  * Explicit Test Configuration: Added @AutoConfigureMockMvc and @AutoConfigureTestRestTemplate to integration tests. Spring Boot 4.0 no longer provides these beans implicitly with @SpringBootTest, favoring explicit opt-in for better modularity and performance.

  2. Spring Data 4 / SpEL Evaluation
   * ValueExpressionDelegate: Migrated repository query tests from QueryMethodEvaluationContextProvider (removed in Spring Data 4) to ValueExpressionDelegate. This aligns with the new centralized "Value Expression" infrastructure for SpEL resolution.
   * Standardized Mapping Exceptions: Updated Spanner and Datastore metadata tests to handle the new exception hierarchy where module-specific errors (e.g., SpannerDataException) are now wrapped in a top-level org.springframework.data.mapping.MappingException.

  3. Spring Boot 4.0 Actuator & Health API
   * Restructured Health API: Updated tests to use the new org.springframework.boot.health.contributor package.
   * Composite Contributors: Refactored Health Indicator tests to correctly handle CompositeHealthContributor types, ensuring tests pass when indicators are grouped by the framework.

  4. Dependency & Package Updates
   * Jakarta EE 11: Migrated javax.annotation imports to jakarta.annotation.
   * Jackson 3: Updated Jackson imports to the new tools.jackson namespace as required by the Spring Boot 4 baseline.
   * Modular Dependencies: Added spring-boot-resttestclient and other granular test starters to module POMs to support the new modular testing architecture.

  Verification Results
   * All unit tests passed locally via ./mvnw test -DskipITs -Dcheckstyle.skip -Dfmt.skip
   * Verified that all deprecated Mockito annotations have been removed.
   * Confirmed that EnvironmentPostProcessor registrations in spring.factories use the new Spring Boot 4 package paths.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImplTests.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImplTests.java
@@ -32,6 +32,7 @@ import com.google.spanner.v1.TypeCode;
 import java.util.List;
 import org.junit.jupiter.api.Test;
 import org.springframework.context.ApplicationContext;
+import org.springframework.core.NestedExceptionUtils;
 import org.springframework.data.core.TypeInformation;
 import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.data.mapping.PersistentPropertyAccessor;
@@ -132,8 +133,9 @@ class SpannerPersistentEntityImplTests {
     assertThatThrownBy(runnable::run)
         .satisfies(
             t -> {
-              assertThat(t.getCause()).isInstanceOf(SpannerDataException.class);
-              assertThat(t.getCause().getMessage()).isEqualTo(expectedMessage);
+              Throwable cause = NestedExceptionUtils.getMostSpecificCause(t);
+              assertThat(cause).isInstanceOf(SpannerDataException.class);
+              assertThat(cause.getMessage()).isEqualTo(expectedMessage);
             });
   }
 
@@ -303,8 +305,10 @@ class SpannerPersistentEntityImplTests {
   void testEmbeddedCollection() {
 
     assertThatThrownBy(() -> this.spannerMappingContext.getPersistentEntity(ChildCollectionEmbedded.class))
-        .satisfies(t -> {assertThat(t.getCause()).isInstanceOf(SpannerDataException.class);
-              assertThat(t.getCause().getMessage())
+        .satisfies(t -> {
+              Throwable cause = NestedExceptionUtils.getMostSpecificCause(t);
+              assertThat(cause).isInstanceOf(SpannerDataException.class);
+              assertThat(cause.getMessage())
                   .contains("Embedded properties cannot be collections: ");});
   }
 
--- a/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentPropertyImplTests.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentPropertyImplTests.java
@@ -25,6 +25,7 @@ import static org.mockito.Mockito.when;
 import java.util.List;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.core.NestedExceptionUtils;
 import org.springframework.data.mapping.PropertyHandler;
 import org.springframework.data.mapping.model.FieldNamingStrategy;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
@@ -50,11 +51,12 @@ class SpannerPersistentPropertyImplTests {
     assertThatThrownBy(() -> context.getPersistentEntity(TestEntity.class))
         .satisfies(
             t -> {
-              assertThat(t.getCause()).isNotNull();
-              assertThat(t.getCause().getMessage())
+              Throwable cause = NestedExceptionUtils.getMostSpecificCause(t);
+              assertThat(cause).isNotNull();
+              assertThat(cause.getMessage())
                   .contains("Invalid (null or empty) field name returned for ").contains(
                       "com.google.cloud.spring.data.spanner.core.mapping.SpannerPersistentPropertyImplTests$TestEntity");
-              assertThat(t.getCause().getMessage())
+              assertThat(cause.getMessage())
                   .matches("(?s).*SpannerPersistentPropertyImplTests\\$TestEntity\\.(id|doubleList|other)\\b.*");
             });
   }


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
```diff
@@ -67,12 +68,19 @@ void healthIndicatorPresent_defaults() throws Exception {
         .withBean("pubSubTemplate", PubSubTemplate.class, () -> mockPubSubTemplate)
         .run(
             ctx -> {
-              PubSubHealthIndicator healthIndicator = ctx.getBean(PubSubHealthIndicator.class);
+              HealthContributor healthIndicator = ctx.getBean(HealthContributor.class);
```
prathmeshcc: In Spring Boot 4.0, the Actuator Health API has been restructured (moved to org.springframework.boot.health.contributor). We now look up the bean by the more general HealthContributor interface instead of HealthIndicator. This is necessary because Spring Boot 4.0 now frequently wraps auto-configured indicators into a CompositeHealthContributor to support better grouping. This change, combined with the subsequent instanceof check, ensures the test remains resilient regardless of whether the indicator is standalone or part of a composite group. Ref: [Migration Guide](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#:~:text=spring%2Dboot%2D-,health,-none)

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImplTests.java
```diff
@@ -114,20 +114,27 @@ void testExpressionResolutionFromApplicationContext() {
...
 
-    assertThatThrownBy(() -> spannerMappingContext.getPersistentEntity(EntityWithWronglyOrderedKeys.class))
-            .isInstanceOf(SpannerDataException.class)
-            .hasMessage("The primary key columns were not given a consecutive order. There is no property annotated with order 2 in EntityWithWronglyOrderedKeys.");
+    assertThatMappingExceptionCause(
+        () -> spannerMappingContext.getPersistentEntity(EntityWithWronglyOrderedKeys.class),
+        "The primary key columns were not given a consecutive order. There is no property annotated with order 2 in EntityWithWronglyOrderedKeys.");
+  }
 
+  private void assertThatMappingExceptionCause(Runnable runnable, String expectedMessage) {
```
prathmeshcc, line 132: In Spring Data 4.0 (part of the 2025.1 release train), the MappingContext has been updated to provide a more consistent error-handling contract across all modules. Metadata discovery and entity mapping failures are now uniformly wrapped in a top-level org.springframework.data.mapping.MappingException. This change ensures our tests accurately reflect this new exception hierarchy by asserting that our domain-specific SpannerDataException is preserved as the underlying cause of the framework's mapping failure.
Ref: [Spring Data 4 Migration Guide](https://github.com/spring-projects/spring-data-commons/wiki/Spring-Data-2025.1-Release-Notes#:~:text=Async%20query%20methods.-,%E2%9A%A0%EF%B8%8F,Breaking%20Change%3A%20Removal%20of%20deprecated%20API,-PersistenceConstructor%20Migration)

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
ldetmer: nit: rename healthIndicator var to healthContributor

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
```diff
@@ -67,12 +68,19 @@ void healthIndicatorPresent_defaults() throws Exception {
...
               assertThat(healthIndicator).isNotNull();
-              assertThat(healthIndicator.getSubscription()).matches(UUID_PATTERN);
-              assertThat(healthIndicator.getTimeoutMillis()).isEqualTo(5000);
-              assertThat(healthIndicator.isAcknowledgeMessages()).isFalse();
-              assertThat(healthIndicator.isSpecifiedSubscription()).isFalse();
+              PubSubHealthIndicator indicator;
+              if (healthIndicator instanceof CompositeHealthContributor composite) {
```
ldetmer: do we need to update our https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicator.java#L56 to instead extend HealthContributor?

Also for users who want access to PubSubHealthIndicator, is this something we need to note in release notes that its access mechanism has changed?

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentPropertyImplTests.java
```diff
@@ -48,12 +48,15 @@ void testNullColumnName() {
...
-            .hasMessageContaining("Invalid (null or empty) field name returned for ")
-            .hasMessageContaining("com.google.cloud.spring.data.spanner.core.mapping.SpannerPersistentPropertyImplTests$TestEntity")
-                .satisfies(t -> {
-                  String msg = t.getMessage();
-                  assertThat(msg).matches("(?s).*SpannerPersistentPropertyImplTests\\$TestEntity\\.(id|doubleList|other)\\b.*");
-                });
+        .satisfies(
+            t -> {
+              assertThat(t.getCause()).isNotNull();
+              assertThat(t.getCause().getMessage())
```
ldetmer: we have to call getCause().getMessage() to get the actual error message?

Does that mean in our code in places we are checking the error message or logging the error message, that we need to update the code to use getCause().getMessage() as well?

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
prathmeshcc: > do we need to update our PubSubHealthIndicator.java to instead extend HealthContributor?

  PubSubHealthIndicator currently extends AbstractHealthIndicator, which implements HealthIndicator, and HealthIndicator itself extends HealthContributor. Since it already implements the interface, and AbstractHealthIndicator provides the idiomatic base implementation for Spring Boot health indicators, I don't think we need to change the class hierarchy.

  > Also for users who want access to PubSubHealthIndicator, is this something we need to note in release notes that its access mechanism has changed?
  Yes, this is a breaking change. Because we now use CompositeHealthContributorConfiguration, the bean is registered as a HealthContributor and can be a CompositeHealthContributor if multiple templates exist. Users who previously injected PubSubHealthIndicator or HealthIndicator directly will need to inject HealthContributor and potentially navigate the composite. We should definitely include this in the "Breaking Changes" section of the release notes/migration guide.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
prathmeshcc: Done.

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentPropertyImplTests.java
prathmeshcc: Yes. Starting with Spring Framework 6.0, NestedRuntimeException was changed to no longer include the nested exception's message in its own getMessage() output (see [spring-framework#25162](https://github.com/spring-projects/spring-framework/issues/25162)). Since SpannerDataException extends DataAccessException (which inherits from NestedRuntimeException), we now have to explicitly access the cause to verify its specific error message in tests.


  > Does that mean in our code in places we are checking the error message or logging the error message, that we need to update the code to use getCause().getMessage() as well?

  It depends on the context:
   1. Logging: Most logging frameworks (Logback, etc.) will still print the full stack trace including all causes and their messages when we pass the exception object to the logger (e.g., LOGGER.error("message", ex)). So standard logging is likely unaffected.
   2. Custom Error Construction: If we have code that manually constructs an error response or a new exception message by calling ex.getMessage(), that code will now only see the top-level message. In those specific cases, we might need to update the logic to also include ex.getCause().getMessage() if the nested information is important for the user/troubleshooting.

I've updated the tests to reflect this Spring 7 behavior.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
ldetmer: I missed the migration to using CompositeHealthContributorConfiguration - can you include link here from the spring boot migration notes that say why we need to migrate to using CompositeHealthContributorConfiguration?

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentPropertyImplTests.java
ldetmer: Will you take care of #2 in a separate PR?

prathmeshcc: I did this in the [1st commit](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/4318/changes/1802968ca33cfd6ec3d68f0c28389152da8fd39c) for 1st PR that I raised.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
prathmeshcc: It is mentioned in [Spring 2.2 Release Notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.2-Release-Notes#health-indicator-changes)

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentPropertyImplTests.java
ldetmer: will revisit in separate PR

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfigurationTests.java
ldetmer: can you write out the actual breaking change you will include in release notes and we can review just to make sure we're making the right change.

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImplTests.java
ldetmer: for the release notes, we should just mention that all [MappingContext](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/mapping/context/MappingContext.html) functions will have any exceptions wrapped in MappingContextException.  If you were relying on specific root cause messages, you should update your code to use t.getCause().getMessage() to retrieve the specific error mesage.

spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/spanner/health/SpannerHealthIndicatorAutoConfigurationTests.java
```diff
@@ -82,11 +83,9 @@ void testSpannerHealthIndicatorCreatedWithQuery() {
 
   @Test
   void testSpannerHealthIndicatorNotCreated() {
-
-    ApplicationContextRunner contextRunnerNew = this.contextRunner.withPropertyValues("management.health.spanner.enabled=false");
-    assertThatThrownBy(() -> contextRunnerNew
-        .run(context -> context.getBean(SpannerHealthIndicator.class)))
-        .isInstanceOf(NoSuchBeanDefinitionException.class);
+    this.contextRunner.withPropertyValues("management.health.spanner.enabled=false")
```
ldetmer: should this also be reverted since we reverted this change?

prathmeshcc: Done.

spring-cloud-gcp-data-spanner/src/test/java/com/google/cloud/spring/data/spanner/core/mapping/SpannerPersistentEntityImplTests.java
jinseopkim0: Would it be possible to use [NestedExceptionUtils.getMostSpecificCause()](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/NestedExceptionUtils.html#getMostSpecificCause(java.lang.Throwable)) here?

prathmeshcc: Done.

PR #4336, opened 2026-02-23
feat(spanner): add support for @ReadOnlyProperty

### LINKED ISSUE #3753: Spring Data Cloud Spanner missing @ReadOnlyProperty support
spring-cloud-gcp-starter-data-spanner 5.11.2

Spring Data Cloud Spanner does not seem to have support for allowing a field of an entity to be populated from query results but prevent it from being written to the database.

This capability is needed for fields that are derived from queries (e.g., computed columns or subquery results) and should not be modified or saved.

Spring Data has the @ReadOnlyProperty which does just that and should be a supported feature.

Example
```
SELECT
  foo.*,
  EXISTS (
    SELECT 1
    FROM foo as bar
    WHERE bar.external_id = foo.id
  ) as has_bar
FROM
  foo
WHERE
  foo.name = @fooName
```

In this case we want the data mapping to populate `has_bar` on our entity when issuing a query but to ignore `has_bar` when inserting or updating the entity (since has_bar is not a real column).

### LINKED ISSUE #3992: Spanner Does Not Ignore Generated Values
**Describe the bug**

Hello,

I am using Spring Data Spanner 6.0.1. There are three members of an object I have, and one is auto generated by the database. I do not know how to ignore this. The error message is 

```
FAILED_PRECONDITION: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: Cannot write into generated column blah
```

The blah has been put in place of the actual column, there is not a typo or anything. If I remove this member it works, and if I include the member and just read from the database it works. There is not a mention in the docs of how to ignore. `@Transient` is not viable because it outright ignores for reads and writes. 

Any assistance you have would be greatly appreciated!

bradperkins-gh:
This change ensures that fields annotated with @ReadOnlyProperty are excluded from mutations while remaining readable.

Sample entity with a @ReadOnlyProperty field:
```
@Table(name = "custom_test_table")
private static class TestEntity {
   @PrimaryKey(keyOrder = 1)
   String id;
   
   @Column(name = "custom_col")
   String something;

   @NotMapped
   String notMapped;

   @ReadOnlyProperty
   String readOnly;
 }
```

The field `String readOnly` will be populated on queries, but will not be applied to mutations. The mutation behavior is the same as fields annotationed with '@NotMapped'.

Fixes #3753
Fixes #3992


bradperkins-gh:
I have signed the Google CLA.

PR #4341, opened 2026-02-25
fix: resolve integration test regressions and update samples for Spring Boot 4.0 / Spring 7 migration

prathmeshcc:
This PR focuses on resolving integration test regressions and updating sample applications to ensure full compatibility with Spring Boot 4.0 and Spring Framework 7.0.

Note: `EXTRA_MAVEN_FLAGS` in GitHub Actions: Standardized Maven flags across all workflows to ensure `-Dcheckstyle.skip=true` and `-Dfmt.skip=true` are applied where appropriate, before running the formatter and checkstyle plugins in the next PR.

Key Improvements

  1. AOT & Native Support
   * GcpRuntimeHints: Created and updated GcpRuntimeHints to provide necessary reflection and resource hints for Spring 7 AOT processing across core and sample modules
     (e.g., BigQuery).
   * Native Build Infrastructure: Integrated the native-maven-plugin with process-test-aot in the samples parent POM to enable reliable native image testing.
   * JDK 25 Baseline: Standardized the NativeTests.yaml workflow to use JDK 25

  2. R2DBC & Cloud SQL Stability
   * Cloud SQL SSL Fix: Upgraded cloud-sql-socket-factory to 1.28.0 to resolve a critical Netty 4.2 incompatibility that caused SSLHandshakeException.
   * Spanner R2DBC: Resolved circular dependencies in CustomConfiguration by implementing manual ConnectionFactory construction using @Value properties.
   * Identifier Quoting: Aligned SpannerR2dbcDialect with Spring Data R2DBC 4.0 by explicitly handling backtick quoting and disabling forced quoting where it caused Spanner
     syntax errors.

  3. Samples & Integration Refinement
   * JUnit 5 Transition: Migrated multiple samples and the spring-cloud-generator module from JUnit 4 to JUnit 5 (e.g., @Before -> @BeforeEach).
   * Dependency Alignment: Restored missing standard dependencies (spring-boot-starter-test, spring-boot-starter-integration) and resolved duplicates (e.g., awaitility)
     across samples.
   * Serialization Fixes: Refactored the Config Sample to return a Map instead of a proxy bean, preventing infinite recursion with Jackson 3.
   * Integration Support: Applied @EnableIntegration and manual taskScheduler configurations where standard autowiring became ambiguous in the new stack.

This baseline ensures that the project is functionally sound and strictly adheres to the architectural standards of the latest Spring ecosystem.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -225,6 +225,12 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
+			<exclusions>
+				<exclusion>
+					<groupId>org.junit.vintage</groupId>
+					<artifactId>junit-vintage-engine</artifactId>
+				</exclusion>
+			</exclusions>
 		</dependency>
 		<dependency>
 			<groupId>org.apache.commons</groupId>
@@ -449,7 +455,7 @@
 				<plugin>
 					<groupId>org.springframework.boot</groupId>
 					<artifactId>spring-boot-maven-plugin</artifactId>
-					<version>4.0.0</version>
+					<version>${spring-boot-dependencies.version}</version>
 					<executions>
 						<execution>
 							<id>process-test-aot</id>
--- a/spring-cloud-gcp-samples/pom.xml
+++ b/spring-cloud-gcp-samples/pom.xml
@@ -185,13 +185,18 @@
 								<enabled>true</enabled>
 							</metadataRepository>
 							<buildArgs>
-								<buildArg>--initialize-at-build-time=org.junit.jupiter.engine.discovery.MethodSegmentResolver</buildArg>
 								<buildArg>--initialize-at-build-time=org.apache.commons.logging.impl.WeakHashtable</buildArg>
 								<buildArg>--initialize-at-run-time=org.springframework.test</buildArg>
 								<buildArg>--initialize-at-run-time=org.springframework.boot.test</buildArg>
 								<buildArg>--initialize-at-build-time=org.springframework.test.context.aot.AotTestContextInitializers__Generated</buildArg>
 								<buildArg>--initialize-at-build-time=org.springframework.test.context.aot.AotTestAttributes__Generated</buildArg>
 							</buildArgs>
+							<exclusions>
+								<exclusion>
+									<groupId>org.junit.vintage</groupId>
+									<artifactId>junit-vintage-engine</artifactId>
+								</exclusion>
+							</exclusions>
 						</configuration>
 						<executions>
 							<execution>
@@ -206,7 +211,6 @@
 					<plugin>
 						<groupId>org.springframework.boot</groupId>
 						<artifactId>spring-boot-maven-plugin</artifactId>
-						<version>4.0.0</version>
 						<executions>
 							<execution>
 								<id>process-test-aot</id>


.github/workflows/NativeTests.yaml
```diff
@@ -66,7 +66,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
-          java-version: 17
+          java-version: 25
```
prathmeshcc, line 69: Updated native test workflows to JDK 25 to match the latest performance and GraalVM compatibility standards recommended for [Spring Framework 7.](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#:~:text=If%20you%E2%80%99re%20using%20GraalVM%E2%80%99s%20native%2Dimage%2C%20v25%20or%20later%20must%20be%20used.)

spring-cloud-gcp-samples/spring-cloud-spanner-r2dbc-samples/src/test/java/com/example/SpringDataR2dbcAppIntegrationTest.java
```diff
@@ -26,19 +29,27 @@ class SpringDataR2dbcAppIntegrationTest {
...
+
+  private WebTestClient webTestClient;
 
   @Autowired private BookRepository bookRepository;
 
   @Autowired private ObjectMapper objectMapper;
 
+  @BeforeEach
+  void setUp() {
+    this.webTestClient = WebTestClient.bindToServer()
```
prathmeshcc, line 48: Switched to manual `WebTestClient` binding in `@BeforeEach`. The default autowired client was failing to bind to the `RANDOM_PORT` reliably in the new test execution lifecycle.

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/spring-cloud-gcp-integration-pubsub-sample-polling-receiver/src/main/java/com/example/PollingReceiverApplication.java
```diff
@@ -31,22 +31,31 @@
...
 @SpringBootApplication
+@EnableIntegration
 public class PollingReceiverApplication {
   private static final Log LOGGER = LogFactory.getLog(PollingReceiverApplication.class);
 
   public static void main(String[] args) {
     SpringApplication.run(PollingReceiverApplication.class, args);
   }
 
+  @Bean
```
prathmeshcc, line 54: Explicitly enabled integration and defined a `taskScheduler` bean. In the new stack, default autowiring for schedulers became ambiguous in samples using both WebFlux and Spring Integration, leading to `UnsatisfiedDependencyException`.

spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/src/main/java/com/example/HelloController.java
```diff
@@ -30,7 +31,9 @@ public HelloController(MyAppProperties sampleConfig) {
   }
 
   @GetMapping("/")
-  public MyAppProperties json() {
-    return this.sampleConfig;
+  public Map<String, Object> json() {
```
prathmeshcc, line 34: Refactored the endpoint to return a `Map` instead of the `@RefreshScope` bean itself. Jackson 3 (used in SB 4) attempts to traverse the proxy object, which results in infinite recursion and a `StackOverflowError`.

spring-cloud-gcp-samples/spring-cloud-spanner-r2dbc-samples/src/main/java/com/example/CustomConfiguration.java
```diff
@@ -31,16 +34,44 @@
...
+    String url = String.format("r2dbc:cloudspanner://spanner.googleapis.com:443/projects/%s/instances/%s/databases/%s",
+        this.project, this.instance, this.database);
+    return ConnectionFactories.get(url);
+  }
+
+  @Bean
+  @Override
+  public R2dbcMappingContext r2dbcMappingContext(Optional<NamingStrategy> namingStrategy, R2dbcCustomConversions r2dbcCustomConversions, RelationalManagedTypes managedTypes) {
+    R2dbcMappingContext context = super.r2dbcMappingContext(namingStrategy, r2dbcCustomConversions, managedTypes);
+    context.setForceQuote(false);
```
prathmeshcc, line 73: Disabled forced quoting in `R2dbcMappingContext`. Spring Data R2DBC 4.0 quotes all identifiers by default, but Spanner's GoogleSQL dialect rejects backticks for standard table/column names in many contexts, leading to syntax errors.

spring-cloud-gcp-samples/spring-cloud-spanner-r2dbc-samples/src/main/java/com/example/CustomConfiguration.java
```diff
@@ -31,16 +34,44 @@
...
+import org.springframework.data.relational.RelationalManagedTypes;
+import org.springframework.data.relational.core.mapping.NamingStrategy;
 import org.springframework.stereotype.Component;
 
 @Configuration
 public class CustomConfiguration extends AbstractR2dbcConfiguration {
 
   @Autowired ApplicationContext applicationContext;
 
+  @Value("${gcp.project}")
```
prathmeshcc, line 47: Manually implemented the `connectionFactory()` bean. In Spring 7, extending `AbstractR2dbcConfiguration` while trying to autowire the `ConnectionFactory` creates a circular dependency. Explicit construction using `@Value` properties resolved this.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -35,14 +35,21 @@
 
 	<properties>
 		<gcp-libraries-bom.version>26.74.0</gcp-libraries-bom.version>
-		<cloud-sql-socket-factory.version>1.25.0</cloud-sql-socket-factory.version>
-		<r2dbc-postgres-driver.version>1.0.7.RELEASE</r2dbc-postgres-driver.version>
+		<cloud-sql-socket-factory.version>1.28.0</cloud-sql-socket-factory.version>
```
prathmeshcc, line 38: Upgraded to 1.28.0 to resolve a critical `SSLHandshakeException` ('Hostname or IP address is undefined'). This was caused by Netty 4.2 (shipped with SB 4.0) having stricter SSL session requirements that the 1.25.x connectors didn't satisfy.

spring-cloud-gcp-samples/pom.xml
```diff
@@ -168,6 +168,54 @@
 							</systemPropertyVariables>
 						</configuration>
 					</plugin>
+					<plugin>
+						<groupId>org.graalvm.buildtools</groupId>
```
prathmeshcc, line 172: Integrated the `process-test-aot` goal. This is mandatory for Spring Boot 4 native tests as it generates the required metadata and initializers during the build phase, preventing `ClassNotFoundException` at runtime.

spring-cloud-gcp-core/src/main/java/com/google/cloud/spring/core/aot/GcpRuntimeHints.java
```diff
@@ -0,0 +1,51 @@
+/*
```
prathmeshcc, line 1: Added `GcpRuntimeHints` to satisfy Spring 7's stricter AOT reflection requirements. These hints are essential for GraalVM to correctly initialize GCP client libraries and proxy beans without relying on runtime reflection.

spring-cloud-gcp-bigquery/src/test/java/com/google/cloud/spring/bigquery/integration/outbound/BigQueryFileMessageHandlerIntegrationTests.java
```diff
@@ -101,13 +101,6 @@ void testLoadFileWithSchema() throws InterruptedException, ExecutionException {
     CompletableFuture<Job> jobFuture =
         (CompletableFuture<Job>) this.messageHandler.handleRequestMessage(message);
 
-    // Assert that a BigQuery polling task is scheduled successfully.
```
ldetmer, line 104: why is this assertion removed?

spring-cloud-gcp-core/src/main/java/com/google/cloud/spring/core/aot/GcpRuntimeHints.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.aot.hint.MemberCategory;
+import org.springframework.aot.hint.RuntimeHints;
+import org.springframework.aot.hint.RuntimeHintsRegistrar;
+import org.springframework.aot.hint.TypeReference;
+import org.springframework.util.ClassUtils;
+
+/**
+ * Runtime hints for Spring Cloud GCP.
+ */
+public class GcpRuntimeHints implements RuntimeHintsRegistrar {
```
ldetmer, line 29: can you link to documentation that mentions this change for  upgrading?  I think its odd that we have to register spring classes, I would expect to only have to register classes we were doing something unusual with.

spring-cloud-gcp-samples/spring-cloud-gcp-bigquery-sample/src/main/java/com/example/BigQuerySampleRuntimeHints.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Runtime hints for the BigQuery sample.
+ */
+public class BigQuerySampleRuntimeHints implements RuntimeHintsRegistrar {
+
+  @Override
+  public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
+    hints.proxies().registerJdkProxy(
+        TypeReference.of(BigQuerySampleConfiguration.BigQueryFileGateway.class),
+        TypeReference.of("org.springframework.aop.SpringProxy"),
```
ldetmer, line 32: samething here, I would think spring would work with native out of box.  Is there documentation that says we need to include these runtime hints for spring framework classes?

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/spring-cloud-gcp-integration-pubsub-sample-polling-receiver/src/main/java/com/example/PollingReceiverApplication.java
```diff
@@ -31,22 +31,31 @@
...
+  public TaskScheduler pollingReceiverTaskScheduler() {
```
ldetmer, line 55: can you give some information about why this was added?

spring-cloud-gcp-samples/spring-cloud-spanner-r2dbc-samples/src/main/java/com/example/CustomConfiguration.java
ldetmer, line 47: what do these changes mean for the end user?  will they use this class differently?

spring-cloud-gcp-samples/pom.xml
```diff
@@ -168,6 +168,54 @@
...
+								<artifactId>junit-platform-native</artifactId>
+								<version>0.11.3</version>
+							</dependency>
+						</dependencies>
+						<configuration>
+							<metadataRepository>
+								<enabled>true</enabled>
+							</metadataRepository>
+							<buildArgs>
+								<buildArg>--initialize-at-build-time=org.junit.jupiter.engine.discovery.MethodSegmentResolver</buildArg>
```
ldetmer: Did you try using the exclusion statement here instead? see my [pom](https://paste.googleplex.com/5769603029532672)?  I really don't think we should be having all these --initialize-at-build unless absolutely necessary.

pom.xml
```diff
@@ -431,7 +449,7 @@
 				<plugin>
 					<groupId>org.springframework.boot</groupId>
 					<artifactId>spring-boot-maven-plugin</artifactId>
-					<version>${spring-boot-dependencies.version}</version>
+					<version>4.0.0</version>
```
ldetmer: what is the reason for not using the variable here?

spring-cloud-gcp-bigquery/src/test/java/com/google/cloud/spring/bigquery/integration/outbound/BigQueryFileMessageHandlerIntegrationTests.java
prathmeshcc: The test was removed because it created a race condition: BigQueryTemplate uses scheduleAtFixedRate to poll for job completion. If the job finishes quickly or the scheduler triggers the first poll immediately, the queue size drops to 0 before the assertion runs, causing a flaky failure. The subsequent `jobFuture.get()` call remains the primary and most robust verification, as it only returns once the background polling task has successfully confirmed the job is complete.

spring-cloud-gcp-core/src/main/java/com/google/cloud/spring/core/aot/GcpRuntimeHints.java
prathmeshcc: We added these hints because our custom Data modules (Spanner, Datastore) don't pull in the standard JPA starters that typically provide these reflection hints for pagination. Spring 7's AOT engine is more conservative and doesn't automatically register implementation classes like PageImpl when only the Page interface is visible to the controller. Without these, Jackson serialization fails in native images.

Supporting Link: [Spring Framework Documentation on Transitive AOT Detection](https://docs.spring.io/spring-framework/reference/6.1/core/aot.html#core.aot.hints.reflection) (noting that manual hints are required when the AOT engine cannot infer concrete types from the static analysis of the classpath)

spring-cloud-gcp-samples/spring-cloud-gcp-bigquery-sample/src/main/java/com/example/BigQuerySampleRuntimeHints.java
prathmeshcc: Spring AOT detection isn't perfect for every edge case, especially with dynamic proxies created by sub-projects like Spring Integration. GraalVM requires an exact match of the entire interface set. We are manually registering the gateway interface along with the standard Spring infrastructure interfaces to ensure the native runtime can successfully [instantiate the proxy ](https://docs.spring.io/spring-framework/reference/core/aot.html#core.aot.hints.proxies:~:text=The%20RuntimeHints%20API%20collects%20the%20need%20for%20reflection%2C%20resource%20loading%2C%20serialization%2C%20and%20JDK%20proxies%20at%20runtime.)that the Spring Integration factory creates.

pom.xml
prathmeshcc: I had updated it for debugging purposes. Reverted it back to use the variable.

spring-cloud-gcp-samples/spring-cloud-spanner-r2dbc-samples/src/main/java/com/example/CustomConfiguration.java
prathmeshcc: These changes don't change how a user interacts with the sample, but they align it with Spring 7's stricter Bean initialization rules. Specifically, we moved from fragile autowiring to explicit construction of the ConnectionFactory to prevent circular dependencies. We also disabled forced quoting to ensure Spanner remains compatible with the new Spring Data R2DBC 4.0 default behavior, which otherwise causes syntax errors.

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/spring-cloud-gcp-integration-pubsub-sample-polling-receiver/src/main/java/com/example/PollingReceiverApplication.java
prathmeshcc: All the taskSchedulers has been added for the reason stated above.

spring-cloud-gcp-samples/pom.xml
prathmeshcc: Great suggestion! I tried adopting the exclusion pattern, and it successfully allowed us to prune several flags. Here is the updated logic:

   1. Exclusion Added: I've added a global exclusion for junit-vintage-engine in the root POM. This confirmed that the MethodSegmentResolver hint was only needed for legacy JUnit 4 support, so I have now removed that flag.
   2. `WeakHashtable` (Restored): Even with the exclusion, the build failed because WeakHashtable (from commons-logging) is still reachable via LogFactory in the standard Spring/GCP stack. Since we are on the latest commons-logging (1.3.5) and it still places this object in the image heap, this build-time initialization remains strictly necessary for the image to generate.
   3. Spring AOT Machinery (Kept): I’ve kept the hints for org.springframework.test and AotTestContextInitializers__Generated. These follow the standard Spring Boot AOT Testing guidelines to ensure that pre-computed context metadata is baked into the binary correctly while keeping environment-specific state (like dynamic ports) fresh at runtime.

This resulting configuration is much cleaner while still satisfying GraalVM's strict initialization requirements for the Spring 7 stack.

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/spring-cloud-gcp-integration-pubsub-sample-polling-receiver/src/main/java/com/example/PollingReceiverApplication.java
ldetmer: are you referring to this "Applied @EnableIntegration and manual taskScheduler configurations where standard autowiring became ambiguous in the new stack."

 my confusion it doesn't look like this replaced anything, so where was the manual taskScheduler configurations occuring?

prathmeshcc: Actually, this didn't replace any existing manual code in the sample. It replaced a Spring Boot Autoconfiguration behavior that previously 'just worked' but became unreliable in the Spring 7 / Boot 4 migration.

   1. `@EnableIntegration`: In the new stack, Spring Boot's automatic enabling of Spring Integration infrastructure is more restrictive when multiple starters (like WebFlux) are present. Adding this explicitly ensures the Integration bus is fully initialized.
   2. `pollingReceiverTaskScheduler`: Without this explicit bean, we encountered UnsatisfiedDependencyException during startup because the container couldn't find a unique TaskScheduler to handle the @Poller in pubsubAdapter. By defining it here, we provide a dedicated, non-ambiguous scheduler for the synchronous pull polling task, ensuring stable application startup in the new environment.

jinseopkim0 (commented):
nit: the title of the PR contains "for for"

PR #4369, opened 2026-03-11
fix: enhance check with maven central

jinseopkim0:
This PR enhances the reliability of the Maven Central verification step in the release automation script (`util/release.go`). 

Previously, the script polled the `central.sonatype.com` web UI. This PR changes the logic to directly query the actual Maven repository (`repo1.maven.org`) to prevent false positives (such as soft 404s) and ensure the artifact is fully indexed before proceeding.

### DIFF
--- a/util/release.go
+++ b/util/release.go
@@ -756,11 +756,14 @@ func pollForPR(branch, query string, excludeSnapshot bool) string {
 // verifyMavenCentral polls Maven Central until the specified version is available.
 func verifyMavenCentral(branch, version string) {
 	prefix := fmt.Sprintf("[%s]", branch)
-	mavenURL := fmt.Sprintf("https://central.sonatype.com/artifact/com.google.cloud/spring-cloud-gcp/%s", version)
-	overviewURL := "https://central.sonatype.com/artifact/com.google.cloud/spring-cloud-gcp/overview"
+	// We check the actual repository instead of the web UI to avoid false positives from soft 404s.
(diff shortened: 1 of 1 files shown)


util/release.go
```diff
@@ -756,25 +756,38 @@ func pollForPR(branch, query string, excludeSnapshot bool) string {
 // verifyMavenCentral polls Maven Central until the specified version is available.
 func verifyMavenCentral(branch, version string) {
 	prefix := fmt.Sprintf("[%s]", branch)
-	mavenURL := fmt.Sprintf("https://central.sonatype.com/artifact/com.google.cloud/spring-cloud-gcp/%s", version)
-	overviewURL := "https://central.sonatype.com/artifact/com.google.cloud/spring-cloud-gcp/overview"
+	// We check the actual repository instead of the web UI to avoid false positives from soft 404s.
+	repoURL := fmt.Sprintf("https://repo1.maven.org/maven2/com/google/cloud/spring-cloud-gcp/%s/spring-cloud-gcp-%s.pom", version, version)
+	metadataURL := "https://repo1.maven.org/maven2/com/google/cloud/spring-cloud-gcp/maven-metadata.xml"
+	searchURL := fmt.Sprintf("https://central.sonatype.com/artifact/com.google.cloud/spring-cloud-gcp/%s", version)
```
lqiu96, line 762: Would this also potentially run into the potential soft 404s? Would it be better to just use either the meatadataURL or repoURL instead?

jinseopkim0: Yes, `metadataURL` and `repoURL` are used in the checks below.

PR #4387, opened 2026-03-27
fix: NullPointerException in spring-cloud-gcp-data-spanner (#4383)

jinseopkim0:
backport https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/4383


jinseopkim0:
> LGTM, but are the native failures expected?

They were transient failures. I re-ran the tests and they are now passing.

PR #4402, opened 2026-04-08
chore: add fallback mechanism to force PR creation via Renovate dashboard

jinseopkim0:
The script will now attempt to:
1. View the Renovate Dependency Dashboard.
2. Find and check the corresponding box for the rate-limited dependency.
3. Poll for up to 30 minutes until Renovate creates the new PR.

### DIFF
--- a/util/release.go
+++ b/util/release.go
@@ -211,14 +211,22 @@ func runReleaseForMain(startStep int) {
 	if shouldRun(step) {
 		confirmStep(branch, "Find and merge gapic-generator-java-bom upgrade PR")
 		fmt.Printf("%s ▶️ STEP %d: Finding gapic-generator-java-bom upgrade PR...\n", prefix, step)
-		gapicPR := findPR(branch, "gapic-generator-java-bom in:title is:open")
+		gapicSearchQuery := "gapic-generator-java-bom in:title is:open"
+		gapicPR := findPR(branch, gapicSearchQuery)
+
 		if gapicPR == "" {
 			if librariesBomOptionalOpt {
 				fmt.Printf("%s ℹ️ Skipping gapic-generator-java-bom upgrade PR (optional).\n", prefix)
 			} else {
-				fatalError(branch, "Could not find an open gapic-generator-java-bom PR. Exiting.")
+				// Fallback: Attempt to force creation from the Dependency Dashboard
+				gapicPR = forcePRCreationFromDashboard(branch, "gapic-generator-java-bom", gapicSearchQuery)
+				if gapicPR == "" {
+					fatalError(branch, "Could not find an open gapic-generator-java-bom PR even after forcing creation. Exiting.")
+				}
 			}
-		} else {
+		}
+
+		if gapicPR != "" {
 			fmt.Printf("%s ✅ Found gapic-generator PR: #%s\n", prefix, gapicPR)
 			approveWorkflowRuns(branch, gapicPR)
 			approvePR(branch, gapicPR)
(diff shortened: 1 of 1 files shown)


util/release.go
```diff
@@ -1174,3 +1190,63 @@ func updateReadmeAdoc(step int) {
 	fmt.Printf("[main] 🔗 PR URL: %s\n", prURL)
 }
 
+// forcePRCreationFromDashboard checks the appropriate box in the Renovate Dependency Dashboard (Issue #1705)
+// to force the creation of a rate-limited PR, then polls until the PR is created.
+func forcePRCreationFromDashboard(branch, keyword, prSearchQuery string) string {
+	prefix := fmt.Sprintf("[%s]", branch)
+	issueNumber := "1705"
```
diegomarquezp: qq are constants in go supposed to be declared in the body? Just curious if there is a distinctive format for a value that "seems" global like this one.

util/release.go
```diff
@@ -1174,3 +1190,63 @@ func updateReadmeAdoc(step int) {
...
+	body, err := runCmd("gh", "issue", "view", issueNumber, "--repo", "GoogleCloudPlatform/spring-cloud-gcp", "--json", "body", "--jq", ".body")
+	if err != nil {
+		fmt.Printf("%s ⚠️ Warning: Failed to fetch Dependency Dashboard: %v\n", prefix, err)
+		return ""
+	}
+
+	lines := strings.Split(body, "\n")
+	modified := false
+	for i, line := range lines {
+		if strings.Contains(line, "- [ ]") && strings.Contains(line, keyword) {
```
diegomarquezp, line 1210: Is it possible that the dashboard has one entry per each branch for the same dependency?
I believe in this case we want to check all occurrences regardless of the branch?

util/release.go
```diff
@@ -1174,3 +1190,63 @@ func updateReadmeAdoc(step int) {
...
+		return ""
+	}
+
+	fmt.Printf("%s ✅ Dashboard updated! Waiting for Renovate to create the PR...\n", prefix)
+
+	maxWait := 30 * time.Minute
+	start := time.Now()
+	for i := 0; i < 60; i++ { // 60 iterations of 30s = 30 mins
+		elapsed := time.Since(start).Round(time.Second)
+		fmt.Printf("\r%s ⏳ Polling for new PR up to %v... (Elapsed: %v)", prefix, maxWait, elapsed)
```
diegomarquezp, line 1240: "for up to %v minutes"

diegomarquezp (commented):
I like the automation intention in general. Hope renovate-bot had a REST API we could make direct calls to instead of string matching in the dependency dashboard, but this is definitely better than the manual work. Just a few questions and minor comment. Thank you for the PR.

util/release.go
jinseopkim0: Ack. Because `maxWait` is a `time.Duration`, Go automatically formats it with units (e.g., 30m0s... (Elapsed: 19m2s)).

util/release.go
jinseopkim0: Thanks for the comment. I've moved it to a package level constant.

util/release.go
jinseopkim0: Thanks for the feedback. `renovate-bot` is only configured for the `main` branch, so this cannot currently happen, but checking each line would be better in case things change. Updated code.

util/release.go
diegomarquezp: Got it, thank you!

diegomarquezp (approved):

PR #4417, opened 2026-04-16
fix(secretmanager): restore parsing of bytestrings

### LINKED ISSUE #4395: Spring Boot 4: Secret Manager placeholder "${sm@...}" does not bind to spring.data.mongodb.uri
## Summary

After upgrading to Spring Boot 4, a MongoDB URI configured via Spring Cloud GCP Secret Manager placeholder syntax does not bind into `spring.data.mongodb.uri`.

The application starts with the normal Spring Boot Mongo auto-configuration, but `MongoProperties.uri` ends up `null`, so Boot falls back to `mongodb://localhost/test`.

This setup worked before the Boot 4 upgrade.

## Versions

- Spring Boot: `4.0.4`
- Spring Cloud GCP: `8.0.1`
- Java: `25`
- MongoDB driver (from Boot stack): `5.6.4`
- Spring Cloud `2025.1.1`

## Configuration

External config is mounted as `/config/application.yaml` in the container.

```yaml
spring:
  config:
    import: sm@
  cloud:
    gcp:
      project-id: my-test-project
  data:
    mongodb:
      uri: "${sm@my_service_test__mongodb_uri}"
```

The referenced secret contains a valid SRV Mongo connection string, e.g.:

```text
mongodb+srv://user:password@cluster.example.mongodb.net/my-db?retryWrites=true&w=majority
```

## What works

We verified all of the following:

1. Spring Boot loads `/config/application.yaml`
2. `spring.config.import=sm@` is processed
3. `SecretManagerConfigDataResource{location=sm@}` is loaded
4. The `spring-cloud-gcp-secret-manager` property source is added
5. A diagnostic reading the resolved property later in startup can see the expected Mongo URI value
6. Spring Boot is using the standard `PropertiesMongoConnectionDetails` path, not a custom override

## What fails

Despite that, Boot’s Mongo properties are not bound from the placeholder-backed value.

We added a `MongoClientSettingsBuilderCustomizer` diagnostic and got:

```text
DIAG MongoProperties.uri=<null>
DIAG MongoConnectionDetails.type=org.springframework.boot.mongodb.autoconfigure.PropertiesMongoConnectionDetails
DIAG MongoConnectionDetails.connectionString=mongodb://localhost/test
```

Mongo then attempts to connect to `localhost:27017`.

## Relevant startup trace

Boot clearly loads the external config and Secret Manager config data:

```text
Loaded resource file [config/application.yaml]
Loaded resource SecretManagerConfigDataResource{location=sm@}
Adding imported property source 'spring-cloud-gcp-secret-manager'
```

But later the Mongo connection details are still:

```text
DIAG MongoProperties.uri=<null>
DIAG MongoConnectionDetails.connectionString=mongodb://localhost/test
```

## Why this looks like a Spring Cloud GCP issue

Spring Boot appears to behave normally here:

- `MongoConnectionDetails.type=PropertiesMongoConnectionDetails`
- Boot falls back to `mongodb://localhost/test` only because `MongoProperties.uri` is null

So the issue seems to be that the Secret Manager placeholder value from `${sm@...}` is not being bound as a usable `String` into `spring.data.mongodb.uri` under this Boot 4 setup.

## Additional observations

While debugging, we also hit an earlier failure when trying to resolve the property too early during bootstrap (probably [#2690](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/2690)):

```text
ConverterNotFoundException: No converter found capable of converting from type [com.google.protobuf.ByteString$LiteralByteString] to type [java.lang.String]
```

That suggests the Secret Manager property source may still be surfacing `ByteString` values in a way that is incompatible with Boot 4 property resolution/binding for this case.

## Expected behavior

`spring.data.mongodb.uri: "${sm@...}"` should bind into `MongoProperties.uri`, and Boot should use that URI to create the Mongo client.

## Actual behavior

`MongoProperties.uri` stays `null`, and Boot falls back to `mongodb://localhost/test`.

## Workaround

Using a plain resolved value instead of `${sm@...}` works, for example:

- materializing the URI directly into `application.yaml`, or
- injecting it through `SPRING_DATA_MONGODB_URI`

## Request

Can you confirm whether `${sm@...}` placeholders for Boot-bound properties like `spring.data.mongodb.uri` are expected to work on Spring Boot 4 with Spring Cloud GCP 8.0.1, and whether this is a known binding regression?

diegomarquezp:
Fixes https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/4395 ,where placeholders backed by Google Cloud Secret Manager (`${sm@my_secret}`) fail to parse into property structures like `spring.data.mongodb.uri`.

#### Problem
Spring Boot 4 scopes its `ConfigurationPropertiesBinder` atop an independent `ConversionService` that does not automatically cascade custom mappings injected during early `EnvironmentPostProcessor` cycles.

#### Solution
We adapted `SecretManagerPropertySource.java` to extract UTF-8 text dynamically upon reading.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.env.EnvironmentPostProcessor.imports
+++ b/spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.env.EnvironmentPostProcessor.imports
@@ -1,6 +1,0 @@
-# Migrated from spring.factories to enable native Spring Boot 3+ discovery
-com.google.cloud.spring.autoconfigure.alloydb.AlloyDbEnvironmentPostProcessor
-com.google.cloud.spring.autoconfigure.sql.CloudSqlEnvironmentPostProcessor
-com.google.cloud.spring.autoconfigure.sql.R2dbcCloudSqlEnvironmentPostProcessor
-com.google.cloud.spring.autoconfigure.secretmanager.GcpSecretManagerEnvironmentPostProcessor
-com.google.cloud.spring.autoconfigure.parametermanager.GcpParameterManagerEnvironmentPostProcessor
--- a/spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.env.EnvironmentPostProcessor.imports
+++ b/spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.env.EnvironmentPostProcessor.imports
@@ -1,0 +1,6 @@
+# Migrated from spring.factories to enable native Spring Boot 3+ discovery
+com.google.cloud.spring.autoconfigure.alloydb.AlloyDbEnvironmentPostProcessor
+com.google.cloud.spring.autoconfigure.sql.CloudSqlEnvironmentPostProcessor
+com.google.cloud.spring.autoconfigure.sql.R2dbcCloudSqlEnvironmentPostProcessor
+com.google.cloud.spring.autoconfigure.secretmanager.GcpSecretManagerEnvironmentPostProcessor
+com.google.cloud.spring.autoconfigure.parametermanager.GcpParameterManagerEnvironmentPostProcessor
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.env.EnvironmentPostProcessor.imports
jinseopkim0, line 1: Is this file change necessary?

jinseopkim0 (changes requested):

spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.env.EnvironmentPostProcessor.imports
diegomarquezp: It wasn't! Thanks for the catch.

jinseopkim0 (approved):

PR #4440, opened 2026-05-13
docs: add SKILLS.md for AI agent backporting guidelines

jinseopkim0:


.agents/skills/backport/SKILL.md
blakeli0, line 1: I think `.agents/skills` is probably a better location. Similar to what we did in [google-cloud-java ](https://github.com/googleapis/google-cloud-java/tree/main/.agents/skills)

SKILLS.md
```diff
@@ -0,0 +1,80 @@
+# Spring Cloud GCP AI Agent Skills: Backporting
```
blakeli0: A skill usually has a name and description for the AI to know when to use it. For example, https://github.com/google/skills/blob/main/skills/cloud/bigquery-basics/SKILL.md

jinseopkim0: Thanks for the feedback and the link. Updated.

blakeli0 (approved):

PR #4442, opened 2026-05-13
ci: disable multithreaded maven install to fix lock contention (NativeTests)

jinseopkim0:
b/512866590

### DIFF
--- a/.github/workflows/NativeTests.yaml
+++ b/.github/workflows/NativeTests.yaml
@@ -80,7 +80,6 @@ jobs:
           ./mvnw \
             --batch-mode \
             --no-transfer-progress \
-            --threads 1.5C \
             --define maven.test.skip=true \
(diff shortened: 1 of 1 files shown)


blakeli0 (approved):

.github/workflows/NativeTests.yaml
```diff
@@ -80,7 +80,6 @@ jobs:
           ./mvnw \
             --batch-mode \
             --no-transfer-progress \
-            --threads 1.5C \
```
blakeli0, line 83: How long the test is going to take when running on a single thread?

jinseopkim0: I spot-checked a few, and it appears to add 2-5 min per test (e.g. bigquery 13 min -> 15 min)

PR #4468, opened 2026-05-19
fix: Trim Cloud SQL instance connection name in JDBC URLs

### LINKED ISSUE #4467: Cloud SQL JDBC URL generation preserves trailing CR/LF in instance connection name
### Description

`spring-cloud-gcp-starter-sql-postgres` currently copies `spring.cloud.gcp.sql.instance-connection-name` into the generated JDBC URL without trimming it first.

This breaks when the value comes from an env/deployment system that adds a trailing carriage return or newline.

Example value:

```text
my-project:europe-west1:my-instance\r
```

Generated URL currently contains:

```text
cloudSqlInstance=my-project:europe-west1:my-instance\r
```

That `\r` then gets passed to the Cloud SQL socket factory as part of the instance name, so the connection fails.

### Expected behavior

The generated JDBC URL should trim accidental surrounding whitespace from `cloudSqlInstance`.

DerSimeon:
This pull request improves the handling of the `instanceConnectionName` property in the Cloud SQL JDBC URL builder by ensuring any leading or trailing whitespace is removed before use. Additionally, a new unit test was added to verify this behavior for PostgreSQL connections.

Improvements to JDBC URL handling:

* The `instanceConnectionName` property is now trimmed before being added to the JDBC URL parameters, preventing issues caused by accidental whitespace in configuration.

Testing enhancements:

* Added a new test (`DefaultCloudSqlJdbcInfoProviderTests`) to confirm that the `instanceConnectionName` is properly trimmed when building a PostgreSQL JDBC URL.

fixes: #4467


spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/DefaultCloudSqlJdbcInfoProviderTests.java
```diff
@@ -0,0 +1,25 @@
+package com.google.cloud.spring.autoconfigure.sql;
```
diegomarquezp, line 17: Can we add a license header to this file?

E.g. https://github.com/googleapis/spring-cloud-gcp/blob/7aafdc0f19ae868fc95c55e1a8ef857b6ea7de8c/spring-cloud-gcp-autoconfigure/src/test/java/com/google/cloud/spring/autoconfigure/sql/R2dbcCloudSqlEnvironmentPostProcessorTests.java#L1-L15

DerSimeon: Sorry for the late reply, added.

DerSimeon:
@diegomarquezp Anything else missing?
No idea why conventionalcommits is failing, even though the commits are correct.

diegomarquezp (approved):

PR #4525, opened 2026-07-08
fix(samples): Fix missing spring-boot-starter-restclient dependency in samples

lqiu96:
When upgrading to Spring Boot 4.1.0, some test auto-configurations were moved to spring-boot-restclient. Looks like from https://github.com/spring-projects/spring-boot/issues/50768 found a similar issue.

The samples should have spring-boot-restclient to auto configure the RestClient

Adds spring-boot-starter-restclient to all samples that use spring-boot-resttestclient to resolve ClassNotFoundException during integration tests.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-alloydb-sample/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-alloydb-sample/pom.xml
@@ -54,6 +54,11 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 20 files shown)


jinseopkim0 (commented):
Would it make sense to add the dependency to `spring-cloud-gcp-samples/pom.xml`?

lqiu96:
> Would it make sense to add the dependency to `spring-cloud-gcp-samples/pom.xml`?

I didn't see `spring-cloud-gcp-samples/pom.xml` define any dependencies. Looks like it's defined in each starter's pom.xml file and I followed the convention

jinseopkim0 (approved):

PR #4537, opened 2026-07-09
test: isolate sample databases per Java version to prevent concurrent CI flakes

jinseopkim0:
Isolate Datastore namespaces, Spanner databases, and Cloud SQL MySQL databases per Java specification version in sample integration tests. 

This prevents parallel runs of different Java versions in GHA matrices from colliding on shared resources (which drops tables and corrupts states concurrently, causing test failures).

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
@@ -18,6 +18,10 @@ package com.example;
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
+import com.google.cloud.spring.data.spanner.core.admin.SpannerDatabaseAdminTemplate;
+import com.google.cloud.spring.data.spanner.core.admin.SpannerSchemaUtils;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
 import org.junit.jupiter.api.extension.ExtendWith;
@@ -34,6 +38,10 @@ import org.springframework.test.context.junit.jupiter.SpringExtension;
 @SpringBootTest
 class MultipleDataModuleIntegrationTest {
 
+  @Autowired private SpannerDatabaseAdminTemplate spannerDatabaseAdminTemplate;
+
+  @Autowired private SpannerSchemaUtils spannerSchemaUtils;
+
   // The Spanner Repo
   @Autowired TraderRepository traderRepository;
 
@@ -43,6 +51,14 @@ class MultipleDataModuleIntegrationTest {
   @Autowired TraderService traderService;
 
   @Autowired PersonService personService;
+
+  @BeforeEach
+  void setUp() {
+    if (!this.spannerDatabaseAdminTemplate.tableExists("traders_repository")) {
+      this.spannerDatabaseAdminTemplate.executeDdlStrings(
+          List.of(this.spannerSchemaUtils.getCreateTableDdlString(Trader.class)), true);
+    }
+  }
 
   @Test
   void testMultipleModulesTogether() {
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/resources/application-test.properties
```diff
@@ -1,3 +1,5 @@
 spring.cloud.gcp.spanner.instance-id=spring-demo
-spring.cloud.gcp.spanner.database=trades
-spring.cloud.gcp.datastore.namespace=spring-demo
+spring.cloud.gcp.spanner.database=trades_${java.specification.version}
+
+spring.cloud.gcp.datastore.namespace=spring-demo-${java.specification.version}
```
lqiu96, line 4: does this re-use the same datastore namespace name? Do we need this to be unique or can it be the same?

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/resources/application-test.properties
```diff
@@ -1,3 +1,5 @@
 spring.cloud.gcp.spanner.instance-id=spring-demo
-spring.cloud.gcp.spanner.database=trades
-spring.cloud.gcp.datastore.namespace=spring-demo
+spring.cloud.gcp.spanner.database=trades_${java.specification.version}
```
lqiu96, line 2: do we also need to care about ITs vs native tests running at the same time?

Java 17 -> trades_1.17
Native 17 -> trades_1.17 (maybe?)

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
```diff
@@ -44,6 +52,14 @@ class MultipleDataModuleIntegrationTest {
 
   @Autowired PersonService personService;
 
+  @BeforeEach
+  void setUp() {
+    if (!this.spannerDatabaseAdminTemplate.tableExists("traders_repository")) {
```
lqiu96: qq: why would the table already exist?

Is this from another test not being able to clean up the repo? Would it be better to also have an afterEach to clean it up?

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/resources/application-test.properties
jinseopkim0: Across separate, concurrent PR builds of the same Java version (e.g. PR 1 vs PR 2 both running Java 17 tests), yes, it does re-use the same namespace name (e.g. spring-demo-17). Ideally, unique is the best, but reliably achieving that may be overkill, since sharing that should be fine for most cases.

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/resources/application-test.properties
jinseopkim0: No, we don't need to worry about ITs vs Native tests colliding. They run on completely separate GCP projects in CI:
- Integration Tests run on `spring-cloud-gcp-ci`.
- Native Tests run on `spring-cloud-gcp-ci-native`.

This provides full resource isolation, so they will never share the same database or namespace even if the names are identical.

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
jinseopkim0: Since `@BeforeEach` runs before every test method in the class, it executes multiple times. The table is created during the first test method's setup and remains in the database (since we only clear rows via `deleteAll()` and avoid slow `DROP TABLE` DDL operations).

jinseopkim0 (commented):
Thanks for the review!

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/resources/application-test.properties
lqiu96: Let's keep it for now since it'll help reduce the conflicts. If we see recurring issue, maybe we can see if we can get like a UUID in there

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleIntegrationTest.java
lqiu96: Can we just convert this to a @BeforeAll and @AfterAll? Just have both run once

jinseopkim0: Thanks for the feedback, I've updates as suggested.

lqiu96 (approved):

PR #4539, opened 2026-07-09
test: isolate Cloud SQL MySQL databases per Java version to prevent concurrent CI flakes

jinseopkim0:
Isolate Cloud SQL MySQL database names per Java specification version in sample integration tests, and set HikariCP property createDatabaseIfNotExist=true.

This prevents parallel runs of different Java versions in GHA matrices from dropping each other's tables and causing concurrent schema/data corruption.

### DIFF
--- a/spring-cloud-gcp-samples/pom.xml
+++ b/spring-cloud-gcp-samples/pom.xml
@@ -239,9 +239,11 @@
 								<gcs-write-bucket>gcp-storage-bucket-sample-output</gcs-write-bucket>
 								<gcs-local-directory>/tmp/gcp_integration_tests/integration_storage_sample</gcs-local-directory>
 								<spring.cloud.gcp.sql.instance-connection-name>spring-cloud-gcp-ci:us-central1:testmysql</spring.cloud.gcp.sql.instance-connection-name>
-								<spring.cloud.gcp.sql.database-name>code_samples_test_db</spring.cloud.gcp.sql.database-name>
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-samples/pom.xml
```diff
@@ -239,9 +239,11 @@
 								<gcs-write-bucket>gcp-storage-bucket-sample-output</gcs-write-bucket>
 								<gcs-local-directory>/tmp/gcp_integration_tests/integration_storage_sample</gcs-local-directory>
 								<spring.cloud.gcp.sql.instance-connection-name>spring-cloud-gcp-ci:us-central1:testmysql</spring.cloud.gcp.sql.instance-connection-name>
-								<spring.cloud.gcp.sql.database-name>code_samples_test_db</spring.cloud.gcp.sql.database-name>
+								<spring.cloud.gcp.sql.database-name>code_samples_test_db_${java.specification.version}</spring.cloud.gcp.sql.database-name>
```
lqiu96, line 242: qq, same questions as the other PR. Would it impact if it's run as IT vs native? Would it be the same?

jinseopkim0: No, we don't need to worry about ITs vs Native tests colliding. They run on separate GCP projects in CI:
- Integration Tests run on `spring-cloud-gcp-ci`.
- Native Tests run on `spring-cloud-gcp-ci-native`.

This provides full resource isolation, so they will not share the same MySQL database instance even if the database name is identical.

lqiu96 (approved):

PR #4540, opened 2026-07-09
ci: optimize integration and native test workflows with path-based job filtering

jinseopkim0:
Optimize `integrationTests.yaml` and `NativeTests.yaml` workflows by introducing a pre-filter step using `dorny/paths-filter`. 

This detects changes and dynamically runs only the test targets affected by the modified files in the PR, minimizing GHA runner resource usage and speeding up PR check feedback times.

### Verification and Testing:
- Verified that modifications to GHA workflow configuration files or global CI/CD directories trigger the fallback `infra` rule, which executes the full test matrix (verified in this PR's GHA checks).
- Created a separate verification pull request (PR #4541) targeting this feature branch, introducing a dummy comment change under `spring-cloud-gcp-data-datastore-sample`. Verified that the filter step successfully completed and triggered only the Datastore integration and native image tests, skipping all other targets.

b/533015135

### DIFF
--- a/.github/workflows/NativeTests.yaml
+++ b/.github/workflows/NativeTests.yaml
@@ -16,10 +16,173 @@ on:
 
 env:
   EXCLUDED_MODULES: spring-cloud-spanner-spring-data-r2dbc
-
 jobs:
+  filter:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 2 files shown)


lqiu96 (approved):
SG, let's try it. We can always add in more/ fix if we notice issues.

Issue #4597, opened 2026-08-11
GcpDatastoreAutoConfiguration  retains a Datastore client indefinitely for every dynamic namespace


gendolf3d:
When a  `DatastoreNamespaceProvider`  returns dynamic/high-cardinality namespaces,  `GcpDatastoreAutoConfiguration`  creates and permanently caches one  Datastore  client per namespace:

```
ConcurrentHashMap<String, Datastore> store = new ConcurrentHashMap<>();
return () -> store.computeIfAbsent(keySupplier.get(), this::getDatastore);
```

The cache has no expiration, size limit, or shutdown cleanup. Each client owns GAX executors, gRPC channels, watchdog threads, and native buffers. Previously used namespaces therefore remain alive indefinitely.

**Environment:**
• Spring Cloud GCP:  8.0.5 
• Spring Boot:  4.1.0 
• Google Cloud Datastore:  3.2.0 
• GAX:  2.82.0 
• gRPC:  1.81.0 
• Java:  25 

The same cache pattern appears in earlier Spring Cloud GCP versions, but newer dependency versions make its resource cost more visible.

**Minimal reproduction:**
Configure a dynamic namespace provider:
```
@Bean
DatastoreNamespaceProvider datastoreNamespaceProvider() {
    return () -> currentRequestDomainId();
}
```
Then perform Datastore operations using many distinct domain IDs. Every unique value creates another client that is retained for the application lifetime.

**Expected behavior:**
Dynamic namespace support should not cause unbounded retention of clients and their native resources. Cached clients should have a documented lifecycle and be closed when removed and during application shutdown.

**Suggested approaches:**
• Provide configurable expiration and/or cache limits.
• Call  Datastore.close()  when clients are evicted and on application shutdown.
• Expose an extension point for applications to provide a lifecycle-managed namespace client cache.
• Alternatively, share the underlying transport across namespace-specific clients.

Immediate eviction needs care because  DatastoreProvider  currently has no lease/release boundary, so a client could otherwise be closed while an operation is still using it.

jinseopkim0:
Starting in version `8.2.1`, you can configure Spring Cloud GCP to use HTTP transport by setting:

```yaml
spring:
  cloud:
    gcp:
      datastore:
        use-http: true
```

PR #4617, opened 2026-09-03
feat(datastore): add use-http property to configure HTTP transport

jinseopkim0:
## Description
In multi-tenant applications with dynamic namespaces, creating multiple Datastore client instances over gRPC can lead to high resource and thread consumption. This change introduces the \`spring.cloud.gcp.datastore.use-http\` configuration property, allowing applications to use HTTP transport to reduce overhead.

Relates to #4597

b/549105996

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
@@ -22,6 +22,7 @@ import com.google.cloud.NoCredentials;
 import com.google.cloud.datastore.Datastore;
 import com.google.cloud.datastore.DatastoreOptions;
 import com.google.cloud.datastore.DatastoreReaderWriter;
+import com.google.cloud.http.HttpTransportOptions;
 import com.google.cloud.spring.autoconfigure.core.GcpContextAutoConfiguration;
 import com.google.cloud.spring.core.DefaultCredentialsProvider;
 import com.google.cloud.spring.core.GcpProjectIdProvider;
@@ -76,6 +77,8 @@ public class GcpDatastoreAutoConfiguration {
 
   private final String host;
 
+  private final boolean useHttp;
+
   GcpDatastoreAutoConfiguration(
       GcpDatastoreProperties gcpDatastoreProperties,
       GcpProjectIdProvider projectIdProvider,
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -76,6 +77,8 @@ public class GcpDatastoreAutoConfiguration {
 
   private final String host;
 
+  private final boolean useHttp;
```
lqiu96: nit: Can we rename this to something like `useHttpJson` or something a bit more descriptive? I think `http` is a bit confusing as gRPC uses Http2

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/datastore/GcpDatastoreProperties.java
```diff
@@ -97,4 +100,12 @@ public boolean isSkipNullValue() {
...
     this.skipNullValue = skipNullValue;
   }
+
+  public boolean isUseHttp() {
+    return useHttp;
+  }
+
+  public void setUseHttp(boolean useHttp) {
+    this.useHttp = useHttp;
+  }
```
lqiu96: nit: Add some simple javadocs for these

lqiu96:
One overall question regarding Spring: Is using properties the Spring convention or would it be better to allow them to configure it via code? No concerns with this approach, just wondering what is the best approach.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
jinseopkim0: Thanks for the suggestion, updated to `useHttpJson` as suggested.

spring-cloud-gcp-autoconfigure/src/main/java/com/google/cloud/spring/autoconfigure/datastore/GcpDatastoreProperties.java
jinseopkim0: Done.

jinseopkim0:
> One overall question regarding Spring: Is using properties the Spring convention or would it be better to allow them to configure it via code? No concerns with this approach, just wondering what is the best approach.

Thanks for the question. Using properties is the standard Spring Boot convention for environment-specific settings (documented in the [Datastore reference docs](https://googlecloudplatform.github.io/spring-cloud-gcp/reference/html/index.html#cloud-datastore-settings) and used across all starters).

lqiu96 (approved):

